PR #40, opened 2015-08-27
job will now log to its named file in production mode resolves #36

### LINKED ISSUE #36: Jobs should log separately
Each job should have a separate log file specific to that job. There are a number of things to consider with this.

* logging in the master
* logging in the worker
* logging for the pipeline

Clearly the pipeline would log to the job specific log. It probably makes sense for all the worker logs to go to the job log. In the master anything that is specific to the job should go in the job log but everything else should go into the general log.

This may require changes to terafoundation in regard to how loggers are created.

jsnoble:


lib/utils/config.js
```diff
@@ -189,7 +198,7 @@ function periodic(msg) {
 function persistent(slicer, start, workerQueue) {
     var nextSlice = slicer;
     var start = start;
-    var logger = _context.logger;
+    var logger = _jobConfig.job_logger;
```
kstaken: This should be setup as jobConfig.logger

PR #98, opened 2015-12-23
worker validation logging and max_window size

jsnoble:


lib/utils/config.js
```diff
@@ -491,6 +491,20 @@ function logFinishedJob(start, job, analyticsData) {
...
 
+function validateNumOfWorkers(context) {
+    var job = validateOperation(context, jobSchema, getJob(), false);
+    var logger = context.logger;
+    var sysWorkers = context.sysconfig.terafoundation.workers;
+
+    if (job.workers > sysWorkers) {
+        logger.warn(' The number of workers specified on the job ( ' + job.workers + ' ) is higher than what has ' +
+            'been allocated in the config file at terafoundation.workers ( ' + sysWorkers + ' ). The number of workers ' +
+            'in the config file will be used ');
```
kstaken: Would be a good idea to explicitly specify the number of workers that will be used. Maybe replace the last sentence.

The job will run using + sysWorkers +  workers. If +  job.workers + are required adjust the value set for terafoundation.workers in the system configuration.

Also what happens here if terafoundation.workers isn't set at all?

lib/readers/elasticsearch_reader.js
```diff
@@ -242,6 +244,27 @@ function newSlicer(context, opConfig, jobConfig) {
...
+    client.cluster.stats({}).then(function(data) {
+        var version = data.nodes.versions[0];
+
+        if (checkVersion(version)) {
+            client.indices.getSettings({}).then(function(results) {
+                var data = results[opConfig.index].settings.index.max_result_window;
+                var window = data ? data : 10000;
+
+                logger.info(' max_result_window for index: ' + opConfig.index + ' is set at ' + window + '. Any slice that ' +
+                    'contains more documents will throw an error by elasticsearch. Please adjust accordingly. ')
```
kstaken: ' max_result_window for index: ' + opConfig.index + ' is set at ' + window + '. On very large indices it is possible that a slice can not be divided to stay below this limit. If that occurs an error will be thrown by Elasticsearch and the slice can not be processed. Increasing max_result_window in the Elasticsearch index settings will resolve the problem.'

lib/readers/elasticsearch_reader.js
```diff
@@ -242,6 +244,29 @@ function newSlicer(context, opConfig, jobConfig) {
     }
 }
 
+function checkVersion(str) {
+    var num = Number(str.replace(/\./g, ''));
+    return num > 210;
```
kstaken: Running on elasticsearch 2.1.0 does not display the warning. Should be >=

PR #185, opened 2016-04-22
Slicer changes

jsnoble:
I still need to verify how multiple slicers react in persistent mode, I also need to go through later and bring file_import and elasticsearch_data_generator up to work in slicers > 1


jsnoble:
should be able to set worker count as a top level teraslice configuration:
workers: 5

PR #295, opened 2016-08-01
Doc updates

jsnoble:


lib/cluster/services/api.js
```diff
@@ -177,14 +177,15 @@ module.exports = function(context, app, cluster_service, jobs_service) {
     app.get('/txt/workers', function(req, res) {
         var defaults = ['assignment', 'node_id', 'job_id', 'pid'];
         var workers = cluster_service.findAllWorkers();
-
+        console.log(workers);
```
kstaken: Is this supposed to be here? There are 4 console.logs in this file which look suspect.

PR #324, opened 2016-10-19
Added TOC to refernces.md

godber:
This may be more trouble than it's worth at this point without a systematic way of keeping the TOC up to date.  I guess I am saying, I don't know that this is worth it now that I have thought about it.  At least not until there is a serious push to maintain quality documentation.


godber:
Feel free to close without prejudice.

PR #336, opened 2016-11-08
made sure queue order is maintained or restart resolves #334

### LINKED ISSUE #334: pending jobs by order
if a job is enqueued, and on restart it re-enqueues, it does not guarantee order of pending jobs

jsnoble:
simple fix, though it only returns the last 10k jobs, is this ok, should the limit be higher especially for systems that have a lot of jobs?

### DIFF
--- a/lib/cluster/services/jobs.js
+++ b/lib/cluster/services/jobs.js
@@ -378,7 +378,6 @@ module.exports = function(context, cluster_service) {
         })
             .then(function() {
                 // Loads the initial pending jobs queue from storage.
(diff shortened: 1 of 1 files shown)


lib/cluster/services/jobs.js
```diff
@@ -379,12 +379,20 @@ module.exports = function(context, cluster_service) {
             .then(function() {
                 // Loads the initial pending jobs queue from storage.
                 // Currently this will not preserve the order. => maybe do it by timestamp if need order
```
godber: Does your change address this comment?

PR #337, opened 2016-11-09
Supporting limit the number of results printed to console for stdout

godber:
The `stdout` operation can now take a `limit` argument that limits
the result set printed out to the consoled based on the provided
number.  The default is still to print out all results.

Drat, I meant to push this to my personal fork instead of putting it here since it is fairly speculative.

If you guys don't like this functionality I am cool with you not accepting it.

However I would appreciate any feedback regarding how it's implemented.


godber:
The failed test appears to be a travis-ci error, compare the configs of the failing and passings tests:

* Failing: https://travis-ci.org/terascope/teraslice/jobs/174382636/config
* Passing: https://travis-ci.org/terascope/teraslice/jobs/174383268/config

Weird.

PR #350, opened 2016-12-14
Moderator

jsnoble:
This is meant to be a standing PR so you can check code/implementation etc

in teraslice config set "moderator": true

and on job you must set at top level configuration to:
```
"moderator": {
    "elasticsearch": ["es5", "default"]
  },
```

the array is the list of connections that correspond to terafoundation database connection names

TODO:
- [x] better error handling and abrupt connection loss logic on moderator
- [x] logic concerning when user specifies a pause command vs moderator resumes it
- [x] monitor state connection and logic to handle that
- [x] verify connection names actually exist on job vs config
- [x] remove repeat pause/resume commands
- [x] make pause and resume limits configurable
- [x] additional hooks into a job submission to check if a connection is ok to start a job
- [ ] documentation
- [x] testing
- [x] exponential backoff with randomness for workers on queue overloads
- [ ] (either now or later) moderator can adjust with slicer the amount of workers to should use
- [ ] (later) hooks into slicer to slow slicer at a intermediate level so there is no abrupt pause command

Questions:
 - should the moderator configuration live outside teraslice namespace and have its own seperate top level config? thinking not only about this but about any future services as well

Additional testing tips:
- lower the queue settings on bulk:
```
 curl -XPUT -sS localhost:9200/_cluster/settings -d '{ "transient" : {
 "threadpool.bulk.size" : "5",
 "threadpool.bulk.queue_size" : "5" } }'
```
- run a data generator job with stress_test set to true

### DIFF
--- a/lib/cluster/services/cluster.js
+++ b/lib/cluster/services/cluster.js
@@ -42,12 +42,13 @@ module.exports = function(context, server) {
     });
 
     //TODO need hooks for moderator
-    messaging.register('moderator:online', function(val) {
-        logger.warn('i got the val', val)
+    messaging.register('moderator:online', 'moderator', function(val) {
+        //registers moderator to socket behind the scenes
+        logger.info(`moderator has connected to cluster_master`)
     });
 
     messaging.register('moderator:pause_jobs', function(connectionList) {
-        logger.error('cluster master getting the pause job ebent')
+        logger.error('cluster master getting the pause job event');
         logger.debug('cluster_master receiving pause job events from moderator', connectionList);
         events.emit('moderate_jobs:pause', connectionList)
     });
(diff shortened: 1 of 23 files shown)


lib/cluster/services/cluster.js
```diff
@@ -664,3 +694,12 @@ module.exports = function(context, server) {
...
 
+
+/*
+
+ curl -XPUT -sS localhost:9200/_cluster/settings -d '{ "transient" : {
+ "threadpool.bulk.size" : "1",
+ "threadpool.bulk.queue_size" : "1" } }'
+
+
+ * */
```
godber: Move this to ticket

PR #365, opened 2017-02-15
Id slicer init errors

jsnoble:
added proper error handlers to id_slicer and to the sublice by key option for the date range slicer.
It also defaults to false


kstaken:
Rebase required after merging of #363

PR #384, opened 2017-03-17
added fields capability to elasticsearch_reader resolves #380

### LINKED ISSUE #380: Add option to select fields to ES reader
Should allow the job to have a fields list on the reader which is passed to elasticsearch to reduce the size of query results when only certain fields are required.

jsnoble:


lib/data_sources/elasticsearch.js
```diff
@@ -383,16 +382,20 @@ module.exports = function(client, logger, opConfig) {
...
         };
 
-        if (source.query) {
-            query.q = source.query;
+        if (opConfig.query) {
+            query.q = opConfig.query;
+        }
+
+        if (opConfig.fields) {
+            query._source = opConfig.fields;
```
kstaken: Unfortunately the way this works is different between ES 2 and ES 5. `_source` filtering only exists in 5.x. In 2.x it is called `fields` so this implementation is only going to work on ES 5.x.

Issue #386, opened 2017-03-21
Error when using a config file without default connector


jeffmontagna:
Discovered that teraslice requires that the default connector section is defined.  We deployed ES 2.3.2 and ES 5.2.2 clusters with a teraslice master/worker pair.  The config file only specified the data nodes as tera4 and dev2.  The default section was removed.  When submitting the job there was an error that job failed.

Output from submitting a job to the tera4 cluster
```
balor:inwork montaj1$ curl -XPOST ts-master:5678/jobs -d@data_generator.json
{
    "error": 500,
    "message": "Job submission failed: Error validating job: Error: elasticsearch_bulk connection_map specifies a connection for [default] but is not found in the system configuration [terafoundation.connectors.elasticsearch]"
}balor:inwork montaj1$
```
After getting this error I setup two local instances of ES 2.3.2. I added the the two servers to the processor-master.yaml file.  Submitting a job worked without issue.  I then re-ran the job, but removed  the "default" section. This resulted in teraslice not starting up.

Config file for local teraslice master : processor-master.yaml
```
terafoundation:
    environment: 'development'

    connectors:
        # ***********************
        # Elastic Search Configuration
        # ***********************
        elasticsearch:
            ##default:
            ##    host:
            ##        - "127.0.0.1:9200"          
            es1:
                host:
                    - "127.0.0.1:9200"
            es2:
                host:
                    - "127.0.0.1:9201"

teraslice:
    ops_directory: '/app/source/examples/ops/'

    master: true
    master_hostname: "127.0.0.1"
    name: "teracluster"
```

Local Jobs File:
```
{
  "name": "Data Generator",
  "lifecycle": "persistent",
  "workers": 1,
  "operations": [
    {
      "_op": "elasticsearch_data_generator",
      "size": 5000
    },
    {
      "_op": "elasticsearch_index_selector",
      "index": "example-logs",
      "type": "events"
    },
    {
      "_op": "elasticsearch_bulk",
      "size": 5000,
      "connection": "es1"
    }
  ]
}
```
Teraslice on output:
```
balor:teraslice montaj1$ node service.js -c ./examples/config/processor-master.yaml
{"name":"teraslice","hostname":"balor.local","pid":79073,"level":30,"msg":"Service starting","time":"2017-03-20T20:51:06.143Z","v":0}
[...]

PR #397, opened 2017-03-29
jobs can be run with out default set resolves # 386

jsnoble:


godber:
Looks like a test fails:

https://github.com/terascope/teraslice/blob/master/spec/processors/elastic_bulk-spec.js#L251-L253

godber:
refs #386

PR #402, opened 2017-04-12
refactored code to use queue module

jsnoble:


lib/processors/save_file.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
```
kstaken: Is this file supposed to be part of this PR?

package.json
```diff
@@ -31,7 +31,8 @@
     "socket.io-client": "^1.4.5",
     "terafoundation": "terascope/terafoundation",
     "uuid": "^2.0.1",
-    "yargs": "^3.18.0"
+    "yargs": "^3.18.0",
+    "queue": "git+https://git@github.com/terascope/queue.git"
```
kstaken: Can't this just be set to `terascope/queue`?

PR #409, opened 2017-05-03
re-enabling all tests

godber:
It appears that a few tests were left in "focus" mode in the last PR.  I have remedied that.


godber:
This should just be accepted.

PR #410, opened 2017-05-03
Doc improvements for custom processors

godber:
This PR makes it possible to use the `preserve_id` option on the `elasticsearch_index_selector` with data that comes from the `elasticsearch_data_generator`.  It assumes that the user has added document IDs to the data sometime between generation and index selection.

The PR changes the `post_validation` for the `elasticsearch_index_selector` to skip validation if the first operation in the job is the `elasticsearch_data_generator`.  It extends existing tests to ensure this works.  There is one commit that just cleans up `spec/processors/elasticsearch_index_selector-spec.js`.

This PR also includes:
* basic documentation on implementing custom operations
* renamed the `doc/reference.md` to `doc/ops-reference.md`
* updates to README.md to accommodate both of the above
* an improved error message in `lib/cluster/runners/op.js` to help debug


docs/custom_operations.md
```diff
@@ -0,0 +1,87 @@
+# Custom Operations
+
+It is possible to implement custom operations and utilize them in your Teraslice
+jobs.
```
godber: I wanted to add a statement here about what sort of operations a user can implement.  Is it pretty much possible to implement anything?  Readers, processors, writers?

docs/custom_operations.md
```diff
@@ -0,0 +1,87 @@
...
+    });
+
+    return function(data) {
+        logger.info('Inside custom processor \'count\'');
+        logger.info('Number of items in data: ' + Object.keys(data).length);
+        return data;
+    };
+}
+
+// FIXME: Help me describe this.
```
godber: At the time I couldn't describe a schema very well.  But now I think I would say "A schema function is required on all operations.  It can be used to verify that jobs meet the requirements of your custom processor or it can just return an empty object if you want to skip these tests."

lib/cluster/runners/op.js
```diff
@@ -74,7 +74,7 @@ module.exports = function(context) {
             catch (err) {
                 //if it cant be required check first error to see if it exists or had an error loading
                 if (error.message !== 'missing path') {
-                    throw new Error(`Error loading module: ${op_name} , error: an error occurred while attempting to load the code`);
+                    throw new Error(`Error loading module: ${op_name}, the following error occurred while attempting to load the code: ${error.message}`);
```
godber: I kinda wonder if this whole `catch (err)` block should just be replaced with a message that prints out the error.  I don't know how much value is added by parsing the error then just printing out different error messages.  If we **did something** with specific errors, then it might make sense to test the error output, but all we do is throw different messages under different errors.

godber:
I really dislike the fact that this implementation assumes that the user adds IDs to the document sometime between data generation and index selection.  Maybe I'd feel better if I added just the right documentation on the `preserve_id` option.

docs/custom_operations.md
jsnoble: they can create any slicer/reader and processor

lib/cluster/runners/op.js
jsnoble: those are different messages per different errors and can help differentiate the real issue. There is a difference between trying to load something thats not there, and a malformed op that throws when I try to load it

jsnoble:
My concern is that this is that I kinda feel like the data generator should be able to have a setting that it returns the data the same way elasticsearch would. I sorta remember you two mentioning something about some issue to solve in index_selector so if it really needs to be there than its fine. It just seems off that we would need to hard code the != datagenerator in the code. As for the id generation, that seems like it should be a feature of data_generator if we returned a response similar to elasticsearch. Maybe we could do both?

godber:
I think if we were to modify the `elasticsearch_data_generator` to generate ES like output we'd end up with code that looks an awful lot like the code that's already in `elasticsearch_index_selector` which is probably a clue that there's a better option.

Maybe we need to be more concrete about the "shape" of data at our inputs and outputs.  That is, rather than using options on our operations to allow them to accept different types/shapes of data we should only accept one type of data and force ourselves to use intermediate processors to modify data so that it matches an operator's required input type.

For example, if the current `elasticsearch_index_selector` were split into two things:

* `elasticsearch_bulk_request_generator` - takes in array of non-es objects, makes a bulk request type array
* `elasticsearch_index_selector` - takes in an bulk request type Array, returns a bulk request type array with modified index

So now I am just stuck with answering the awkward question: "Isn't it natural for `elasticsearch_bulk_request_generator` to set an index, thus forcing us to duplicate the functionality of the `elasticsearch_index_selector`?"  I'm not sure about this one yet.

godber:
OK, I have reverted my main code changes in this pull request so that it contains the following changes:

* basic documentation on implementing custom operations
* renamed the doc/reference.md to doc/ops-reference.md
* updates to README.md to accommodate both of the above
* an improved error message in lib/cluster/runners/op.js to help debug custom processors

kstaken:
This needs a rebase before it can be merged.

godber:
Oh, sorry, I should have rebased.  But the merge conflict is resolved.

PR #432, opened 2017-06-21
fixed job_service race condition on startup

jsnoble:


lib/cluster/services/jobs.js
```diff
@@ -181,15 +181,36 @@ module.exports = function(context, cluster_service) {
...
+                    logger.error(`could not findJob`, errMsg);
+                    return Promise.reject(errMsg);
+                })
+        }
+        else {
+            return new Promise(function(resolve, reject) {
+                var cycle = setInterval(function() {
+                    if (ex_store) {
+                        clearInterval(cycle);
+                        ex_store.search(query, null, 10000)
```
kstaken: The same 10 lines occurs in 2 places. Is there some reason this can't be DRY?

PR #433, opened 2017-06-21
added queue length to be configurable

jsnoble:


lib/config/schemas/system.js
```diff
@@ -122,6 +122,20 @@ var schema = {
             }
         }
     },
+    slicer_queue: {
+        doc: 'time in milliseconds that the slicer will wait for worker connection before terminating the job',
```
kstaken: This isn't named well. Should probably be slicer_queue_length. All the validation doc and error messages are also incorrect. 

Also why are we accruing all these custom formats to validate numbers? Doesn't convict have that ability it self?

jsnoble: made some of the changes, the innate validation is Number, but it basically just checks that is !== NaN. However we need to make sure that it cant be zero either, if it was then the slicer would never run

PR #439, opened 2017-06-30
WIP Child services

jsnoble:
This separates all the major functionalities into separate processes

- [ ] Improve error handling

- [ ]  verify worker allocation behavior

- [ ] Add ability to prevent workers and slicers to be put on cluster_master


jsnoble:
We will revisit this in the future

PR #453, opened 2017-07-11
adding a noop processor

godber:
To support an upcoming change in the teraslice test harness I have added a `noop` operator/processor for teraslice.  There are no changes to other teraslice code so this can probably just be accepted


godber:
The reason I needed something like this is because all teraslice job definitions require at least two operations.  The test framework will make use of this `noop` to satisfy this requirement.

godber:
Merging

PR #455, opened 2017-07-13
Extract validateOperation from validators.job

godber:
To support development of the Teraslice Test harness, specifically
https://github.com/terascope/teraslice_op_test_harness/issues/2

I have extracted `validateOperation`.  It is renamed to
`validateConfig` since it handles both jobConfigs and opConfigs.

I have also removed the third argument, which was a boolean
that indicated whether the config passes was an operation or not.
I replaced it with a test for the presence of `._op` which should
be present on all operations.

**This is a work in progress.  If we want to go this route, a future commit should modify `lib/config/validators/job.js` to use the function added here.**

### DIFF
--- a/lib/config/validators/config.js
+++ b/lib/config/validators/config.js
@@ -1,0 +1,39 @@
+var _ = require('lodash');
+var convict = require('convict');
+var convictFormats = require('../../utils/convict_utils');
+var commonSchema = require('../schemas/job').commonSchema();
+
+// Add custom teraslice formats
+convictFormats.forEach(function(format) {
+    convict.addFormat(format);
+});
+
+/**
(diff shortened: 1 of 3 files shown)


godber:
This fails above because `workers` is machine specific, I'll have to change my assertion.

godber:
The changes in this PR are used by https://github.com/terascope/teraslice_op_test_harness/pull/3

godber:
I had a moment of weakness where I tried just moving the `convict.addFormat(format)` up higher into the global module scope so that `validateOperation` would work right.  Then I hit other issues with context that convinced me that decoupling that function from the job validator was the best course of action.

Regardless, this works, there are very simple specs for `job.js` and `config.js` now.

PR #461, opened 2017-07-20
Fixes #451: validate connectors

### LINKED ISSUE #451: Validate job connections on job submission
It doesn't appear that connections are tested when a job is submitted.  So if you fat finger a connection name you end up with failed slices.  It should be possible to validate all connections at job submission time.

erik-stephens:


spec/config/validators/config-spec.js
```diff
@@ -41,3 +41,41 @@ describe('When passed a valid jobSchema and jobConfig', function() {
         expect(jobConfig).toEqual(validJob);
     });
 });
+
+describe('When passed a job without a known connector', function() {
```
godber: I'd probably move this to it's own file: `spec/config/schemas/job-spec.js`.

I guess it's really a question of how to structure a unit test suite, and I think mirroring the code directory structure under `spec/` is a pretty straight forward way to do that.

erik-stephens: On 2nd thought, I think the same spec file makes sense since I'm testing against validators/config.js and not validators/job.js.

godber: Yeah, you're right.

godber:
fixes #451

kstaken:
Is this still WIP?

erik-stephens:
No, I think just waiting for someone to bless it.

PR #464, opened 2017-07-21
first pass on event passing to ops resolves #460

### LINKED ISSUE #460: Events for _op processing
With the kafka stuff there are two different scenarios that have cropped up where the reader needs to hook into the process.

  1. When a job stops the kafka reader needs to be able to receive notification of that so it can stop waiting for data and immediately complete the slice and the process can exit.
  2. When the job is running the kafka reader currently moves the offset as the data is read. What needs to happen in that case is that the reader needs to read the data but the offset shouldn't move until the slice has been successfully processed. This requires an event at the end of the slice.

These are just a couple concrete examples though. There will be others so I want to outline what we think the reasonable events might be. From the two use cases above I see:

 * job:stopped or job:stopping
 * slice:success <- only triggered on success
 * slice:failure <- only triggered on failure
 * slice:finalize <- always called when a slice is complete regardless of outcome.

So what other scenarios make sense?

Maybe
 * job:paused
 * job:resume
 
Other slice states?

jsnoble:


jsnoble:
this needs to be used in conjunction with the changes in terafoundation

godber:
@jsnoble these changes right? 

https://github.com/terascope/terafoundation/pull/89

jsnoble:
yes

kstaken:
Unit tests are failing on this now.

jsnoble:
tests pass now

Issue #465, opened 2017-07-22
Expose version & some configuration details via api


erik-stephens:
As an operator of teraslice, would be nice to be able to easily query the version and location of storage via the api.  Endpoint isn't import but some ideas: /about or even the root page ala elasticsearch.

jsnoble:
Im not sure what you mean by location of storage, do you mean if writiing to a file what the file path is, or if sending to elasticsearch then the index? as for the version, its being listed in /cluster/state. Each node lists what version of teraslice it is using

godber:
I think Erik means which ES server stores the state and other indices, probably should provide those index names too.

Ah, yes, @erik-stephens look at `/cluster/state` it shows the version https://github.com/terascope/teraslice/blob/master/docs/api.md#get-clusterstate

Thanks Jared.

godber:
Is is a requirement that all members of a teraslice cluster have the same nodejs and teraslice versions?  If so is this enforced in any way?

godber:
Actually, I think what we need is a `/cluster/config` endpoint that provides access to some, if not all, of the `terafoundation.yaml` information.  Without this, we end up building operational solutions that require access to that file, which ends up being a lot of hoop jumping.

godber:
I am just killing this discussion for now.  There are likely to be much broader changes to the configuration mechanisms of teraslice that would render this moot anyway.  Everyone is aware of the needs and risks of this.

PR #468, opened 2017-07-22
Issue#465 expose config api

godber:
Adding `/cluster/config` API endpoint that exposes `context.sysconfig`.

This PR contains a second commit that just updates style things.

refs #465

### DIFF
--- a/lib/cluster/services/api.js
+++ b/lib/cluster/services/api.js
@@ -20,10 +20,11 @@ module.exports = function(context, app, services) {
         }
     }));
 
-    app.use(function(err, req, res, next){
+    app.use(function(err, req, res, next) {
         if (err instanceof SyntaxError) {
             sendError(res, 400, `the json submitted is malformed`);
-        } else {
+        }
+        else {
             next();
         }
     });
@@ -45,7 +46,7 @@ module.exports = function(context, app, services) {
                 sendError(res, 400, '/asset endpoints do not accept json');
                 return;
             }
-            redirect(req, res)
+            redirect(req, res);
         });
 
     app.route('/txt/assets*')
@@ -53,7 +54,8 @@ module.exports = function(context, app, services) {
 
 
     app.post('/jobs', function(req, res) {
-        //if no job was posted an empty object is returned, so we check if it has values
+        // if no job was posted an empty object is returned, so we check if it
+        // has values
         if (!req.body.operations) {
             sendError(res, 400, 'No job was posted');
         }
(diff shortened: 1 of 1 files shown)


godber:
Accessing the new API endpoint like this

```bash
curl -Ss localhost:5678/cluster/config
```

returns this:

```json
{
    "terafoundation": {
        "environment": "development",
        "log_path": "/Users/godber/Workspace/terascope/opensource/teraslice",
        "logging": [
            "console"
        ],
        "log_level": "info",
        "log_buffer_limit": 30,
        "log_buffer_interval": 60000,
        "workers": 4,
        "connectors": {
            "elasticsearch": {
                "default": {
                    "host": [
                        "127.0.0.1:9200"
                    ],
                    "sniffOnStart": false,
                    "sniffOnConnectionFault": false,
                    "requestTimeout": 120000,
                    "deadTimeout": 30000,
                    "maxRetries": 3,
                    "__reused": true,
                    "log": {
                        "_events": {},
                        "_eventsCount": 3
                    },
                    "hosts": [
                        "127.0.0.1:9200"
                    ],
                    "keepAlive": true,
                    "minSockets": 10,
                    "maxSockets": 11
                }
            }
        }
    },
    "teraslice": {
        "ops_directory": "/app/source/examples/ops/",
        "assets_directory": "/Users/godber/Workspace/terascope/opensource/teraslice/assets",
        "shutdown_timeout": 60000,
        "reporter": "",
        "hostname": "10.167.0.1",
        "workers": 4,
        "master": true,
        "master_hostname": "127.0.0.1",
        "port": 5678,
        "name": "teracluster",
        "state": {
            "connection": "default"
        },
        "timeout": 300000,
        "slicer_timeout": 180000,
        "slicer_queue_length": 10000,
        "slicer_port_range": "45679:46678",
        "moderator": false,
        "moderator_limit": 0.85,
        "moderator_resume": 0.5,
        "moderator_interval": 10000
    },
    "_nodeName": "stretch.local.1"
```

godber:
This particular implementation could be risky if it is possible for the connection definitions to contain authentication information.

erik-stephens:
That risk was my only concern with this.  I suspect there will be a relatively small subset of the config that will warrant being exposed, so I vote for having the endpoint explicitly define what parts of the config are exposed.

kstaken:
Yeah this seems both very helpful and highly problematic at the same time. It's definitely possible for connectors to define auth info. 

It's also only exposing the config for the master node and the config on the workers could be different.

Obviously if you think about the requirements for a future GUI it's going to be useful to have access to almost all of this information. 

So I guess I have no answers.

godber:
We need to think this through more.

PR #477, opened 2017-07-27
Moderator docs

jsnoble:
- Improved documentation
- Ensured consistent moderator key values listed on jobs

### DIFF
--- a/lib/cluster/services/jobs.js
+++ b/lib/cluster/services/jobs.js
@@ -322,7 +322,7 @@ module.exports = function(context, cluster_service) {
 
     function submitJob(job_spec, shouldRun) {
         return ensureAssets(job_spec)
-            .then(function(stuff) {
+            .then(function(_job_spec) {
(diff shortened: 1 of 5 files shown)


lib/cluster/services/jobs.js
```diff
@@ -322,7 +322,7 @@ module.exports = function(context, cluster_service) {
 
     function submitJob(job_spec, shouldRun) {
         return ensureAssets(job_spec)
-            .then(function(stuff) {
+            .then(function(_job_spec) {
```
kstaken: Is this argument actually used for anything? Looks strange.

jsnoble: it originally didn't but i wanted to make it more semantic about whats coming out of it. Though its not used here, i am using for the api change you mentioned in the other ticket about allowing the job to maintain the original asset names and not be automatically changed to their id's.

PR #483, opened 2017-08-01
Fixes #482: integration tests moved to this repo

### LINKED ISSUE #482: Incorporate integration tests
Integrate the following integration test suite into teraslice:

https://github.com/terascope/teraslice-integration-tests

They can include the ES5.5.0 changes from here https://github.com/terascope/teraslice-integration-tests/issues/14

erik-stephens:
The tests have been consistently passing for me.  Every once in a while one of the cluster state tests will fail.  This PR includes changes to how the tests get references to things like elasticsearch clients - moved to memoized functions in a `misc` module (couldn't think of a better name).


integration-tests/spec/setup.js
```diff
@@ -0,0 +1,158 @@
...
+                        '_op': 'elasticsearch_index_selector',
+                        'index': indexName,
+                        'type': 'events'
+                    },
+                    {
+                        '_op': 'elasticsearch_bulk',
+                        'size': 5000
+                    }
+                ]
+            }
```
erik-stephens: Not proud of this function, especially how an unhandled exception won't get caught by the caller.  I used the callback interface for the stats() since we want to recreate index in two cases: a) index doesn't exist or b) doesn't have correct count.  Couldn't figure out a good way to do that with promise interface.

Issue #487, opened 2017-08-04
integration test framework doesn't catch changes in dependencies


godber:
The integration test framework needs to build docker images with the `--no-cache` option to include updates in teraslice dependencies (like terafoundation).

PR #489, opened 2017-08-04
Docker build now uses --no-cache

godber:
* docker build will now "find" changed dependencies
  because it no longer uses the docker build cache
* add clean-all target that removes the node_modules
  directory and the images.log.
* changed wording on clean help to better reflect
  what it actually does

closes: #487


godber:
After applying these changes, updating terafoundation, then running:

```
make clean-all
make test
```

All of these tests will pass:

screen shot 2017-08-04 at 6 19 12 am

PR #493, opened 2017-08-06
Fix #492: improve dev workflow

### LINKED ISSUE #492: Improve developer workflow
The integration tests require a rebuild of teraslice docker image after each edit.  Plus, npm-linked modules will not work either.  Update the README to make it more clear what the intended workflow is.

### LINKED ISSUE #495: Back port test changes
This PR will need to be incorporated into the local integration tests once the issues are resolved.

https://github.com/terascope/teraslice-integration-tests/pull/17/files

erik-stephens:
To tackle the `npm link` issue, I bind-mount any modules that are linked.  The teraslice project root is bind-mounted as well to avoid rebuilds after each edit.

To tackle the developer vs QA/CI concerns, I introduce a `MODE` environment variable.  This will determine things like the Dockerfile and mounts.  The differences are pretty small but important in a couple of ways:

1. Wanted to let docker-compose handle the building of the images as needed since doing that dependency tracking via make is not as elegant.

1. There isn't a way to specify `--no-cache` in docker-compose.yml, so I experimented with a recipe that lets us take advantage of dockerfile caching yet ensure we're using correct dependencies (assuming versioned package.json - looking at you git dependencies).

I suspect that last point will be a bit controversial but I think it worth considering.  As we move to more automatic image builds, having to do full npm installs each time will be a drag.

Fixes #495.


godber:
I was able to get working directory teraslice changes to show up in the integration tests but not for terafoundation.  Not sure what the issue was and I think I am out of time.

erik-stephens:
Fixed an issue where default was `MODE=dev ` (notice tailing space).  Local changes should show up in the test image now.

godber:
@erik-stephens could you rebase this branch?  And then @jsnoble make sure it works and makes sense to you and either provide feedback or merge.

erik-stephens:
Fixes #495

Issue #502, opened 2017-08-15
Add asset support to script processor


jeffmontagna:

kstaken:
This has been completed via #503

Issue #494, opened 2017-08-07
Implement script example


godber:
Teraslice includes a basic external script interface, it should have an example in the `examples/` directory.

https://github.com/terascope/teraslice/blob/master/docs/ops-reference.md#script

I will try and think of a meaningful example, but if anyone else thinks of one in the meantime feel free to share.

kstaken:
This is being addressed in #503

kstaken:
@jeffmontagna can this be closed?

jeffmontagna:
yes

kstaken:
Resolved in #503

PR #503, opened 2017-08-15
added support for assets to script processor

jeffmontagna:


lib/processors/script.js
```diff
@@ -79,11 +90,16 @@ function schema() {
         options: {
             doc: 'Obj containing options to pass into the process env',
             default: {}
+        },
+        asset: {
```
erik-stephens: Looks like it's `asset_name` everywhere else.

godber:
refs: #502

lib/processors/script.js
```diff
@@ -1,16 +1,27 @@
...
+
     var args = opConfig.args;
     var options = opConfig.options;
-
     var logger = jobConfig.logger;
 
+    if (opConfig.asset_name == undefined || opConfig.asset_name == '') {
+        command = opConfig.command;
+    } else {
+      context.assets.getPath(opConfig.asset_name).then(function(apath) {
```
jsnoble: this is an async call, and you are getting the data correctly, however as it stands now this adds in a race condition. The promise is called, but the op immediately returns the main function to start processing. It is possible for the `command` variable to not be set by the time its called which will cause problems. Assuming the pr for async op is pulled in, this op will need to be similar to how the kafka_reader op initializes (the pr for it at least)

kstaken:
I'm assuming this still needs updates based on @jsnoble's comments. The dependent PR #500 has been merged.

docs/ops-reference.md
```diff
@@ -296,20 +296,22 @@ Example configuration
 ```
 {
      "_op": "script",
-     "command": "python",
-     "args": ["someFile.py", "-someFlag"],
+     "command": "someFile.py",
+     "args": ["-someFlag1", "-someFlag2"],
+     "asset_name": "someAsset",
```
kstaken: This should probably just be named `asset`

lib/processors/script.js
```diff
@@ -1,16 +1,27 @@
...
+        command = path.join(apath, opConfig.command);
```
kstaken: Looks like this is 2 space indentation in a file where everything else is 4 spaces.

jsnoble:
It would be great if documentation for this script could be updated. Also in relation to this pr is https://github.com/terascope/teraslice/issues/494 . Could you add a example in the example directory of creating and using a scipt

kstaken:
What is holding this in WIP?

jeffmontagna:
removed WIP.  Hold up was on the changes for the updated test harness. Forgot to removed WIP on Friday.

kstaken:
I'm getting a failing unit test. Fails repeatedly for me.

```
   ✗ data out is the same as data in (arrayLike) (5 secs)
      - Failed: processing script stdout pipe
          at /Users/kstaken/projects/opensource/test/teraslice/spec/processors/script/script-spec.js:114:17
          at run (/Users/kstaken/projects/opensource/test/teraslice/node_modules/mocker-data-generator/node_modules/babel-polyfill/node_modules/core-js/modules/es6.promise.js:66:22)
          at /Users/kstaken/projects/opensource/test/teraslice/node_modules/mocker-data-generator/node_modules/babel-polyfill/node_modules/core-js/modules/es6.promise.js:79:30
          at flush (/Users/kstaken/projects/opensource/test/teraslice/node_modules/mocker-data-generator/node_modules/babel-polyfill/node_modules/core-js/modules/_microtask.js:18:9)
          at nextTickCallbackWith0Args (node.js:489:9)
          at process._tickCallback (node.js:418:13)
      - Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
```

jeffmontagna:
I re-ran tests on my laptop and the tests are passing.  They did pass CI too.  Maybe the test harness needs to be updated to the latest version?

kstaken:
Not sure why it doesn't fail for you but in that test the order of context and opConfig are still reversed. Also in the clauses where you have an async test and a call to fail() you also need to call done. Making those tests have a finally() block that calls done is probably the best way to do it.

PR #507, opened 2017-08-23
functionality to restart workers periodically resolves #498

### LINKED ISSUE #498: Periodically restart workers to accommodate leaky connectors
Some connectors may have more memory leaks than others and accumulate memory after running for long periods.  A common way to address this is to discard workers and create a new worker after a period of time.  This will reduce the accumulated leaked memory.

jsnoble:


lib/config/schemas/job.js
```diff
@@ -137,6 +137,22 @@ function jobSchema(context) {
                     }
                 }
             }
+        },
+        recycle_worker: {
+            doc: 'The number of slices a worker processes before it reinstantiate itself and clears its memory',
```
kstaken: "The number of slices a worker processes before it exits and restarts."

PR #509, opened 2017-08-23
added memory statistics between each op resolves #508

### LINKED ISSUE #508: Track worker memory usage
We determined to add to analytics index the memory usage before and after each op

jsnoble:


lib/utils/analytics.js
```diff
@@ -89,10 +108,17 @@ function analyzeStats(logger, operations, obj) {
...
 
+    var memory = obj.memory.map(function(arr) {
+        return calculateStats(arr);
+    });
+
     time.forEach(function(data, index) {
-        logger.info('operation ' + operations[index]._op + '\n average completion time of: ' + data.average + ' ms, ' +
-            'min: ' + data.min + ' ms, and max: ' + data.max + 'ms \n' +
-            ' average size : ' + size[index].average + ', min: ' + size[index].min + ', and max: ' + size[index].max)
+        logger.info(`
```
erik-stephens: How about something like `logger.info(msg, field1=val1, field2=val2)`?  Benefits:

- We don't have to futz with formatting - can leave that to bunyan.
- Easier to mine logs if already in structured format - no need for brittle regex matching.

kstaken:
This needs a rebase

PR #515, opened 2017-08-25
removed duplication between once and persistent fns resolves #420

### LINKED ISSUE #420: refactor slicer.js - remove duplication between persistent and once
The functions `persistent` and `once` in `lib/cluster/slicer.js` are nearly identical, they can probably be refactored to eliminate the duplication.

Given the criticality of these functions, this duplication is a significant risk.

screen shot 2017-05-31 at 8 34 16 am

jsnoble:


kstaken:
The integration tests hang at the end of the first test. I spent a huge amount of time thinking that the integration tests were broken when it appears to be this branch. Nothing shows up in the logs, it just looks like the first job runs but then doesn't end correctly and the tests appear to hang.

PR #517, opened 2017-08-29
WIP: adding job lifecycle tests to test persistent jobs

godber:
This PR tries to test the job lifecycle behaviors of once and persistent jobs.  It does so by running essentially the same job, but letting the once job run to completion and letting the persistent job run a little longer and expecting the persistent job to end up with more docs than the once job.

In the persistent job I use `.delay(5000)` and then stop the job ASSUMING that it has run long enough to end up with more than 1000 docs.  I suppose there could be cases where that is not enough time.  It seems unlikely, but when I used a delay of 3000, the test actually failed once.

It might be better if I just looped and used a shorter delay and checked the doc count and just wait in that loop until the doc count exceeds 1000.  The test would be more reliable this way.

### DIFF
--- a/integration-tests/spec/cases/jobs/lifecycle.js
+++ b/integration-tests/spec/cases/jobs/lifecycle.js
@@ -1,0 +1,66 @@
+'use strict';
+
+let misc = require('../../misc')();
+let teraslice = misc.teraslice;
+
+module.exports = function() {
+    describe('A once job should', function() {
+        fit('generate a 1000 docs and complete', function(done) {
+          let jobSpec = misc.newJob('gen');
+
+          teraslice().jobs.submit(jobSpec)
+            .then(function(job) {
+              return job
+                .waitForStatus('running')
+                .then(function() {
+                  return job.waitForStatus('completed');
+                })
+                .then(function() {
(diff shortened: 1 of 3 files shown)


godber:
Obviously the `fit`s need to be changed to `it` before merging.

godber:
Actually, the `delay(5000)` is not a long enough delay on my laptop, so this test fails out of the gate for me.

erik-stephens:
How about defining a waitForCount() similar to job.waitForStatus() of teraslice-client-js?  Not sure best place to define, spec/misc.js is only thing that comes to mind.

godber:
I am looking at `waitForValue()` right now ... https://github.com/terascope/teraslice/blob/master/integration-tests/spec/wait.js#L29

So yeah, something along those lines.

godber:
Not sure `forValue()` would work as is since it's condition is `===` 

https://github.com/terascope/teraslice/blob/master/integration-tests/spec/wait.js#L39

The test would have to be "lucky" and catch the count at the threshold value.

godber:
I'm just going to close this, some day someone will complete it.

PR #522, opened 2017-09-05
fixed stop job timeout issues resolves #434

### LINKED ISSUE #434: Timeouts on stop job
When a job is stopped and the workers take some time to exit you get the following error.
```
curl -XPOST localhost:5678/ex/56dea3e4-820c-4508-a946-5704ff273675/_stop
{
    "error": 500,
    "message": "Could not stop job, error: Error communicating with node: localhost, Could not send msg: cluster:job:stop, data: {\"ex_id\":\"56dea3e4-820c-4508-a946-5704ff273675\",\"_msgID\":\"SJQofzdQb\"}"
}
```

This appears to be the result of a timeout but I think there are several issues here.
1. The error message is misleading. If this is a timeout it should probably be more explicit about it.
2. The default timeout is probably too short. What is the impact from increasing it to something like 5m?
3. The job does appear to stop and all workers exit however it is not labeled as `stopped` and remains in it's prior state. Immediately re-running `_stop` will work and set the state to `stopped` even though the job doesn't appear to actually be running anymore.

jsnoble:
The issue was found to reside in node_master code regarding the stop job event. A stop job event will notify each node_master and will expect a response back from each node indicating that it stop. In the case of the node where the actual slicer resides, the slicer handles the reply, but on other nodes where the workers reside, the node_master has special logic to reply itself when it sees all the appropriate child processes are gone. The issue happens in how the node_master monitors. If there is another slicer not relevant to the job being stopped on the same node as regular workers that should shutdown, the node_master would not properly execute the special logic, thus leading the job to hang even though it should have finished

### DIFF
--- a/integration-tests/spec/cases/cluster/job-state.js
+++ b/integration-tests/spec/cases/cluster/job-state.js
@@ -1,0 +1,58 @@
+'use strict';
+
+var _ = require('lodash');
+var Promise = require('bluebird');
+var misc = require('../../misc')();
+var wait = require('../../wait')();
+
+module.exports = function() {
+    var teraslice = misc.teraslice();
+    
+    describe('worker allocation', function() {
+
+        it('should cycle through after state changes with other jobs running', function(done) {
+            var job_spec1 = misc.newJob('generator');
+            var job_spec2 = misc.newJob('generator');
+            var job1_id;
+            var job2_id;
+            job_spec2.operations[1].name = 'second_generator';
+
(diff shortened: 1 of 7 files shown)


godber:
This appears to work, I am not convinced that the new integration test really quite captures the failure scenario they way we might like.  In the event that the included fix is not present, those integration tests just hang as far as I can tell.  Either that or the timeout for them failing is very long.  I did manually check that I can see the `_stop` command hanging without the fix ... and then completing when I put the fix back in place.

PR #528, opened 2017-09-07
Refactoring the slice allocation logic to remove duplication that has…

kstaken:
… been bugging me


kstaken:
Tests pass. Merging so we can do a new build.

PR #529, opened 2017-09-07
Controlling the slicer queue length on data generator

kstaken:


kstaken:
Tests pass and queue is behaving as expected.

PR #531, opened 2017-09-08
adding istanbul 2.0 code coverage to default test command

### LINKED ISSUE #530: Add code coverage to test report
We should at least be aware of our test coverage, so lets just add it to our test output for now.

godber:
This changes out `npm test` command to use istanbul/nyc to check the code coverage of our tests.  The coverage report will be shown right after the test output by default.

It looks like this:

screen shot 2017-09-08 at 5 41 44 am

closes #530

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ build/Release
 # https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git
 node_modules
 assets/*/
+.nyc_output/
--- a/package.json
+++ b/package.json
@@ -5,7 +5,10 @@
   "main": "service.js",
   "scripts": {
     "lint": "eslint lib",
-    "test": "nyc -x spec/ jasmine"
+    "jasmine": "jasmine",
+    "test": "nyc -x spec/ --reporter=text-summary jasmine && nyc report --reporter=html",
+    "testcov": "nyc -x spec/ jasmine && nyc report --reporter=html",
+    "showcov": "open coverage/index.html"
   },
   "repository": {
     "type": "git",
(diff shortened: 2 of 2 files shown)


godber:
The only think I don't like about this particular report is that it doesn't provide a simple high level summary.  Comparing the output of one run to another is too hard ... you cant easily tell which direction you've gone.

erik-stephens:
Agreed.  Assuming `nyc report --reporter=json-summary` was run previously, we can do something like:
```
$ ./node_modules/.bin/nyc report --reporter=json-summary --report-dir coverage/now

$ ./node_modules/.bin/istanbul-diff coverage/coverage-summary.json  coverage/now/coverage-summary.json
Coverage increased +16% (49) lines. Sensational. (total)
```

Persisting/committing the report so it can be compared to later might not be worth the hassle.  Since this repo is public, might be better to use something like coveralls.io.

godber:
I think coveralls.io is a good next step but is more than I personally want to tackle at the moment.  Maybe next week.

I did find a good summary output and I think it makes sense to include it in the default test run.  The output looks like this:

screen shot 2017-09-08 at 8 18 31 am

And I think the following npm scripts should cover our short term needs:

* `npm test` - runs unit tests with coverage enabled and gives summary, then generates html report under `coverage/`
* `npm run jasmine` - runs **just** the unit tests, just in case something blows up or you don't want to wait for the coverage stuff to run
* `npm run testcov` - runs unit tests with coverage enabled and gives complete text output, then generates html report under `coverage/`
* `npm run showcov` - assumes that `test` or `testcov` have been run and runs `open coverage/index.html`

godber:
I am going to go ahead and accept this since it's an obvious win and extending it with coveralls can be done easily after the fact.

Everyone will need to run `npm install` again to get the new dependency otherwise `npm test` will fail.

PR #532, opened 2017-09-08
Worker disconnect timeout

jsnoble:
config changes:
- timeout => network_timeout
- worker_disconnect_timeout //defaults to 5min


godber:
I have run the integration tests three times on this PR ... one test failed once:

```
    cluster state
      ✓ should match default configuration (0.007 sec)
      ✓ should update after adding and removing a worker node (4 secs)
      ✓ should update after adding and removing 20 worker nodes (12 secs)
      ✓ should be correct for running job with 1 worker (4 secs)
      ✗ should be correct for running job with 3 workers (4 secs)
        - Failed: Job has status: "failed" which is terminal so status: "completed" is not possible.
            at tryCatcher (/home/godber/Workspace/terascope/teraslice/integration-tests/node_modules/bluebird/js/release/util.js:16:23)
            at Promise._settlePromiseFromHandler (/home/godber/Workspace/terascope/teraslice/integration-tests/node_modules/bluebird/js/release/promise.js:512:31)
            at Promise._settlePromise (/home/godber/Workspace/terascope/teraslice/integration-tests/node_modules/bluebird/js/release/promise.js:569:18)
            at Promise._settlePromise0 (/home/godber/Workspace/terascope/teraslice/integration-tests/node_modules/bluebird/js/release/promise.js:614:10)
            at Promise._settlePromises (/home/godber/Workspace/terascope/teraslice/integration-tests/node_modules/bluebird/js/release/promise.js:689:18)
            at Async._drainQueue (/home/godber/Workspace/terascope/teraslice/integration-tests/node_modules/bluebird/js/release/async.js:133:16)
            at Async._drainQueues (/home/godber/Workspace/terascope/teraslice/integration-tests/node_modules/bluebird/js/release/async.js:143:10)
            at Immediate.Async.drainQueues [as _onImmediate] (/home/godber/Workspace/terascope/teraslice/integration-tests/node_modules/bluebird/js/release/async.js:17:14)
            at processImmediate [as _immediateCallback] (timers.js:396:17)
```

It passed the other two times, so I am going to assume the test is unreliable.

godber:
Using the following test job, which sets `"recycle_worker": 10`:

```
{
    "name": "Generate data then send it to noop",
    "lifecycle": "persistent",
    "workers": 1,
    "recycle_worker": 10,
    "operations": [
        {
            "_op": "elasticsearch_data_generator",
            "size": 20000
        },
        {
            "_op": "noop"
        }
    ]
}
```

I have manually verified the following things:

* workers recycle without the job/slicer terminating

Here is the first worker, note the PID

first-worker

Here is the second worker, after two slices, note the PID

second-worker

* some time (maybe 60s to 90s) after killing the only worker on the test job the slicer stops and marks the job as `failed`.  I expected it to fail faster.

I set the master configuration to `worker_disconnect_timeout: 30000` as follows:

```
terafoundation:
    environment: 'development'
    log_level: info
    connectors:
        # ***********************
        # Elastic Search Configuration
        # ***********************
        elasticsearch:
            default:
                host:
                    - "elasticsearch:49200"

teraslice:
    ops_directory: '/app/source/examples/ops/'
    workers: 8
    master: true
    master_hostname: "127.0.0.1"
    port: 45678
    name: "teracluster"
    worker_disconnect_timeout: 30000
```

godber:
Now that I think about it I didn't set `worker_disconnect_timeout` in the worker config.  But perhaps that doesn't matter.  This has always been confusing for me actually.  What gets set where and when it matters.

kstaken:
If the slicer was allocated on the worker then the default would have been used.

godber:
If that's the case then I am pretty sure that the slicer died too soon.  But not too soon to kill the job.

godber:
Ok, I have confirmed that with the `worker_disconnect_timeout` set to 30000ms in the right place ... that the slicer will die and mark the job as `failed` after about 30s.

godber:
After reverting the worker and master configs back to the default, the slicer dies and marks the job `failed` right at 5 minutes.  I must be seeing ghosts.

godber:
One other thing to note is that while doing the above manual tests I ran the integration tests at least three more times without failure.  I think this MR is good.

PR #548, opened 2017-09-14
refactored makeLogger call to use single config object resolves #543

### LINKED ISSUE #543: logger instantiation refactor
there are several spot where creating has extra data thats not needed to be created, as well as some improper module name collision (worker logger vs worker new slice logger)

jsnoble:
for the logger that is generated per each slice, I have elected to set the module name for that logger the slice


lib/cluster/assets.js
```diff
@@ -10,7 +10,7 @@ var parseError = require('../utils/error_utils').parseError;
 
 // this is a child process spawned by node_master to download assets
 module.exports = function(context) {
-    var logger = context.foundation.makeLogger('asset_manager', 'asset_manager', {module: 'asset_manager'});
```
godber: So many of the changes here are "like" this one ... removing the first two arguments that are strings.  Was there an API change or what?

jsnoble:
yes, a while back in terafoundation, it is required for the main logger but
any child logger does not need to

godber:
The integration tests pass on this ... and here is the output from the only tangible change:

```
[2017-09-15T23:59:49.189Z]  INFO: teraslice/591 on d67dbb8e823c: (ex_id=80703533-948c-4ed5-be81-666d9dca3cab, module=slice, worker_id=172.21.0.4__59, slice_id=c92fa326-4f51-4263-9a62-783811c9c415)
```

PR #553, opened 2017-09-18
Make forValue return when it fails

godber:
forValue shouldn't schedule itself to run again unless result isn't
equal to value AND counter does not exceed iterations.


godber:
screen shot 2017-09-18 at 11 05 12 am

godber:
Passed

PR #561, opened 2017-09-21
fixed schema validation and faulty regex

jsnoble:
i switched the regex to /(\d+)(\D+)/i which looks for numbers next to non-number characters. It looks like the schema failed becuase it did not split the numbers from the time selectors when it was being passed to dateOptions. We really need to test our schema seperate. the unit tests did not cover schema validations, we should test those in the future


kstaken:
Persistent jobs still fail with this change.

```
Error on slicer initialization, will attempt to retry 3 times: TypeError: end.format is not a function\n    at divideRange (/Users/kstaken/projects/opensource/teraslice/lib/readers/elasticsearch_date_range/slicer.js:380:47)\n    at getTimes (/Users/kstaken/projects/opensource/teraslice/lib/readers/elasticsearch_date_range/slicer.js:394:25)\n
```

kstaken:
integration and manual tests pass and persistent jobs will now run.

Issue #569, opened 2017-09-26
Assets don't deploy correctly on a cluster


kstaken:
The first run of a job almost always fails because assets aren't fully downloaded before the job tries to run. Running the job again typically works so this is just a synchronization issue.

jsnoble:
fixed under #570

PR #570, opened 2017-09-26
fixed asset checking fix on initialization, removed unneeded store cr…

jsnoble:
…eation


jsnoble:
Fixes #569, ran intergration tests alongside @kstaken

PR #582, opened 2017-10-10
cleaned up slicer and worker shutdown, fixed syntax error and error h…

### LINKED ISSUE #576: long running jobs get marked failed on the last slice
We've been running some id_reader jobs that run for about 24 hours and then get marked `failed` on the final slice when the job is completed. They're in `running` up to the end and there aren't any `error` slices so it seems to be some kind of issue that occurs as the job is finishing. Not entirely sure this is reader related but seems likely since the rest of the job is pretty normal stuff.

jsnoble:
…andling in job complete event resolves #576

### DIFF
--- a/lib/cluster/slicer.js
+++ b/lib/cluster/slicer.js
@@ -316,7 +316,7 @@ module.exports = function(context) {
                         })
                         .catch(function(err) {
                             const errMsg = parseError(err);
-                            logger.error(`execution ${ex_id} was not able run the job complete sequence and update its status, slicer will soon exit, error: ${errMsg}`);
+                            logger.error(`execution ${ex_id} has run to completion but the process has failed while updating the execution status, slicer will soon exit, error: ${errMsg}`);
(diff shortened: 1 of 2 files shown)


kstaken:
This needs a rebase

lib/cluster/slicer.js
```diff
@@ -314,9 +313,14 @@ module.exports = function(context) {
                             }
 
                             messaging.send(msg);
+                        })
+                        .catch(function(err) {
+                            const errMsg = parseError(err);
+                            logger.error(`execution ${ex_id} was not able run the job complete sequence and update its status, slicer will soon exit, error: ${errMsg}`);
```
kstaken: There's a typo in this message. "was not able run the job"

Would this be correct to say?

execution ${ex_id} has run to completion but the process has failed while updating the execution status, slicer will soon exit, error: ${errMsg}

kstaken:
Integration and manual tests pass.

PR #585, opened 2017-10-10
assets store registry of its apis is now available for worker and sli…

### LINKED ISSUE #583: asset store api loading
worker and slicer had once loaded the assets store without using it, it was subsequently removed, however it was being used to register the getAssets api at context.apis.getAssets. The removal of this store needs to be reverted and comments left in place about why the assets store is being loaded

jsnoble:
…cer resolves #583 

api should be avaiable at context.apis.assets.getPath


kstaken:
Integration tests pass.

PR #592, opened 2017-10-24
fixed elasticsearch slicer recursion algorithm resolves #591

### LINKED ISSUE #591: elasticsearch slicer recursion fix
There is an issue with the algorithm of the elasticsearch slicer which would improperly divide the time when the time diff got low enough that it would prevent further recursion down to 1 s/ms resolution.

jsnoble:


kstaken:
Integration tests pass.

PR #578, opened 2017-10-06
fixed issues where a query with no results threw an error resolves #577

jsnoble:

PR #594, opened 2017-10-26
fix job with supplied start failure by converting str to moment

### LINKED ISSUE #593: date range slicer error
Between 0.16.0 and 0.17.0 some range jobs started throwing slicer errors, 
(from 0.17.0),

Could not initialize slicers, error: TypeError: updatedOpConfig.start.format is not a function\n    at updateJob (/opt/teraslice/lib/readers/elasticsearch_date_range/slicer.js:161:59)

The error shows up in 0.22.0 as well.

twkiel:
Resolves #593 where if a job is supplied the start parameter it will fail

### DIFF
--- a/lib/readers/elasticsearch_date_range/slicer.js
+++ b/lib/readers/elasticsearch_date_range/slicer.js
@@ -159,7 +159,7 @@ function newSlicer(context, opConfig, job, retryData, logger, client) {
             var operations = JSON.parse(process.env.job).operations;
             operations.shift();
             var updatedOpConfig = Object.assign({}, opConfig, update);
-            updatedOpConfig.start = updatedOpConfig.start.format(dateFormat);
+            updatedOpConfig.start = moment(updatedOpConfig.start).format(dateFormat);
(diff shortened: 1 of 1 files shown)


godber:
I haven't managed to fully grok the code, but using moment on `start` but not `end` suggests to me that the actual error is elsewhere in the code.  Why are start and end different at that point.  Reading the code a bit leads me to `getIndexDate` which calls `_parseDate` ... which in turn uses `moment`.  Perhaps there's a branch in `getIndexDate` that should be calling moment.

godber:
I guess the fundamental design issue is that `updatedOpConfig.start` and `updatedOpConfig.end` are of two different types when they should probably be one.  Perhaps the way everything is implemented expects these two different types.  So changing it elsewhere might result in bad news elsewhere in the code.  Maybe your change is appropriate.

kstaken:
This may be related to #578

twkiel:
I tested the 578 PR and it doesn't resolve this particular issue, but it does look similar.

kstaken:
Closing this PR, replaced by #597. @twkiel please try #597.

PR #597, opened 2017-10-30
Es reader update fix

jsnoble:


kstaken:
Integration tests pass.

kstaken:
I tested this manually and it appears to work. This needs tests as this breaking should never have happened.

PR #601, opened 2017-11-01
added cluster name specificity to template names resolves #595

### LINKED ISSUE #595: Template conflict with multiple instances of teraslice on one ES cluster
Teraslice uses the `analytics_template` and `state_template` templates to enforce mappings onto it's backend support indices.  If you have multiple versions of teraslice that use the same backend elasticsearch and these require different templates then there will be a conflict.

The template names and their corresponding matching strings should probably include the teraslice instance name to disambiguate.

jsnoble:


lib/cluster/storage/backends/mappings/analytics.json
```diff
@@ -1,5 +1,5 @@
 {
-    "template": "*__analytics*",
+    "template": "__analytics*",
```
godber: These are the strings used to match the indices that the mappings get applied to right?  Don't these have to only apply to the indices specific to the cluster?  That is shouldn't these have the form like `${clusterName}__analytics*`?  Same goes for `logs` and `state`.

Or is this somehow being handled elsewhere?

jsnoble: yes in the elasticsearch_store  since all the stores use it but one

godber: Oh, OK, I see it now.

godber:
Integration tests pass

screenshot_2017-11-16_07-02-50

godber:
The templates have the cluster name in them now:

```
curl -Ss localhost:49200/_template | jq . | grep template
  "teracluster_state_template": {
    "template": "teracluster__state*",
  "teracluster_analytics_template": {
    "template": "teracluster__analytics*",
```

PR #608, opened 2017-11-07
Teraslice testing

jsnoble:
- fixed subslice by keys functionality
- fixed an internal compare Interval function to return correct interval
- updated job after interval was made so that it shows up in the ex
- tested recursive slicing down to smallest number
- tested expansion of slices
- tested proper slice creation
- tested slicer does not break on query with no results
- tested proper updates with when opConfig has start/end and start & end as well as auto

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,7 +5,8 @@
         "sourceType": "script"
     },
     "env": {
-        "node": true
+        "node": true,
+        "jasmine": true
     },
     "rules": {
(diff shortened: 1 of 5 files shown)


godber:
I re-ran travis today to make sure it ran on the latest version of the test harness, which was merged after this was create yesterday.

godber:
Integration tests pass

screen shot 2017-11-15 at 2 50 46 pm

PR #613, opened 2017-11-14
node master will now use SIGKILL on worker after timeout resolves #605

### LINKED ISSUE #605: Workers don't stop in all cases
We've been seeing various cases where workers are getting orphaned on the cluster. They still show up in /txt/workers but the slicer for the job will be gone and the workers are left doing nothing. Sometimes this corresponds with a timeout on the stop request but I don't think that always happens.

The most recent case when this occurred the workers were in a crash cycle due to a bug in the HDFS library. First call to _stop got stuck and eventually returned an empty response. Second call to _stop returned with success almost immediately but not all workers were stopped.

This is not the only scenario where it occurs though. The node master is going to need to be more aggressive about watching the workers and making sure they get stopped when required. Including using `kill -9` if necessary.

jsnoble:
NOTE CONFIG SETTING CHANGES:

- changed `network_timeout` to `action_timeout`
- added `network_latency_buffer` (defaults to 15s) which combines with action_timeout to make the final network_timeout

- can now add a a timeout parameter to stop job which will take the place of action_timeout, note that `network_latency_buffer` will still be added to it for final timeout

```
curl -XPOST localhost:5678/jobs/8b9596db-db1b-4db4-8957-b2bdc4625335/_stop?timeout=120000
```

### DIFF
--- a/lib/cluster/node_master.js
+++ b/lib/cluster/node_master.js
@@ -188,7 +188,7 @@ module.exports = function module(context) {
         callback: (data) => {
             const intervalTime = 200;
             const filterFn = worker => worker.ex_id === data.ex_id;
-            let stopTime = data.timeout || config.network_timeout;
+            let stopTime = data.timeout || config.action_timeout;
 
             messageWorkers(data, { message: 'worker:shutdown' }, filterFn);
             // If this is coming from the api, we need to listen for it, else it just a blanket stop
--- a/lib/cluster/node_master.js
+++ b/lib/cluster/node_master.js
@@ -190,7 +190,7 @@ module.exports = function module(context) {
             const filterFn = worker => worker.ex_id === data.ex_id;
             let stopTime = data.timeout || config.network_timeout;
 
-            messageWorkers(data, { message: 'worker:shutdownzzz' }, filterFn);
+            messageWorkers(data, { message: 'worker:shutdown' }, filterFn);
             // If this is coming from the api, we need to listen for it, else it just a blanket stop
             // and no reply is necessary
             if (data.response && data.to) {
--- a/lib/cluster/node_master.js
+++ b/lib/cluster/node_master.js
@@ -21,6 +21,8 @@ module.exports = function module(context) {
     const systemPorts = portAllocator();
     const config = context.sysconfig.teraslice;
     const assetsPath = config.assets_directory;
+    const events = context.apis.foundation.getSystemEvents();
+
     // Need sendNodeState before we instantiate messaging
     const sendNodeState = _.debounce(() => {
         const state = getNodeState();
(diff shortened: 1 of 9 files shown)


lib/cluster/node_master.js
```diff
@@ -187,8 +187,10 @@ module.exports = function module(context) {
         event: 'cluster:execution:stop',
         callback: (data) => {
             const intervalTime = 200;
-            let stopTime = config.network_timeout;
-            messageWorkers(data, { message: 'worker:shutdown' }, worker => worker.ex_id === data.ex_id);
+            const filterFn = worker => worker.ex_id === data.ex_id;
+            let stopTime = data.timeout || config.network_timeout;
+
+            messageWorkers(data, { message: 'worker:shutdownzzz' }, filterFn);
```
kstaken: worker:shutdownzzz?

kstaken:
Had a scenario today where workers were refusing to exit and decided to try this branch. This did not work to kill the workers.

jsnoble:
Sorry about that, just changed it back. I changed the message name so that it would not be recoginzed so that I could verify that the process would be killed after the alloated time

kstaken:
This doesn't have any effect unless a stop job is called through the API. And even then it doesn't work. I tried logging stopTime on line 215 of node_master and it is NaN for me.

This needs to work in all scenarios where workers should be stopped.

kstaken:
```let stopTime = data.timeout || config.network_timeout;```

Both of those options show up as undefined for me.

**Comment by jsnoble** [comment]
Ok, fixed the wrong reference, it should say config.action_timeout

**Comment by kstaken** [comment]
This PR is not ready. 
* Worker shutdown behavior should be handled in a consistent manner for all cases not just coming through the API. Hitting ctrl-c, normal termination of the node_master, remove worker and stopping the job all need to guarantee workers exit.
* The API still times out and provides no result before workers exit with default timeouts.
* network_timeout has been renamed but appears to still be referenced in: ./cluster/services/cluster/backends/native.js:            }, context.sysconfig.teraslice.network_timeout);

**Comment by kstaken** [comment]
Even when the timeout triggers and kill is called it still doesn't really kill the workers. They are removed from cluster state so don't show up under /txt/workers but the processes are still there at the OS level.

**Comment by kstaken** [comment]
I was able to get it to actually kill the processes with this.

```
    function killWorkers(filterFn) {
        const spawn = require( 'child_process' ).spawnSync; <-- obviously wouldn't belong here.
        const workers = context.cluster.workers;
        const allWorkersForJob = _.filter(workers, filterFn);

        allWorkersForJob.forEach(worker => {
            spawn('kill', ['-9', worker.process.pid]);
        });
    }
```

godber:
This one is still a WIP right?

kstaken:
Manually confirmed ctrl-c, API _stop and API _remove workers will eventually kill stuck workers. 

The default timeout for _stop is longer than the default timeout for curl so you'll get a curl timeout before the workers are actually killed. Also it appeared the timeout on _remove and ctrl-c was shorter than for _stop.

kstaken:
Integration tests pass.

PR #618, opened 2017-11-16
added validation on start job as well resolves #616

### LINKED ISSUE #616: Teraslice should validate jobs on start as well
When upgrading an existing teraslice cluster it is possible to hit a scenario where code changes to operations add new parameters.  Starting stored jobs will fail because validation doesn't happen on start.  Validation should happen on start and the defaults and other validation steps will be handled.

jsnoble:


godber:
I am having a hard time verifying that this change does what we want.

The process I am following is that I ran the integration test framework and left the images running just to get things set up.

Then I picked a job to "tweak" , in this case it was the `basic reindex` job, I then dumped that job to a file

```
curl -Ss localhost:49200/teracluster__jobs/job/e014bfe5-82da-43a8-8d70-920db35ec9b8 | jq . > before.json
```

Then I copied that file and removed all ES metadata so that the new job was just the `_source` property from the ES document.  I deleted the `delay` property from the `elasticsearch_reader` so the job looked like this:

```json
{
    "name": "basic reindex",
    "lifecycle": "once",
    "analytics": true,
    "max_retries": 3,
    "slicers": 1,
    "workers": 1,
    "operations": [
      {
        "index": "example-logs-10",
        "type": "change",
        "size": 1000,
        "start": null,
        "end": null,
        "interval": "auto",
        "full_response": true,
        "date_field_name": "created",
        "query": "bytes:>=99999999",
        "fields": null,
        "subslice_by_key": false,
        "subslice_key_threshold": 50000,
        "time_resolution": "s",
        "key_type": "base64url"
      },
      {
        "_op": "elasticsearch_index_selector",
        "index": "test-reindex-bad-query",
        "type": "change",
        "preserve_id": false,
        "id_field": "_key",
        "timeseries": "",
        "index_prefix": "",
        "date_field": "",
        "delete": false,
        "update": false,
        "update_retry_on_conflict": 0,
        "update_fields": [],
        "upsert": false,
        "create": false,
        "script_file": "",
        "script": "",
        "script_params": {}
      },
      {
        "_op": "elasticsearch_bulk",
        "size": 100,
        "connection_map": {
          "*": "default"
        },
        "multisend": false,
        "multisend_index_append": false,
        "connection": "default"
      }
    ],
    "moderator": null,
    "recycle_worker": null,
    "job_id": "e014bfe5-82da-43a8-8d70-920db35ec9b8",
    "_context": "job",
    "_created": "2017-11-16T21:36:46.366Z",
    "_updated": "2017-11-16T21:36:46.366Z"
  }
```

 I then posted that back to the same JOB ID:

```
curl -XPOST -Ss localhost:49200/teracluster__jobs/job/e014bfe5-82da-43a8-8d70-920db35ec9b8 -d @new.json | jq .
```

Then I started the job:

```
curl -XPOST -Ss localhost:49200/teracluster__jobs/job/e014bfe5-82da-43a8-8d70-920db35ec9b8/_start
```

and inspected the job again:

```
curl -Ss localhost:45678/jobs/e014bfe5-82da-43a8-8d70-920db35ec9b8/
```

At this point I had hoped that the job would have gone through validation and the `"delay": "30s",` property would be present in the job again (provided because the job had been validated after calling `_start`).

I can't tell if my process is no good or if the branch just isn't accomplishing what I had expected.  Am I misunderstanding something.  What did you do to manually verify this?

jsnoble:
I tested this by first submitting a simple regular job then shutdown teraslice. I added another field in the schema for one of ops of that job then started terslice back up. I then proceeded to call _start on  the job_id that I previously submitted and inspected the ex in elasticsearch and saw that the new field was there. To test that it would fail a validation appropriately, I did the same thing but made sure the new schema would throw an error if it was left empty and verified it

kstaken:
@godber this should be reflected in the `ex` record after start rather than the `job` record.

godber:
Ok, I see the `delay` value is restored in the ex.  So I guess this works as expected.  I will file my other concerns as a separate ticket.

PR #628, opened 2017-12-04
slicer can not direct slices to a particular worker resolves #624

### LINKED ISSUE #624: allow slices to be sent to specific workers
We need the capabilities to allow the slicer to determine where a particular slice goes to. We need to be able to enforce that a slice of a particular type only head to a given worker. After the worker completes a slice it would need to respond back with enough meta data to empower this.

jsnoble:
this needs to be paired with the pr in the queue package.

In order to send a slice to a specific worker you will need to add a key to the slice from the reader-slicer

example slice from elasticsearch date slicer
```   
{
    start: someStartDate,
    end: someEndDate,
    count: data.count,
    request_worker: '172.29.0.170__4'  // the actual workerId
}
```

events to listen to:

'slice:invalid' => returns the slice, itmeans the slice was not able to be allocated in the worker queue, the state slice will be marked as invalid and you will need to change request_worker to another worker

'worker:ready' => returns workerId, use this to determine what workers have connected and their actual id

'network:disconnect' => returns a workerId, tells you what worker has dropped a connection to the slicer

'slice:success' => slice, the slice has been reported back by the worker and was sucessfull

slice:failure' => slice, the slice failed

### DIFF
--- a/lib/cluster/slicer.js
+++ b/lib/cluster/slicer.js
@@ -104,7 +104,7 @@ module.exports = function module(contextConfig) {
             logger.info(`slicer for job: ${exId} has received a pause notice`);
             engineCanRun = false;
             clearInterval(engine);
-            events.emit('job:pause');
+            events.emit('execution:pause');
             messaging.respond(msg);
         }
     });
@@ -115,7 +115,7 @@ module.exports = function module(contextConfig) {
             logger.info(`slicer for job: ${exId} has received a resume notice`);
             engine = setInterval(engineFn, 1);
             engineCanRun = true;
-            events.emit('job:resume');
+            events.emit('execution:resume');
             messaging.respond(msg);
         }
     });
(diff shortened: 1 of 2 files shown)


godber:
You mean this PR: https://github.com/terascope/queue/pull/5 right?

erik-stephens:
When a new slicer is created, how should it enumerate the current workers?

erik-stephens:
After chatting with @jsnoble, listening for `worker:ready` is the way to go, even on new slicer.  I understand the process to be:

1. job submitted
1. new slicer created
1. cluster master allocates workers to the job, emitting said event
1. new slicer waits until requested allocation filled or times out

Open questions:

1. How long should a slicer should wait?
1. If slicer times out waiting for full allocation, what is expected behavior?  Fail hard?  Move on as best it can?  Support either route via job configuration?

kstaken:
Should we just close this PR?

jsnoble:
another pr will have this feature

PR #629, opened 2017-12-05
fixed asset deployment across cluster resolves #627

### LINKED ISSUE #627: Asset deployment issue on cluster
Just ran a job that requested 30 workers but only 10 showed up under /txt/slicers. All 30 workers were allocated and visible under /txt/workers. This was a 3 node cluster. This was the first run of a job using a newly deployed asset and I believe that the workers that succeeded in running were on the master while the 20 allocated to other nodes did not start. I suspect this is a new iteration of the asset loading problem. I let the job run for several minutes but the workers never became available. Stopping and starting the job allowed all 30 workers to function.

jsnoble:


kstaken:
Integration tests pass

PR #635, opened 2017-12-13
Fixing some json / formatting

mkhpalm:


godber:
Thanks for cleaning those up!

PR #651, opened 2018-01-12
fixed update to save unparse job config resolves #634

### LINKED ISSUE #634: Roles of Job vs Execution Context

jsnoble:
integration tests pass


jsnoble:
This requires changes made in teraslice-client module

integration-tests/spec/cases/assets/simple.js
```diff
@@ -44,41 +42,38 @@ module.exports = function simpleAssetTest() {
     }
 
     describe('Asset Tests', () => {
-        describe('After uploading an asset', () => {
```
godber: You didn't like my grouping of tests with additional nested subscribes?

jsnoble: it was against how we formatted all the rest of the tests, describe was intended to determine the 'class' of test, and 'it' was to test all things related to that class

godber: Ok, but sometime you should take another look at the Jasmine docs and read about BDD to see what I think it's trying to achieve.  I don't think it matters much one way or another though.

integration-tests/spec/fixtures/jobs/generator-asset.json
```diff
@@ -3,6 +3,7 @@
     "slicers": 1,
     "lifecycle": "persistent",
     "workers": 3,
+    "assets": ["ex1"],
```
godber: Is there a specific reason you add this here?  In all of your versions of the test and my versions of the test the `.assets` property gets overwritten by `jobSpec.assets = [JSON.parse(result)._id];`.  I had deliberately omitted it from here to keep the fixture generic.  Though I guess it's confusing that the job requires an asset but omits it.

kstaken:
@godber are your questions blockers on merging this?

godber:
Nah, go ahead.

kstaken:
Integration tests pass and job update behavior manually verified.

PR #655, opened 2018-01-30
persitent jobs can return to running resolves #654

### LINKED ISSUE #654: Change meaning of `failing` status for persistent jobs
On persistent jobs `failing` has proven to be a useless status. Once a single slice fails the job is forever marked as `failing` which from an operational perspective is not helpful. It would be more useful if `failing` was a transient state (persistent jobs only) that the job enters when there is some kind of recoverable disruption. If the job returns to processing slices it should return to `running` status.

We would need to set some parameters around the transition from `failing` back to `running` either based on number of successful slices or a period of time without failures and this setting should be controllable on a per job basis. 

In some ways this is a stopgap in the process of fully reworking state transition tracking and how we deal with failures but this should hopefully be a minor change that will have solid operational benefits.

Ultimately state on the job will need to be a timeline rather than a single indicator and we'll need to have an audit log of state changes that gets tracked as the job runs.

jsnoble:


kstaken:
I can't get this to work. My test involves a data generator -> kafka job where I then stop the kafka cluster and bring it back up. The job goes into `failing` and stays there. I've tried it both with the default timeout and also setting `"probation_window": "10000",` with the same results.

Also the setting `probation_window` is not ideal. Naming this is a tough one but maybe something like `failing_recovery_window`?

kstaken:
This needs a rebase.

kstaken:
Previously failing manual tests now pass.

kstaken:
Integration and unit tests pass.

PR #666, opened 2018-02-16
WIP: Refactoring asset loading to be sequenced more explicitly

kstaken:
Current status
 * Appears to work but needs more testing to verify the benefit
 * Integration tests will pass but have been really flaky.
 * Unit tests do NOT pass yet
 * Still more cleanup of debug messages and commented code required.

The goal of this PR is to explicitly sequence the loading of assets with startup of the slicer and workers. This removes any timers or intervals from the startup sequence for the job. The slicer and worker will not start running until the preliminary setup is complete. 

The availability of the slicer and worker are now tied to the firing of the `asset:loaded` event and that event will fire whether there are assets on the job or not. We should probably consider renaming that event since it's more like `execution:prepared` or something now.

This change was largely prompted by the asset related logic in the execution runner which just didn't make any sense to me. Since we're going to be adding a new runner for streaming seemed a good time to clean this up.

@jsnoble please look this over and let me know if it looks like the approach will break something.

### DIFF
--- a/lib/cluster/assets_loader.js
+++ b/lib/cluster/assets_loader.js
@@ -17,7 +17,7 @@ module.exports = function module(context) {
         // first step we normalize all identifiers to their proper id
         return assetStore.parseAssetsArray(assetsArray)
             .then(idArray => Promise.map(idArray, (assetIdentifier) => {
-                if (!alreadyDownloaded(assetIdentifier)) {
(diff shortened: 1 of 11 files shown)


kstaken:
This is aborted and will be handled with more extensive refactoring based around the teraslice-worker package.

Issue #589, opened 2017-10-12
txt/slicer endpoint


jsnoble:
There is a potential race condition of using txt/slicers while a job is being allocated. The slicer might technically show up in the cluster state but it might not fully be instantiated thus it could not return its details causing the whole call to hang.

jsnoble:
Cannot reproduce this problem with teraslice v 0.28. We have since made major refactors to the cluster services and changes to the slicer. The slicer will be undergoing major refactoring once again and will be checked again after those changes are brought in to see if this is still an issue

PR #667, opened 2018-02-16
slicer endpoint should not hang on new slicer creation resolves #659 …

### LINKED ISSUE #659: /txt/slicers hangs while a job is scheduling
Hitting /txt/slicers while a job is in the process of scheduling causing the endpoint to hang and never return.

jsnoble:
…#589

### DIFF
--- a/lib/cluster/services/api.js
+++ b/lib/cluster/services/api.js
@@ -55,7 +55,7 @@ module.exports = function module(context, app) {
 
 
     app.post('/jobs', (req, res) => {
-        // if no job was posted an empty object is returned, so we check if it has values 
+        // if no job was posted an empty object is returned, so we check if it has values
         if (!req.body.operations) {
             sendError(res, 400, 'No job was posted');
         } else {
@@ -206,7 +206,7 @@ module.exports = function module(context, app) {
             .then(responseObj => res.status(200).send(`${responseObj.workerNum} workers have been ${responseObj.action} for execution: ${responseObj.ex_id}`))
             .catch((errObj) => {
                 if (errObj.code) {
-                    logger.error(errObj.error);
+                    logger.error(errObj.message);
                     sendError(res, errObj.code, errObj.message);
                     return;
                 }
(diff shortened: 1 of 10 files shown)


lib/cluster/services/api.js
```diff
@@ -465,8 +488,15 @@ module.exports = function module(context, app) {
...
-                sendError(res, err.code, err.message);
+            .catch((errObj) => {
+                if (errObj.code) {
+                    logger.error(errObj.message);
+                    sendError(res, errObj.code, errObj.message);
+                    return;
+                }
+                const errMsg = `could not get slicer statistics, error: ${parseError(errObj)}`;
+                logger.error(errMsg);
+                sendError(res, 500, errMsg);
```
kstaken: Come on ... this code sequence is repeated 4 times... I can't explain how frustrating that is after all the time we spent trying to clean this module up. 

Does this really require special casing like this? Where does that `.code` attribute come from? Why can't it simply be handled in the general error handler?

lib/cluster/services/cluster/backends/native.js
```diff
@@ -601,22 +601,20 @@ module.exports = function module(context, messaging, executionService) {
         }));
     }
 
-    function getSlicerStats(exId) {
+    function getSlicerStats(exIds, specificId) {
```
kstaken: This looks questionable to me. What required this API to change in this way?

godber:
I am not really seeing #659 as being resolved.

When I run the integration tests, I have typically been able to run `tstop -p 45678` and watch the jobs progress ... now it appears to hang.  Checking with curl it appears that `/txt/slicers` blocks while a job is running and `/txt/workers` and `/txt/nodes`.

Pretty easy to reproduce ... just start up the integration tests and walk through the endpoints ... 

```
curl -Ss localhost:45678/txt/workers
curl -Ss localhost:45678/txt/jobs
curl -Ss localhost:45678/txt/ex
curl -Ss localhost:45678/txt/slicers
```

They will all return except slicers.  You might catch the tests between jobs and `/txt/slicers` returns then ... but its not hard to get it to fail.

kstaken:
I talked to @jsnoble about this and he thinks the issue with hitting it while the integration tests are running is down to the rapid fire nature of how jobs are scheduled in the tests and tstop catching them after a job has ended but before the slicer has fully shutdown. This is a different issue and he will open a separate ticket and put a fix in separate PR. 

As for #659 this PR does appear to resolve that issue. The integration tests pass otherwise so going ahead with a merge.

PR #671, opened 2018-03-05
fixed message passing to asset loader resolves #670

### LINKED ISSUE #670: asset loader crashing
It was determined that in certain cases it was possible that the asset_loader would be sent messages for the execution, since it did not have the right listeners it would end up crashing the loader making the assets downloaded either corrupted or unzipped as well as stalling the workers/execution_controller on that particular node.

jsnoble:


kstaken:
Anxiously awaiting messaging tests. Integration tests pass.

Issue #675, opened 2018-03-12
execution_controller crash


kstaken:
We're testing 0.29 with a kafka job. The job ran for a few minutes, processed a few hundred slices and then crashed with this error:

```
    "_failureReason": "TypeError: Cannot read property 'initialize' of undefined\n    at /app/source/node_modules/teraslice/lib/cluster/execution_controller/index.js:38:23\n    at bound (domain.js:287:14)\n    at runBound (domain.js:300:12)\n    at tryCatcher (/app/source/node_modules/teraslice/node_modules/bluebird/js/main/util.js:26:23)\n    at Promise._settlePromiseFromHandler (/app/source/node_modules/teraslice/node_modules/bluebird/js/main/promise.js:510:31)\n    at Promise._settlePromiseAt (/app/source/node_modules/teraslice/node_modules/bluebird/js/main/promise.js:584:18)\n    at Promise._settlePromises (/app/source/node_modules/teraslice/node_modules/bluebird/js/main/promise.js:700:14)\n    at Async._drainQueue (/app/source/node_modules/teraslice/node_modules/bluebird/js/main/async.js:123:16)\n    at Async._drainQueues (/app/source/node_modules/teraslice/node_modules/bluebird/js/main/async.js:133:10)\n    at Immediate.Async.drainQueues [as _onImmediate] (/app/source/node_modules/teraslice/node_modules/bluebird/js/main/async.js:15:14)\n    at processImmediate [as _immediateCallback] (timers.js:383:17)"
```

kstaken:
@ciorg also just encountered this same issue on his laptop. Here's a little more context. 

Something else that's weird about this is that the analytics is showing 8 steps in the job but the job really only has 4. It's likely this exception occurred after a failed slice.

```
[2018-03-12T18:16:05.163Z]  INFO: teraslice/30355 on charlies-MacBook-Pro.local: analytics for slice time : 5216,7,0,5224,0,0,0 size : 10,10,20,0,0,0,0 memory : 107104,829824,21712,77552,2040,3848,1848  (ex_id=756c4c7a-4273-4ec3-9ea7-fad6f05f02eb, job_id=b772b4aa-c662-41a5-a701-d182ed02a75a, module=slice, worker_id=172.29.1.118__37, slice_id=823e0961-2514-4902-8060-740178e8ec1a)
[...]

PR #676, opened 2018-03-12
Execution fix

jsnoble:
This addresses the initialize bug as well as the analytics bug on retries adding to much data than the steps. There are tests for the analytics bug


kstaken:
fixes #675 

Not seeing failure scenario on this branch in manual testing. Unit and integration tests pass.

PR #679, opened 2018-03-13
 tests and api lifecycle changes

jsnoble:


jsnoble:
this needs to be used in conjunction with the pr in teraslice-client-js  https://github.com/terascope/teraslice-client-js/pull/11

godber:
This passes when using the specified branch

screen shot 2018-03-29 at 2 39 04 pm

Jared, many thanks for directly linking to the required branch!

PR #683, opened 2018-03-19
fixed formatting and slicer expansion bug resolves #680

### LINKED ISSUE #680: worker crashes when Teraslice tries to slice records with a large time gap (10 days)
While moving records (auto generated data) from one ES index to another Teraslice failed to find a usable slice interval and crashed.  There was a time gap of about 10 days in the index's records that were being moved over.  The slice trying to span the time gap ended up grabbing the remaining records (6.25M) which caused the error.

ERROR: teraslice/16628 on: invoking elasticsearch_api _performSearch resulted in a runtime error: error: query_phase_execution_exception Result window is too large, from + size must be less than or equal to: [2000000] but was [6275000].
ERROR: teraslice/16628 on: failed to process false, slice state has been marked as error

Searched for excessive record density around the slice's beginning and ending times, but found nothing to suggest that this was the cause of the crash.

jsnoble:


godber:
The unit tests pass

screen shot 2018-03-29 at 5 41 35 pm

the integration tests pass

screen shot 2018-03-29 at 5 51 32 pm

godber:
I lack an independent way to validate this fix so I'm accepting the included unit test as sufficient.

It would have been preferable in this case to have two commits, a formatting commit then a fix commit.

PR #684, opened 2018-03-19
fixed assets bug

jsnoble:


godber:
Integration and unit tests pass:

screen shot 2018-04-02 at 3 31 49 pm

godber:
I finally managed to reproduce the error on an internal test cluster using Zach's Monitoring job.  I verified that the same job worked on the test cluster using this rest branch.

I was never able to reproduce the race condition with my simpler tests.  The Monitoring job has a fairly large asset in comparison.  So perhaps that was the difference.

Issue #574, opened 2017-09-29
Add date_field validation to elasticsearch_index_selector


godber:
If you create a job that uses timeseries indices like this:

```
{
  "name": "Data Generator",
  "lifecycle": "persistent",
  "workers": 1,
  "operations": [
    {
      "_op": "elasticsearch_data_generator",
      "size": 5000
    },
    {
      "_op": "elasticsearch_index_selector",
      "type": "events",
      "index_prefix": "test-",
      "timeseries": "daily",
      "date_field": "FOO"
    },
    {
      "_op": "elasticsearch_bulk",
      "size": 5000
    }
  ]
}
```

you will get an error like the one shown below because the field `FOO` doesn't exist

```
[2017-09-28T23:37:51.926Z] ERROR: teraslice/5564 on stretch.local: (ex_id=e9e0272c-8c35-4e5e-adc9-bf081e2edd3d, module=slice, worker_id=172.29.0.181__4, slice_id=3360576f-60c2-4aa6-9153-135dc26f661b)
    An error has occurred: RangeError: Invalid time value
        at Date.toISOString (native)
        at formattedDate (/Users/godber/Workspace/terascope/opensource/teraslice/lib/processors/elasticsearch_index_selector.js:17:58)
        at indexName (/Users/godber/Workspace/terascope/opensource/teraslice/lib/processors/elasticsearch_index_selector.js:25:25)
        at generateRequest (/Users/godber/Workspace/terascope/opensource/teraslice/lib/processors/elasticsearch_index_selector.js:80:27)
        at /Users/godber/Workspace/terascope/opensource/teraslice/lib/processors/elasticsearch_index_selector.js:164:13
        at /Users/godber/Workspace/terascope/opensource/teraslice/lib/utils/analytics.js:24:32
        at /Users/godber/Workspace/terascope/opensource/teraslice/lib/cluster/worker.js:203:24
        at bound (domain.js:287:14)
        at runBound (domain.js:300:12)
        at tryCatcher (/Users/godber/Workspace/terascope/opensource/teraslice/node_modules/bluebird/js/main/util.js:26:23)
        at Promise._settlePromiseFromHandler (/Users/godber/Workspace/terascope/opensource/teraslice/node_modules/bluebird/js/main/promise.js:510:31)
        at Promise._settlePromiseAt (/Users/godber/Workspace/terascope/opensource/teraslice/node_modules/bluebird/js/main/promise.js:584:18)
[...]

PR #685, opened 2018-03-26
added better error messages to incorrect date_field settings resolves…

jsnoble:
… #574


godber:
Integration tests pass:

screen shot 2018-04-02 at 4 36 40 pm


The error message is improved and more explicit:

```
An error has occurred: Error: opConfig date field: FOO either does not exists or is not a valid date on the records processed
```

PR #695, opened 2018-04-19
improved error handling on asset store to not crash cluster master, i…

### LINKED ISSUE #694: Asset uploading fails and crashes the API
I just tried to update an asset bundle on a cluster and the upload appeared to fail. 

```
curl -Ss -XPOST -H "Content-Type: application/octet-stream" ts_gen4.tera1.terascope.io:5678/assets --data-binary @out/asset-${CI_BUILD_REF:0:6}.zip > result.txt
curl: (56) Recv failure: Connection reset by peer
```

On trying it again it appeared to not error but the asset wasn't actually uploaded and all APIs started throwing errors. The cluster needed to be restarted to recover. After that uploading the asset again worked fine.

```
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Cannot GET /txt/slicers</pre>
</body>
</html>
```

jsnoble:
…mprove http error responses resolves #694

### DIFF
--- a/lib/cluster/cluster_master.js
+++ b/lib/cluster/cluster_master.js
@@ -4,6 +4,8 @@ const Promise = require('bluebird');
 const makeLogs = require('./storage/logs');
 const messageModule = require('./services/messaging');
 const _ = require('lodash');
+const http = require('http');
+const express = require('express');
 
 module.exports = function (context) {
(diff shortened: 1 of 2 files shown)


kstaken:
Integration and manual tests pass.

PR #699, opened 2018-05-01
allowed stop to bypass saftey check resolves #697

### LINKED ISSUE #697: Can't restart a job that wasn't stopped
If you end up with an ex in `running` status that is not really `running` then there is no way to restart the job without either manually mutating the ex record or simply deleting the ex record. 

This circumstance  can easily happen if a cluster is restarted or the cluster master simply crashes. 

If you call stop (or any operation) on an ex and the ex is not running then it should be smart enough to put it into a state that is recoverable. Perhaps it should move it to `terminated` status.

```
curl -sS -XPOST localhost:5678/jobs/303f7f11-8f76-40fa-8ead-bd0fef3bcaec/_stop
{
    "error": 404,
    "message": "Could not find active execution processes for ex_id: 8d67c0ab-5bf4-4027-9519-536bfa375538"
}
```

jsnoble:


godber:
Works for me.

godber:
Uh, not to imply that I've tested this, just that this solution looks good enough.

PR #703, opened 2018-05-09
Minimal way of versioning API endpoints. Resolves #702

### LINKED ISSUE #702: Support versioned HTTP API requests
To support changes to the HTTP API without breaking existing tools, and job, Teraslice needs an API versioning convention. @kstaken and I discussed using a route prefix for all existing machine-readable API endpoints. For example,  `/cluster/state` should be moved to `/v1/cluster/state` but `/txt/workers` shouldn't be changed.

Ideally, this can be done without making breaking changes by supporting both the `/v1/` prefix and the current non-prefix endpoints.

peterdemartini:
I intentionally made this with very few changes, but I have more ideas for making it more flexible and maintainable.


kstaken:
needs a rebase

jsnoble:
unit and integration tests pass

PR #707, opened 2018-05-15
Fix error handing of a failing processor, resolves #706

### LINKED ISSUE #706: Handle errors properly when assets are loaded but not throw an error when instantiating
When instantiating a custom processor in a loaded asset bundle, and the processor throws an error, the Teraslice executor will consider the asset as "not loaded" and ignore the error and wait for the assets to load.

peterdemartini:
See #706


kstaken:
Needs a rebase

jsnoble:
Unit and integration tests pass. I submitted an asset bundle with faulty code. On master branch it just sat there waiting. This branch I was able to verify that the error was propagated correctly

PR #708, opened 2018-05-16
Api endpoint json assets metadata

ciorg:
api endpoint /assets or /assets/assetName returns asset metadata in array of json objects.
Created integration tests for new api endpoints and txt/assets, txt/assets/assetName since their underlying code was changed.


lib/cluster/services/api.js
```diff
@@ -53,6 +53,8 @@ module.exports = function module(context, app) {
     app.route('/txt/assets*')
         .get(_redirect);
 
+    app.route('/assets*')
```
jsnoble: this should be placed on the app.route('/assets*') that is a few lines above to not duplicate it. You just need to move the .get() to the top one

lib/cluster/services/assets.js
```diff
@@ -76,12 +76,23 @@ module.exports = function (context) {
         createAssetTable(query, req, res);
     });
 
+    app.get('/assets', (req, res) => {
+        const query = 'id:*';
+        assetsSearch(query, req, res)
+            .then(results => res.status(200).send(JSON.stringify(results, null, 4)));
+    });
+
+    app.get('/assets/:name', (req, res) => {
```
jsnoble: we should add /assets/:name/:version as well and add a integration test for that as well

jsnoble:
unit and intergration tests pass. Manual testing of new api endpoints pass

PR #717, opened 2018-06-01
teraslice is compliant with terafoundation changes

jsnoble:


kstaken:
Integration tests pass.

Issue #511, opened 2017-08-23
Production cluster management - Kubernetes


kstaken:
The built in clustering is great for getting up and running quickly but for large scale production deployments we want to use a more mature cluster manager to manage resources. The goal is for the clustering interface to be generic and support multiple clustering backends but to start we'll support Kubernetes for the production cluster solution along with our built in native clustering.

This is a tracking ticket for discussion and design. More detail will follow.

### Reminders
- [ ] Don't forget to make sure assets work.
- [ ] How does the master get started?

kstaken:
Clustering API has been refactored to enable creation of new clustering backends.

godber:
I don't know all of the ways `clusterState` gets used yet but with the native cluster controller (and no jobs running) it looks like this:

```json
{
  "r7c1n7.example.lan": {
    "node_id": "r7c1n7.example.lan",
    "hostname": "10.210.0.104",
    "pid": 1,
    "node_version": "v4.6.1",
    "teraslice_version": "0.28.0",
    "total": 50,
    "state": "connected",
    "available": 48,
    "active": [
      {
        "worker_id": 1,
        "assignment": "cluster_master",
        "pid": 14
      },
      {
        "worker_id": 2,
        "assignment": "assets_service",
        "pid": 20
      }
    ]
  },
  "r7c1n8.example.lan": {
    "node_id": "r7c1n8.example.lan",
    "hostname": "10.210.0.105",
    "pid": 1,
    "node_version": "v4.6.1",
    "teraslice_version": "0.28.0",
    "total": 50,
    "state": "connected",
    "available": 50,
    "active": []
  },
  "r6c1n8.example.lan": {
    "node_id": "r6c1n8.example.lan",
    "hostname": "10.210.0.57",
    "pid": 1,
    "node_version": "v4.6.1",
    "teraslice_version": "0.28.0",
    "total": 50,
    "state": "connected",
    "available": 50,
    "active": []
  }
}
```

It probably makes sense for this to include all k8s nodes, but things like `node_version` don't quite make sense without a persistent node master running.

jsnoble:
you may be right that we would need to reconsider what meta data we publish but as for node_version, wont these k8 nodes be running on several different servers that could potentially have different versions of node? would others? Or is this meta data useful at all as it stands?

godber:
[...]

PR #738, opened 2018-06-15
Issue 511 add k8s clustering node8

godber:
Here is a rough cut at adding the k8s implementation for teraslice clustering.  There are a few things I'd like to do before we merge this.  I need to go through and review my FIXME comments, some of them are intended to ask for feedback from you guys.  I need to go through and think about all of the error situations.

It should probably be mostly harmless to merge this (when it merges, I now see I am in conflict again somehow) since all of the functionality requires users to switch the clustering implementation from default.

https://github.com/terascope/teraslice/issues/511


lib/cluster/services/cluster/backends/kubernetes/index.js
```diff
@@ -0,0 +1,383 @@
...
+        allocateSlicer, // k8s
+        // shutdown,  FIXME: ??
+        stopExecution,
+        pauseExecution,
+        resumeExecution,
+        removeWorkers, // k8s
+        addWorkers, // k8s
+        setWorkers, // k8s
+        readyForAllocation,
+        // clusterAvailable FIXME: ??
```
kstaken: I think you'd want this to return false if for some reason the k8s services were unavailable and it's not possible to manage a job.

lib/cluster/services/cluster/backends/kubernetes/deployments/slicer.json
```diff
@@ -0,0 +1,69 @@
+{
+    "apiVersion": "apps/v1",
+    "kind": "Deployment",
+    "metadata": {
+        "labels": {
+            "app": "teraslice",
+            "nodeType": "execution_controller",
+            "exId": "<EXID>"
```
godber: Every occurrence of `<EXID>` ends up being overwritten by the code (`backends/kubernetes/index.js`).  I wasn't sure if this was a good idea or not.  I didn't want to leave the keys out of this file which is effectively a template so I just used this convention (`<VAR>`) to identify the things that require replacing.  If no one has any objections I'll just leave it as is.

lib/cluster/services/cluster/backends/kubernetes/deployments/slicer.json
```diff
@@ -0,0 +1,69 @@
...
+            "metadata": {
+                "labels": {
+                    "app": "teraslice",
+                    "nodeType": "execution_controller",
+                    "exId": "<EXID>"
+                }
+            },
+            "spec": {
+                "containers": [{
+                    "image": "teraslice:asgdev",
```
godber: This line selects the teraslice Docker/k8s container name to run when it launches a slicer/execution_controller (`teraslice:asgdev`).  I wasn't sure what to use at the time.  I guess I am still not sure.  I think `teraslice:latest` is an obvious choice, but operationally speaking we might not want that.  Perhaps this is a value that should be set in the code and somehow read from configuration?  Or from the job?

lib/cluster/services/cluster/backends/kubernetes/deployments/slicer.json
```diff
@@ -0,0 +1,69 @@
...
+                    "name": "config",
+                    "configMap": {
+                        "name": "teraslice-worker",
+                        "items": [{
+                            "key": "teraslice-worker.yml",
+                            "path": "teraslice.yaml"
+                        }]
+                    }
+                }],
+                "imagePullSecrets": [{ "name": "docker-tera1-secret" }]
```
godber: This line assumes that the k8s cluster has a secret named `docker-tera1-secret` loaded.  This should probably come from configuration or the job.

godber: Or we can choose a more generic name and view it as a requirement imposed on the k8s cluster by teraslice.

lib/cluster/services/cluster/backends/kubernetes/deployments/slicer.json
```diff
@@ -0,0 +1,69 @@
...
+                            "fieldRef": {
+                                "fieldPath": "status.podIP"
+                            }
+                        }
+                    }]
+                }],
+                "volumes": [{
+                    "name": "config",
+                    "configMap": {
+                        "name": "teraslice-worker",
```
godber: This assumes there is a config map named `teraslice-worker` loaded into the k8s cluster.  I think this is something we can safely require as long as we document the requirement in a `docs/deploy-k8s.md`.

lib/cluster/services/cluster/backends/kubernetes/deployments/worker.json
```diff
@@ -0,0 +1,69 @@
+{
+    "apiVersion": "apps/v1",
+    "kind": "Deployment",
+    "metadata": {
+        "labels": {
+            "app": "teraslice",
+            "nodeType": "worker",
+            "exId": "<EXID>"
```
godber: Same `<VAR>` convention as above.

lib/cluster/services/cluster/backends/kubernetes/deployments/worker.json
```diff
@@ -0,0 +1,69 @@
...
+            "metadata": {
+                "labels": {
+                    "app": "teraslice",
+                    "nodeType": "worker",
+                    "exId": "<EXID>"
+                }
+            },
+            "spec": {
+                "containers": [{
+                    "image": "teraslice:asgdev",
```
godber: Same container name as mentioned above.

lib/cluster/services/cluster/backends/kubernetes/deployments/worker.json
```diff
@@ -0,0 +1,69 @@
...
+                            "fieldRef": {
+                                "fieldPath": "status.podIP"
+                            }
+                        }
+                    }]
+                }],
+                "volumes": [{
+                    "name": "config",
+                    "configMap": {
+                        "name": "teraslice-worker",
```
godber: Same configmap requirement as mentioned above.

lib/cluster/services/cluster/backends/kubernetes/deployments/worker.json
```diff
@@ -0,0 +1,69 @@
...
+                        "items": [{
+                            "key": "teraslice-worker.yml",
+                            "path": "teraslice.yaml"
+                        }]
+                    }
+                }],
+                "imagePullSecrets": [{ "name": "docker-tera1-secret" }]
```
godber: Same secret assumption as above.

lib/cluster/services/cluster/backends/kubernetes/deployments/slicer.json
kstaken: Is this something that the system should automatically setup?

lib/cluster/services/cluster/backends/kubernetes/deployments/slicer.json
kstaken: Does seem like something that will need to be configurable.

godber: Should I try to set a cluster level default that has to be specified in the config file (configmap) that can then be overridden by the job?  I would think that is possible, not too hard, and about the best option.

lib/cluster/services/cluster/backends/kubernetes/deployments/slicer.json
godber: I wouldn't think so, this is the config file for the teraslice worker:

```
terafoundation:
    environment: 'development'

    connectors:
        # ***********************
        # Elastic Search Configuration
        # ***********************
        elasticsearch:
            default:
                host:
                    - "192.168.99.1:9200"
teraslice:
    ops_directory: '/app/source/examples/ops/'
    master: false
    master_hostname: "teraslice-master"
    name: "teracluster"
    cluster_manager_type: kubernetes
```

lib/cluster/services/cluster/backends/kubernetes/index.js
```diff
@@ -0,0 +1,383 @@
...
+        return k8s.list('app=teraslice', 'pods')
+            .then((k8sPods) => { k8sState.gen(k8sPods, clusterState); });
+    }
+
+    function getClusterStats() {
+        return clusterStats;
+    }
+
+    // FIXME: copied from native
+    function _iterateState(cb) {
```
godber: This function `_iterateState` is copied from `native.js`.  There is another function I've copied in this module that makes use of it, `findSlicersByExecutionID`.  In `native.js` there are two more functions that use it.  All of these probably need to be extracted out of `native.js` so they can be used both there and there.  Does that seem sensible?  "Extracted to where?", is my next question

godber: Here it is in `native.js`:

https://github.com/terascope/teraslice/blob/v0.36.5/lib/cluster/services/cluster/backends/native.js#L462-L488

lib/cluster/services/cluster/backends/kubernetes/index.js
```diff
@@ -0,0 +1,383 @@
...
+    //     "failed": 0,
+    //     "subslices": 0,
+    //     "queued": 1,
+    //     "slice_range_expansion": 0,
+    //     "processed": 0,
+    //     "slicers": 1,
+    //     "subslice_by_key": 0,
+    //     "started": "2018-04-23T15:35:43.292-07:00"
+    // }]
+    function getSlicerStats(exIds, specificId) {
```
godber: This also needs to be refactored out of `native.js`.  I am not familiar enough to really say if that implies that there are other things that should also be extracted with this.  Doesn't look like it immediately.

lib/cluster/services/cluster/backends/kubernetes/index.js
```diff
@@ -0,0 +1,390 @@
...
+ Exceptions
+ rejected - when a job is rejected prior to scheduling
+ failed - when there is an error while the job is running
+ aborted - when a job was running at the point when the cluster shutsdown
+ */
+
+// FIXME: I want to write tests for this but I don't know how to deal with the
+// messaging stuff
+
+module.exports = function kubernetesClusterBackend(context, messaging, executionService) {
```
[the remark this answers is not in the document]
godber: @jsnoble I left this in here as a reminder to ask you if you thought it is OK that I never ended up using `executionService` in my k8s clustering implementation.  Does that seem OK?

lib/cluster/services/cluster/backends/kubernetes/index.js
```diff
@@ -0,0 +1,390 @@
...
+    const events = context.apis.foundation.getSystemEvents();
```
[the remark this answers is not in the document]
godber: @jsnoble I left this in here as a reminder to ask you if you thought it is OK that I never ended up using `events` in my k8s clustering implementation. Does that seem OK?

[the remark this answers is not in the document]
jsnoble: Yeah, the events was only used to message the 'cluster:available_workers' event which that concept does not really exist in kubernetes. I just allowed any pending workers for running executions to be allocated

lib/cluster/services/cluster/backends/kubernetes/index.js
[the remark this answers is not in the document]
jsnoble: this might be a little bit of an issue. The executionService was used as a backup to mark executions as `failed` in the 'execution:error:terminal' event. This happens on a terminal error on the execution_controller and its plausible that it would not be able to mark the state as failed itself. We don't want executions exiting without being marked. The executionService was also used to potentially terminate executions if they disconnected for to long from cluster_master. Im not sure how applicable this is in kuberenetes. What should be done if cluster_master does not have an active connection?

lib/cluster/services/cluster/backends/kubernetes/deployments/slicer.json
godber: I'm waffling a bit on my statement above.  I've always found the worker configs to be kind of redundant.  I could see a scenario where the operator just has to create the configmap for the master and then the master manages the config for the workers.

godber: One thing I need to do is destroy my already configured minikube environment and go through the setup process again and document it in a readme of some sort.  I think that will shed a little light on this.

lib/cluster/services/cluster/backends/kubernetes/index.js
godber: Ok, I'll make a note of that in the code.  This isn't actually called anywhere so I don't think I'll implement it yet.

lib/cluster/services/cluster/backends/kubernetes/index.js
```diff
@@ -0,0 +1,390 @@
...
+                return Promise.reject(error);
+            });
+    }
+
+
+    // FIXME: These functions should probably do something with the response
+    // FIXME: I find is strange that the expected return value here is
+    //        effectively the same as the function inputs
+    async function addWorkers(executionContext, numWorkers) {
+        const response = await k8s.scaleExecution(executionContext.ex_id, numWorkers, 'add');
```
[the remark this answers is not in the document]
godber: I think I need to make my `k8s` module return the full response from the k8s api, right now it just blindly returns the body in most cases.  I think the full status code and body should make it up here and these functions can then respond if they don't get the expected return value.

lib/cluster/services/cluster/backends/kubernetes/index.js
```diff
@@ -0,0 +1,390 @@
...
+     * Creates k8s Service and Deployment for the Teraslice Execution Controller
+     * (formerly slicer).  This currently works by creating a service with a
+     * hostname that contains the exId in it listening on a well known port.
+     * The hostname and port are used later by the workers to contact this
+     * Execution Controller.
+     * @param  {[type]} execution        [description]
+     * @param  {String} recoverExecution ex_id of old execution to be recovered
+     * @return {Promise}                 [description]
+     */
+    function allocateSlicer(execution, recoverExecution) {
```
godber: I need to explore the failure modes of this function.

lib/cluster/services/cluster/backends/kubernetes/index.js
```diff
@@ -0,0 +1,390 @@
...
+    }
+
+    /**
+     * Creates k8s deployment that executes Teraslice workers for specified
+     * Execution.
+     * @param  {Object} execution  Object that contains information of Execution
+     * @param  {number} numWorkers number of workers to allocate
+     * @return {Promise}           [description]
+     */
+    function allocateWorkers(execution, numWorkers) {
```
godber: I need to explore the failure modes of this function.

lib/cluster/services/cluster/backends/kubernetes/index.js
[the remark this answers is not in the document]
godber: If the worker loses its connection with the master I think the following would be true:

* would not be able to get slicer stats
* would not be able to tell workers to stop

If they become disconnected, will they try and reconnect automatically?  Is that something built into socket.io?

lib/cluster/services/cluster/backends/kubernetes/k8s.js
```diff
@@ -0,0 +1,280 @@
+'use strict';
+
+const { Client, config } = require('kubernetes-client');
+
+// FIXME: Most of my low level methods return `response.body` which hides the
```
peterdemartini: I think returning `response.body` for successful responses is appropriate, if there is a specific need for a status code, the lower level library should transform the response into something more friendly, or wrap the return value, i.e. `{"code": 200, "result": ...}`

When it comes to errors, it usually appriate to create a new Error intance and set the code to the status code.

```js
const err = new Error('Something bad happened')
err.code = response.statusCode
return Promise.reject(err)
```

lib/cluster/services/cluster/backends/kubernetes/k8s.js
```diff
@@ -0,0 +1,280 @@
...
+            } else {
+                this.logger.error(`Error: Wrong objType provided to get: ${objType}`);
+            }
+        } catch (e) {
+            this.logger.error(`Error getting ${objType} with selector ${selector}: ${e}`);
+            // throw e;
+        }
+        // FIXME: I should probably be looking at createdManifest.statusCode to
+        // make sure it's acceptible
+        return response.body;
```
peterdemartini: If it is relevant, sometimes it fair to assume any status code `400` and above is an error, and should be rejected.

```js
if (response.statusCode >= 400) {
   const err = new Error('Something bad happened')
   err.code = response.statusCode
   return Promise.reject(err)
}
```

lib/cluster/services/cluster/backends/kubernetes/k8s.js
```diff
@@ -0,0 +1,280 @@
...
+    // on this to require `objType` to support patching other things
+    async patch(record, name) {
+        let response;
+
+        try {
+            response = await this.client.api.apps.v1.namespaces('default')
+                .deployments(name).patch({ body: record });
+        } catch (e) {
+            this.logger.error(`Error trying to patch ${name}: ${e}`);
+            // FIXME: What is the appropriate way to handle errors in this case?
```
peterdemartini: I would say, something like:
```js
const err = new Error(`Error trying to patch ${name}: ${e}`)
this.logger.error(err.stack)
return Promise.reject(err)

File: lib/cluster/services/cluster/backends/kubernetes/k8s.js [hunk 20]
```diff
@@ -0,0 +1,280 @@
...
+        this.logger.info(`New Scale for exId=${exId}: ${newScale}`);
+
+        const scalePatch = {
+            spec: {
+                replicas: newScale
+            }
+        };
+
+        const patchResponseBody = await this.patch(scalePatch, workerDeployment.metadata.name);
+        // FIXME: This returns after waiting on the k8s stuff, should it?  I
```
peterdemartini: I believe it should wait if possible unless this particular action takes unreasonably long. If there is a case not waiting, just logging the error isn't sufficient because other components may be dependent on the success or failure of this case so there should be an appropriate way of getting the state of the request or cancelling it, which is more complicated.

lib/cluster/services/cluster/backends/kubernetes/k8sState.js
```diff
@@ -0,0 +1,55 @@
...
+    // avoid looping over the list of pods twice if possible.
+    k8sPods.items.forEach((element) => {
+        clusterState[element.status.hostIP] = {
+            node_id: element.status.hostIP,
+            hostname: element.status.hostIP,
+            pid: 1,
+            node_version: 'N/A',
+            teraslice_version: 'N/A',
+            total: 42,
+            state: 'connected', // FIXME: I had set this to
```
peterdemartini: Yes, this will need to be mapped to known teraslice state statuses.

spec/lib/cluster/services/cluster/backends/kubernetes/k8s-spec.js
```diff
@@ -0,0 +1,197 @@
...
+// output for nock
+//   env DEBUG='nock*' make test
+
+const fs = require('fs');
+const K8s = require('../../../../../../../lib/cluster/services/cluster/backends/kubernetes/k8s');
+const nock = require('nock');
+const path = require('path');
+
+function print(err, result) {
+    // FIXME: How much does this smell, do we have an existing convention?
```
peterdemartini: I would say use the debug library, `yarn add --dev debug`

```js
const debug = require('debug');

const logger = {
    error(...args) {
        debug('teraslice-kubernetes-k8-spec:error')(...args);
    },
    warn(...args) {
        ([warnMsg] = args);
        debug('teraslice-kubernetes-k8-spec:warn')(...args);
    },
    info(...args) {
        debug('teraslice-kubernetes-k8-spec:info')(...args);
    },
    trace(...args) {
        debug('teraslice-kubernetes-k8-spec:trace')(...args);
    },
    debug(...args) {
        debug('teraslice-kubernetes-k8-spec:debug')(...args);
    },
    flush() {
        return Promise.resolve();
    }
};
```

Now when you run your tests set the env `DEBUG='teraslice-kubernetes-k8-spec:*'`

lib/cluster/services/cluster/backends/kubernetes/index.js
```diff
@@ -0,0 +1,392 @@
...
+     * @param  {number} timeout Timeout for messaging the nodes in FIXME: units and meaning
+     * @return {Promise}
+     */
+    async function stopExecution(exId, timeout) {
+        const stopMessage = { message: 'cluster:execution:stop' };
+        if (timeout) {
+            stopMessage.timeout = timeout;
+        }
+        const workers = _iterateState(worker => worker.ex_id === exId);
+        const msgResponse = await _msgWorkersWithExId(exId, stopMessage, workers);
```
godber: I might be able to just delete this `_msgWorkersWithExId` line because when `deleteExecution` is called the workers should just stop cleanly.

lib/cluster/services/cluster/backends/kubernetes/index.js
```diff
@@ -0,0 +1,307 @@
...
+
+    function resumeExecution(exId) {
+        const workers = stateUtils._iterateState(
+            clusterState,
+            worker => worker.assignment === 'execution_controller' && worker.ex_id === exId
+        );
+        return _msgWorkersWithExId(exId, { message: 'cluster:execution:resume' }, workers);
+    }
+
+    const getSlicerStats = stateUtils.newGetSlicerStats(clusterState, context, messaging);
```
godber: Now that I've extracted the `getSlicerStats` function it's `clusterState` is ending up as `{}` even after it should have nodes and workers and whatnot in it.  This works fun in the native cluster scenario where cluster state is updated by message handlers, but in this case it is just rebuild on a timer.

godber: I bet the problem is the reassignment on line 80.

lib/cluster/services/cluster/backends/kubernetes/k8sState.js
```diff
@@ -9,35 +9,29 @@
...
+    // Loop over the nodes in clusterState and set active = [] so we can append
+    // later
+    Object.keys(clusterState).forEach((nodeId) => {
+        clusterState[nodeId].active = [];
     });
 
-    // next add workers
+    // add a worker for each pod
     k8sPods.items.forEach((element) => {
+        if (!Object.prototype.hasOwnProperty.call(clusterState, element.status.hostIP)) {
```
peterdemartini: It might make be easier to use lodash, since they have methods like `_.has` or `_.set` which are easier to use and understand.

godber: Many thanks for the suggestion, will try to remember to use it.

Dockerfile
```diff
@@ -21,6 +21,6 @@ EXPOSE 5678
 
 VOLUME /app/config /app/logs
 
-ENV TERAFOUNDATION_CONFIG /app/config/processor-master.yaml 
+ENV TERAFOUNDATION_CONFIG /app/config/teraslice.yaml
```
godber: Note that this changes the path to the default config file in the Docker container.  I think `teraslice.yaml` is preferable to `processor-master.yaml` ... what if your node isn't a master??

lib/cluster/services/cluster/backends/kubernetes/index.js
godber: The problem above has been resolved.

docs/k8s-clustering.md
```diff
@@ -0,0 +1,108 @@
...
+You need Elasticsearch running and listening on a port accessible by the
+Teraslice master and worker nodes.  In the case of minikube based deployments
+(dev/test only) you'll need to make ES listen on your VirtualBox interface.  The
+IP address of that interface is what you need to set as:
+
+    `terafoundation.connectors.elasticsearch.default.host`
+
+in both the `teraslice-master` and `teraslice-worker` ConfigMaps.
+
+### ConfigMaps
```
peterdemartini: What is the file name for the and location for ConfigMaps?

examples/k8s/Makefile
```diff
@@ -0,0 +1,42 @@
...
+	kubectl logs -l app=teraslice,nodeType=master | bunyan
+
+logss: ## show logs for k8s teraslice slicers
+	kubectl logs -l app=teraslice,nodeType=execution_controller | bunyan
+
+logsw: ## show logs for k8s teraslice workers
+	kubectl logs -l app=teraslice,nodeType=worker | bunyan
+
+submit: ## submit test job
+	@curl -sS -XPOST 192.168.99.100:30678/jobs -d@./sample-job.json
```
peterdemartini: This IP address should be dynamic

examples/k8s/Makefile
```diff
@@ -0,0 +1,42 @@
...
+	    printf "  %s  :%s\n" "$$var" "$$help"; \
+	  done \
+	done
+
+show: ## show k8s deployments and services
+	kubectl get deployments,svc,po -o wide
+
+destroy: ## delete k8s deployments and services
+	kubectl delete deployments,services -l app=teraslice
+	curl -XDELETE -Ss '192.168.99.1:9200/teracluster__*'
```
peterdemartini: This IP address should be dynamic

spec/lib/cluster/services/cluster/backends/kubernetes/files/job-running-v1-clusterState.json
```diff
@@ -0,0 +1,39 @@
+{
+  "192.168.99.100": {
```
peterdemartini: This IP probably be something different? If it is a test, we should put, 127.0.0.1

godber:
This branch needs modification to work on a k8s cluster that requires authentication and or encryption.

The teraslice cluster master should probably die when encountering this error:

```
[2018-07-06T19:20:50.258Z] ERROR: teraslice/12 on teraslice-master-5f874b475-62dv5: Error calling k8s loadSpec: Error: unable to get issuer certificate (module=kubernetes_cluster_service)
```

if not, it continues and just get's these perpetually:

```
[2018-07-06T19:20:55.186Z] ERROR: teraslice/12 on teraslice-master-5f874b475-62dv5: (module=kubernetes_cluster_service)
    Error: Request k8s.list of pods with selector app=teraslice failed: TypeError: Cannot read property 'v1' of undefined
        at K8s.list (/app/source/lib/cluster/services/cluster/backends/kubernetes/k8s.js:92:25)
        at _getClusterState (/app/source/lib/cluster/services/cluster/backends/kubernetes/index.js:77:20)
        at Timeout.setInterval [as _onTimeout] (/app/source/lib/cluster/services/cluster/backends/kubernetes/index.js:59:9)
        at ontimeout (timers.js:498:11)
        at tryOnTimeout (timers.js:323:5)
        at Timer.listOnTimeout (timers.js:290:5)
```

lib/cluster/services/cluster/backends/kubernetes/index.js
```diff
@@ -0,0 +1,305 @@
+'use strict';
```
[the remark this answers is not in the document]
godber: I don't know how to get rid of the 'use strict'; issue, but the rest of your requested changes have been made.

peterdemartini:
@godber while testing on minikube, I got the error when running a job `User "system:serviceaccount:default:default" cannot create services in the namespace "default"`

lib/cluster/services/cluster/backends/kubernetes/deployments/slicer.json
```diff
@@ -0,0 +1,69 @@
...
+            "metadata": {
+                "labels": {
+                    "app": "teraslice",
+                    "nodeType": "execution_controller",
+                    "exId": "<EXID>"
+                }
+            },
+            "spec": {
+                "containers": [{
+                    "image": "teraslice:k8sdev",
```
peterdemartini: This image name should be dynamic. Make something like`process.env.TERASLICE_K8_SLICER_IMAGE || process.env.TERASLICE_K8_IMAGE || 'teraslice:k8sdev'`

lib/cluster/services/cluster/backends/kubernetes/deployments/worker.json
```diff
@@ -0,0 +1,69 @@
...
+            "metadata": {
+                "labels": {
+                    "app": "teraslice",
+                    "nodeType": "worker",
+                    "exId": "<EXID>"
+                }
+            },
+            "spec": {
+                "containers": [{
+                    "image": "teraslice:k8sdev",
```
peterdemartini: This image name should be dynamic. Make something like`process.env.TERASLICE_K8_WORKER_IMAGE || process.env.TERASLICE_K8_IMAGE || 'teraslice:k8sdev'`

examples/k8s/Makefile
```diff
@@ -0,0 +1,44 @@
...
+	kubectl logs -l app=teraslice,nodeType=worker | bunyan
+
+submit: ## submit test job
+	curl -sS -XPOST $(TS_MASTER_URL)/jobs -d@./sample-job.json
+
+k8s-master: ## start teraslice master in k8s
+	kubectl create -f ./masterDeployment.yaml
+
+build: ## build the teraslice:k8sdev container
+	docker build -t teraslice:k8sdev ../../
```
peterdemartini: It would be more flexible to use something like `docker build -t $(TERASLICE_K8_IMAGE) ../../` and set a default for it `TERASLICE_K8_IMAGE ?= teraslice:k8sdev`

examples/k8s/Makefile
```diff
@@ -0,0 +1,44 @@
...
+show: ## show k8s deployments and services
+	kubectl get deployments,svc,po -o wide
+
+destroy: ## delete k8s deployments and services
+	kubectl delete deployments,services -l app=teraslice
+
+logs: ## show logs for k8s deployments and services
+	kubectl logs -l app=teraslice --tail=10 | bunyan
+
+logsm: ## show logs for k8s teraslice master
```
peterdemartini: I suggest we use the long form for these log commands. For example, `logs-master`, `logs-worker` and `logs-slicer`, it is easier to read and less human error prone.

lib/cluster/services/cluster/backends/kubernetes/deployments/slicer.json
godber: Yeah, you're going to find a lot of these "foo needs to be made dynamic".  For each one the question is then going to be "How does the user or operator provide this information dynamically?"  The answers are going to be some combination of the following:

* Provided at cluster configuration
* Provided by job

godber: Your suggestion of using the env variables is interesting, but keep in mind teraslice is responsible for creating the deployment it can just read it from the job or cluster config as well.  There are lots of options.

godber:
@peterdemartini could you paste the output of the following commands:

`kubectl version --short`

`minikube version`

peterdemartini:
```
Client Version: v1.11.0
Server Version: v1.10.0
minikube version: v0.28.0
```

godber:
This branch doesn't quite work in our real test cluster (not minikube).  It has the following issues:

* Submitted job immediately goes to failed state, no errors on job
* Cluster state comes back as `{}` and a bunch of related errors (things not showing up in API.

godber:
@peterdemartini I was able to recreate your issue in minikube just by upgrading all my things to the latest versions and starting the k8s cluster and then the teraslice cluster as my notes show.  I think my minikube was several months old and running k8s without any RBAC enabled.

If you were to create a file named `teraslice-default-binding.yaml` with the following contents:

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: teraslice-default-binding
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cluster-admin
subjects:
- kind: ServiceAccount
  name: default
  namespace: default
```

and then apply that with the command: `kubectl create -f teraslice-default-binding.yaml`

You can then run:

```
make destroy  # if teraslice is still running
make k8s-master
```

And you should see teraslice working, cluster state won't be empty and the api endpoints will work as expected.

The problem is, the snippet above gives the default service account in the default namespace admin privileges.  So basically any pod can do anything in the default namespace.  If the teraslice clusters were in their own namespaces, a minor change to that file above would constrain their abilities to within that namespace.

lib/cluster/services/cluster/backends/kubernetes/deployments/slicer.json
godber: A solution for this turns out to be necessary to deploy from an internal docker registry, so I need to address this now.

godber:
I am going to merge this, I am pushing one more thing, but it's only changes in the makefile.  I won't wait for the checks to complete after the next commit.

lib/cluster/services/cluster/backends/state-utils.js
```diff
@@ -63,15 +65,17 @@ function newGetSlicerStats(clusterState, context, messaging) {
...
-                    address: slicer.node_id,
+                    // use pod_ip as address (k8s), otherwise, use node_id
+                    address: _.get(slicer, 'pod_ip', slicer.node_id),
                     message: 'cluster:slicer:analytics',
-                    payload: msg,
+                    payload: { ex_id: slicer.ex_id },
                     response: true
-                }));
+                };
+                nodeQueries.push(messaging.send(msg));
```
godber: The changes in this file will impact teraslice execution outside of kubernetes mode.

lib/cluster/node_master.js
```diff
@@ -322,8 +322,21 @@ module.exports = function module(context) {
...
+                // this enviroment variable needs to be set by the deployment
+                nodeId = process.env.POD_IP;
+            }
+        } else {
+            nodeId = context.sysconfig._nodeName;
+        }
+
         const state = {
-            node_id: context.sysconfig._nodeName,
+            node_id: nodeId,
```
godber: The changes in this file will impact teraslice execution outside of kubernetes mode.

PR #743, opened 2018-06-18
job updates now use inexedWithId for full doc replacements resolves #737

### LINKED ISSUE #737: job properties are not automatically set to the default if previously registered on the cluster
Jobs that do not explicitly set a property will retain the same value for that property from the job file on the cluster.

For example register a job that specifies 5 slicers, then remove the slicer property, update the job again and view the job on the cluster.  It will still have 5 slicers.  This can be confusing since it doesn't return to the default if the property has been removed.

I was tricked by this thinking that a job was using one slicer after I had removed the slicer property and updated it, but in reality the job was running with 5 slicers.  This also applies to workers, analytics, lifecycle, etc...

jsnoble:


kstaken:
Ready to merge pending TravisCI checks.

Issue #745, opened 2018-06-19
Extract clusterState related functions from native.js


godber:
My k8s pr (#738) for issue #511 copies the following things from `native.js`:

* `_iterateState`
* `findSlicersByExecutionID`
* `getSlicerStats`

There are related functions in `native.js` that should probably be extracted as well, even though they are not in use by the k8s backend.

godber:
I'd like to implement this separately from the k8s branch because I am afraid the other branch might diverge and I don't want to keep having to rebase this.

PR #747, opened 2018-06-21
Issue 745 extract iterate state

godber:
Extract iterateState and friends out to stateUtils

My pending k8s backed cluster manager will need to share the
iterateState function and its associated functions so I extract them out
here.

closes https://github.com/terascope/teraslice/issues/745


lib/cluster/services/cluster/backends/stateUtils.js
```diff
@@ -0,0 +1,111 @@
...
+                    clearTimeout(timer);
+                    const formatedData = results.map(formatResponse);
+                    const sortedData = _.sortBy(formatedData, ['name', 'started']);
+                    const reversedData = sortedData.reduce((prev, curr) => {
+                        prev.push(curr);
+                        return prev;
+                    }, []);
+                    resolve(reversedData);
+                })
+                .catch(err => reject({ message: parseError(err), code: 500 }));
```
peterdemartini: We should clear the timeout in the catch as well.

lib/cluster/services/cluster/backends/stateUtils.js
```diff
@@ -0,0 +1,111 @@
+'use strict';
```
peterdemartini: We should rename this `state-utils.js`

spec/lib/cluster/services/cluster/backends/stateUtils-spec.js
```diff
@@ -0,0 +1,103 @@
...
+
+            const getSlicerStats = stateUtils.newGetSlicerStats(clusterState, context, messaging);
+
+            try {
+                slicerStats = await getSlicerStats(
+                    ['8d7669e1-d351-4a0d-b36e-b11da5ade26a'],
+                    '8d7669e1-d351-4a0d-b36e-b11da5ade26a'
+                );
+            } catch (e) {
+                console.log(e);
```
peterdemartini: We should use expect to make sure this error shouldn't happen. For example, `expect(e).not.toBeDefined()`

godber:
Thanks for the feedback Peter, I've pushed my changes.

PR #748, opened 2018-06-28
fixes for asset builds

jsnoble:
There are some issues to be talked about before this is brought in. Moving out processors from teraslice and into assets brought to light some inconsistencies we have with how we use the context.apis functionality. Also, since we are having the cluster_master do validation checks on the submitted jobs, it was having to load modules that supposed certain things existed which where not true. Different context.apis where available to different parts of a processor file which is not good. We need to better evaluate the proper api signature of these functions have have a discussion on how to load these apis more independently and to every process.

### DIFF
--- a/lib/cluster/runners/execution.js
+++ b/lib/cluster/runners/execution.js
@@ -5,13 +5,12 @@ const parseError = require('@terascope/error-parser');
 const moment = require('moment');
 const _ = require('lodash');
 const path = require('path');
-const existsSync = require('../../utils/file_utils').existsSync;
+const { existsSync } = require('../../utils/file_utils');
 
 /*
  * This module defines the job execution context on the worker nodes.
  */
-module.exports = function module(context) {
-    const opRunner = require('./op')(context);
+module.exports = function module(context, justRegisterApi) {
     // used for testing so we dont have to pollute the process env in testing
     const processAssingment = context.__test_assignment || process.env.assignment;
     const isSlicer = processAssingment === 'execution_controller';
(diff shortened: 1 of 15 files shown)


lib/cluster/runners/execution.js
```diff
@@ -20,21 +19,17 @@ module.exports = function module(context) {
...
+    function getOpConfig(job, name) {
+        return job.operations.find(op => op._op === name);
     }
 
-    /*
-     * This sets up the APIs intended for use by custom readers / processors.
-     */
-
     context.apis.registerAPI('job_runner', {
         getOpConfig
```
peterdemartini: Making this API change is a breaking change, and I don't think it is necessary. Why not keep this file the same and in `lib/config/validators/job.js` we add duplicate registering the getOpConfig API (see my other comment)?

lib/config/validators/job.js
```diff
@@ -6,13 +6,15 @@ const convictFormats = require('../../utils/convict_utils');
...
-module.exports = function (_context) {
+module.exports = (_context) => {
     let context = _context;
     const commonSchema = require('../schemas/job').commonSchema();
     let opRunner;
     let jobSchema;
 
     if (context) {
+        // We need a better story around registering apis
+        require('../../cluster/runners/execution')(context, true);
```
peterdemartini: I think doing something like this would allow us to not make a breaking change.

```js
context.apis.registerAPI('job_runner', {
         getOpConfig: (name) => {
             return job.operations.find(op => op._op === name)
         }
})
```

jsnoble:
We want to work with backward compatibility first and add code duplication to assets so that we have time to settle on the apis interface. We will shutdown this pr for now

Issue #755, opened 2018-07-09
Updating a job deletes the job


kstaken:
In Teraslice 0.37.0 updating a job appears to delete the job from the system.

kstaken:
Fixed in #756

PR #756, opened 2018-07-09
added important meta data back to updated job

jsnoble:
the fields: job_id (which is used for searching and other vital queries) and _created (which is used for sorting) where missing. This pr adds them back in


kstaken:
fixes #755

PR #761, opened 2018-07-16
Worker disconnect fixes, resolves #759

### LINKED ISSUE #759: Worker disconnect and reconnect issues
I have a job that blocks the event loop for 2 minutes, which will cause the socket to disconnect on a worker. After the job is done blocking, a few things happen when the worker reconnects: the worker sends the `worker:slice:complete` message, the slicer enqueues the worker and sends it a new slice. Slightly after that, the worker sends another `worker:slice:complete` message, with `retry` set to true, when the `network:connect` event fires, which also causes the slicer to enqueue the worker and send a new slice. Both of those new slice messages end up getting a timeout error on the slicer.

I was able to replicate this on version `0.37.0`, `0.37.1` and the master branch of teraslice. Using socket.io version `1.4.5` did not seem to make a difference.

peterdemartini:
I found a couple issues with workers that disconnect and reconnecting, particularly around race conditions and double `slice:completed` messages.

In addition to this fix, I updated the npm token for deploys, and made workers shutdown more gracefully.


peterdemartini:
We are currently testing this PR with the latest version socket.io and it seems to be operating well. If it continues to work, I will update socket.io in this PR.

lib/cluster/execution_controller/engine.js
```diff
@@ -80,40 +81,46 @@ module.exports = function module(context, messaging, exStore, stateStore) {
...
         callback: (msg, workerId) => {
             const workerResponse = msg.payload;
-            executionAnalytics.increment('processed');
+            const sliceId = _.get(workerResponse, 'slice.slice_id');
+
+            const cachekey = JSON.stringify(_.pick(workerResponse, ['slice', 'worker_id']));
+            const alreadyCompleted = _.get(_completedSlicesCache, cachekey, false);
+            const shouldEnqueue = !workerResponse.isShuttingDown;
+
+            _.set(_completedSlicesCache, cachekey, true);
```
jsnoble: This is an unbounded memory leak. For every slice that is completed it is added to the cache. We have executions that never finish and go through several hundreds of millions of slices. This will cause it to eventually crash which we can't afford as this is the "brain" of the execution. You could instead key the cache by the worker identifier instead (which is something we are going to need real soon already to know which worker did what).  Even though this still potentially could have stale caches left when workers are removed, it is still relatively bounded. This would not be above a few hundred tops.

lib/cluster/node_master.js
```diff
@@ -234,7 +234,7 @@ module.exports = function module(context) {
         callback: () => {
             logger.error('terminal error in cluster_master, flushing logs and shutting down');
             logger.flush()
-                .then(() => process.exit());
+                .then(() => process.exit(1));
```
jsnoble: We cannot exit properly as noted in the statement above.
https://github.com/terascope/terafoundation/blob/master/lib/master.js#L78-L92

lib/cluster/worker/index.js
```diff
@@ -63,11 +63,11 @@ module.exports = function module(context) {
                     error: errMsg
                 })
                     .then(() => logger.flush())
-                    .then(() => process.exit())
+                    .then(() => process.exit(0))
                     .catch((flushErr) => {
                         const flushErrMsg = parseError(flushErr);
                         logger.error(flushErrMsg);
-                        process.exit();
+                        process.exit(1);
```
jsnoble: Same issues as noted above
https://github.com/terascope/terafoundation/blob/master/lib/master.js#L78-L92

lib/cluster/worker/index.js
```diff
@@ -79,20 +79,37 @@ module.exports = function module(context) {
...
+            })
             .then(logger.flush)
             .catch((err) => {
+                exitCode = 1;
                 const errMsg = parseError(err);
                 logger.error(`Error while attempting to shutdown worker ${ID}, error: ${errMsg}`);
                 return logger.flush();
             })
-            .finally(process.exit);
+            .finally(() => process.exit(exitCode));
```
jsnoble: same issues as noted above
https://github.com/terascope/terafoundation/blob/master/lib/master.js#L78-L92

lib/cluster/execution_controller/engine.js
```diff
@@ -260,6 +269,7 @@ module.exports = function module(context, messaging, exStore, stateStore) {
     }
 
     function _sendSlice(sliceData, address) {
+        logger.info(`Sending slice to ${address}`);
```
jsnoble: This is nice to have, but im not sure this is something of the info level but at the debug level. We already hear how big the log files for teraslice get and we need to take an inventory about what we log. We should ask ops people what they think about this. It would be usefull

lib/cluster/execution_controller/engine.js
```diff
@@ -122,10 +129,11 @@ module.exports = function module(context, messaging, exStore, stateStore) {
         identifier: 'worker_id',
         callback: (msg, workerId) => {
             executionAnalytics.increment('workers_disconnected');
-            logger.warn(`Worker: ${workerId} has disconnected for reason ${msg}`);
+            const reason = !_.isEmpty(msg) ? ` for reason ${JSON.stringify(msg)}` : '';
+            logger.warn(`Worker: ${workerId} has disconnected${reason}`);
```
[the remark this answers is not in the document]
peterdemartini: Sometimes there isn't a human-readable reason, just an empty object, so I moved the spaces to Line 132 when formatting the reason.

jsnoble:
LGTM

PR #770, opened 2018-07-31
WIP: Re-useable components & error handling

peterdemartini:
Make Teraslice modules and components more resuable from external projects, like the [Teraslice Worker](https://github.com/terascope/teraslice-worker).

### DIFF
--- a/lib/cluster/services/execution.js
+++ b/lib/cluster/services/execution.js
@@ -155,19 +155,26 @@ module.exports = function module(context) {
                 })
                 .catch((err) => {
                     const errMsg = parseError(err);
-                    logger.error('could not set to pending', errMsg);
-                    return Promise.reject(errMsg);
+                    const error = new Error(`could not set to pending ${errMsg}`);
+                    logger.error(error.message);
+                    return Promise.reject(error);
                 }))
             .catch((err) => {
                 const errMsg = parseError(err);
-                logger.error('could not create execution context', errMsg);
-                return Promise.reject(errMsg);
+                const error = new Error(`could not create execution context ${errMsg}`);
+                logger.error(error.message);
+                return Promise.reject(error);
             });
     }
 
     function getExecutionContext(exId) {
         return exStore.get(exId)
-            .catch(err => logger.error(`error getting execution context for ex: ${exId}`, err));
+            .catch((err) => {
+                const errMsg = parseError(err);
+                const error = new Error(`error getting execution context for ex: ${exId} ${errMsg}`);
+                logger.error(error);
+                return Promise.reject(error);
+            });
     }
 
     function getRunningExecutions(exId) {
@@ -184,7 +191,8 @@ module.exports = function module(context) {
         return searchExecutionContexts(query, null, 1, '_created:desc')
             .then((ex) => {
                 if (ex.length === 0) {
-                    return Promise.reject(`no active execution context was found for ex_id: ${exId}`);
+                    const error = new Error(`no active execution context was found for ex_id: ${exId}`);
+                    return Promise.reject(error);
                 }
                 return ex[0];
             });
@@ -209,6 +217,9 @@ module.exports = function module(context) {
     }
 
     function _canRecover(ex) {
+        if (!ex) {
+            throw new Error('Cannot recover invalid execution');
+        }
         if (ex._status === 'completed') {
             throw new Error('This job has completed and can not be restarted.');
         }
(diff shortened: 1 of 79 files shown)


lib/cluster/services/execution.js
```diff
@@ -155,19 +155,26 @@ module.exports = function module(context) {
...
+                return Promise.reject(error);
             });
     }
 
     function getExecutionContext(exId) {
         return exStore.get(exId)
-            .catch(err => logger.error(`error getting execution context for ex: ${exId}`, err));
+            .catch((err) => {
+                const errMsg = parseError(err);
+                const error = new Error(`error getting execution context for ex: ${exId} ${errMsg}`);
```
jsnoble: same as above

lib/cluster/services/execution.js
```diff
@@ -184,7 +191,8 @@ module.exports = function module(context) {
         return searchExecutionContexts(query, null, 1, '_created:desc')
             .then((ex) => {
                 if (ex.length === 0) {
-                    return Promise.reject(`no active execution context was found for ex_id: ${exId}`);
+                    const error = new Error(`no active execution context was found for ex_id: ${exId}`);
```
jsnoble: This returns to the api side, and this is not a real error. There was no error in code execution or database call. I dont think throwing an error with a stack trace is the right move here. Again, does a database call throw a stack trace for a document not found error in the response?

lib/cluster/services/execution.js
```diff
@@ -209,6 +217,9 @@ module.exports = function module(context) {
     }
 
     function _canRecover(ex) {
+        if (!ex) {
+            throw new Error('Cannot recover invalid execution');
```
jsnoble: Im not sure this guard is needed as getExecutionContext throws an error if the document was not found. If we keep it then maybe a better error as this is an issue of nothing vs something that is invalid which to me signals there was a parsing/schema issue

lib/cluster/services/execution.js
```diff
@@ -254,7 +265,11 @@ module.exports = function module(context) {
                 if (cleanup) execution.recovered_slice_type = cleanup;
                 return createExecutionContext(execution);
             })
-            .catch(err => Promise.reject(parseError(err)));
+            .catch((err) => {
+                const errorMsg = parseError(err);
+                const error = new Error(`Unable to recover execution ${errorMsg}`);
```
jsnoble: same issue above of parsing an error to only make a new error with different stack traces

lib/cluster/services/execution.js
```diff
@@ -329,38 +344,36 @@ module.exports = function module(context) {
...
+                }))
             .error((err) => {
             // TODO: verify whats coming here
                 if (_.get(err, 'body.error.reason') !== 'no such index') {
                     logger.error(`initialization failed loading state from Elasticsearch: ${err}`);
                 }
                 const errMsg = parseError(err);
-                logger.error('Error initializing, ', errMsg);
-                return Promise.reject(errMsg);
+                const error = new Error(`Error initializing, ${errMsg}`);
```
jsnoble: same as above

lib/cluster/storage/backends/elasticsearch_store.js
```diff
@@ -177,9 +180,29 @@ module.exports = function module(context, indexName, recordType, idField, _bulkS
...
 
-    function shutdown() {
+    function shutdown(forceShutdown) {
         clearInterval(flushInterval);
-        return _flush();
+        if (forceShutdown !== true) {
+            return _flush();
+        }
+        return new Promise((resolve) => {
+            logger.info(`attempting to shutdown, will destroy in ${config.shutdown_timeout}`);
```
jsnoble: Im not sure that the shutdown logic is as it should be. As it stands, when trying to shutdown the module there are multiple different id's added to the destroyFns object for a given store. If you catch it from the first isAvailable then its an id, but if it goes to the catch it creates another id then deletes it then goes to isAvailable again and creates another id so you have to spread your logic of creating an id and deleting it around the whole file. It makes it difficult to understand what is going on and prone to errors and race conditions. Also in general (i know it was like this before you change things) we need to guard agains an async call inside an interval. This can cause to have multiple calls in flight and cause race conditions. Im referencing line 379 and other parts of the code that does this.

lib/cluster/worker/executor.js
```diff
@@ -164,6 +164,8 @@ module.exports = function module(context, messaging, stateStore, analyticsStore,
 
         _executeSlice(slice, operations, sliceLogger, specData)
             .catch(retrySliceFn);
+
+        return null; // See http://goo.gl/rRqMUw
```
jsnoble: this is not used in a promise. this is a sync function call and does not have to do with bluebird

Issue #777, opened 2018-08-10
The execution_controller remains after a Teraslice job is stopped.


godber:
Teraslice currently removes the k8s `execution_controller` Job when a teraslice job is stopped or completes.  The pods remain even after the job is deleted.  In the case of a job completing, the pod doesn't show up in `/txt/workers` because it is not in the running state.  In the case of Stop being called, the pod remains in the running state, so it shows up in `/txt/workers.

PR #778, opened 2018-08-10
Ensure execution_controller pods are remove on job stop

godber:
This is called the Cascading Deletion Policy by k8s.  Basically we just
need to pass a body with the request that will ensure pods get deleted
too.  I suppose we might encounter conditions where we don't want the
pods deleted when the job is, but for now, we'll force that.

closes: #777


godber:
Note I've updated the version in `package.json` here to help with a speedy release.

PR #781, opened 2018-08-15
Separate components and fix circular dependencies

### LINKED ISSUE #784: terasilce-worker slice analytics missing ex_id and job_id
The original worker creates analytics records like:

```
{
  "@timestamp": "2018-06-19T18:05:28.176Z",
  "ex_id": "ec92b54c-8232-4e12-bd94-060f861e21b0",
  "job_id": "87779ae9-c2bc-4f0a-925e-82f11428d90e",
  "worker_id": "10.0.0.1__1",
  "slice_id": "07a6484d-9909-4ed0-bf0b-d40c0a738707",
  "slicer_id": 0,
  "op": "teraslice_kafka_reader",
  "order": 0,
  "count": 1000,
  "time": 37905,
  "memory": 149584
}
```

and the new worker creates records like:

```
{
  "@timestamp": "2018-08-14T23:48:37.510Z",
  "worker_id": "10.1.0.1__E6vadmCWM63WJoV",
  "slice_id": "c836830e-dddb-468f-9f83-02ac6f51e430",
  "slicer_id": 0,
  "op": "example-op",
  "order": 1,
  "count": 5000,
  "time": 0,
  "memory": 696
}
```

Looks like it's just `job_id` and `ex_id` missing.

peterdemartini:
* Breakout Operation and Validation into modules (using TypeScript)
* Rename `entrypoint.js` and `command.js` to `service.js`
* Fix Deploy issues with Lerna
* Export `index.js` in teraslice (temporary for `teraslice-worker`)
* Update documentation
* A few minor bug fixes
* Resolves #784
* Switches from jasmine to jest


packages/docker-compose-js/test/docker-compose-spec.js
```diff
@@ -3,19 +3,19 @@
...
-    beforeEach(() => {
+    beforeAll(() => {
         sut = compose(path.join(__dirname, 'fixtures', 'example.yaml'));
     });
 
     it('should be able to call compose.pull()', (done) => {
         sut.pull()
-            .then(done)
-            .catch(done.fail);
+            .then(() => { done(); })
```
jsnoble: is there any reason why we are using done in this manner `.then(() => { done(); })`? In all the other tests we generally do .finally(done) or .then(done)

packages/teraslice-operations/src/operation-loader.ts
```diff
@@ -0,0 +1,113 @@
...
+        this.verifyOpName(name);
+
+        let filePath: string|null = null;
+        let codeName: string = '';
+
+        if (!name.match(/.js$/)) {
+            codeName = `${name}.js`;
+        }
+
+        const findCode = (rootDir: string): void => {
```
jsnoble: We generally write our functions using declarations (ie function foo(){}  ) and we only tend to use anonymous functions as callbacks to things like map. It would be good to remain consistent

packages/teraslice-worker/test/lib/execution-context/execution-context.spec.js
```diff
@@ -54,19 +54,13 @@ describe('Execution Context', () => {
         });
```
jsnoble: I see different tests with `{name}-spec.js` and `{name}.spec.js`  What is the correct way to do so? Either way we should have a consistent way to name test files. Btw there are several other tests files besides this that are named differently

packages/teraslice/index.js
```diff
@@ -0,0 +1,57 @@
...
+        schema_formats: formats,
+        ops_directory: opsDirectory,
+        cluster_name: clusterName,
+        logging_connection: loggingConnection
+    }, sysconfig);
+}
+
+module.exports = {
+    dateFormat,
+    terasliceOpPath,
```
jsnoble: Should we be also exposing the lower level store class so that others can make their store? I know in teraserver I have a user store, and other things might want to create their own store. However im not sure how much we want to expose this?

packages/docker-compose-js/test/docker-compose-spec.js
peterdemartini: Because in jest, the first parameter passed to done is an error.

packages/teraslice-operations/src/operation-loader.ts
peterdemartini: I know but then I lose “this” binding

packages/docker-compose-js/test/docker-compose-spec.js
jsnoble: hmm, then is jest not running the other tests? we are not seeing errors there and I though you switched over all to jest. Are they switched or do we still have jasmine?

peterdemartini: I updated the rest of the tests. It was an easy search and replace.

packages/teraslice/index.js
peterdemartini: I don't think a lot of these export for a long time since they are just be used only for teraslice-worker. Would be better if I add a comment saying declaring the intent of these exports?

jsnoble:
Tested on localhost, both test suites pass. Concerns were addressed

Issue #786, opened 2018-08-16
Support for running in configurable k8s namespace


godber:
We need to be able to run teraslice in a specified namespace.

PR #787, opened 2018-08-18
Issue 786 namespace support

godber:
This is really straight forward.

Most of the work was in building the Makefile (private) that sets up and tears down k8s environments with namespaces.  Maybe I should try and jam that into the `examples/k8s` directory.


godber:
refs: #786

peterdemartini:
We should probably add the setup in the example Makefile.

godber:
I've added the Makefile and associated things, I am working on updating the `docs/k8s-clustering.md`

PR #790, opened 2018-08-20
Fix test coverage issues and consolidate a couple internal packages

peterdemartini:
* Jest coverage fixes
* monorepo fixes
* Consolidate teraslice-operations and teraslice-validations into one package (job-components)
* Stub out some of the new job APIs interface (but they aren't used or exposed).
* Test recovery modes separately


e2e/test/cases/data/reindex-spec.js
```diff
@@ -72,7 +72,7 @@ describe('reindex', () => {
             .finally(() => { done(); });
     });
 
-    it('should complete after lifecycle changes', (done) => {
+    fit('should complete after lifecycle changes', (done) => {
```
jsnoble: should revert the `fit` to `it`

PR #798, opened 2018-08-22
changes to work with preserve_id

jsnoble:


jsnoble:
removing this becuase it was decided to only encasuplate this in the elasticsearch-asset

Issue #794, opened 2018-08-21
Add support for targeting jobs to specific nodes - affinity


godber:
There are a few conditions where users may want to force pods to run on specific nodes or subsets of nodes.  For example:

* When a user has a k8s cluster that spans multiple zones, the user may want to force pods to run in a zone specified by a label
* To support targeting nodes with specific network permissions, these node can be specified by labels.

These constraints should be configurable at the job level.

godber:
I propose we add a `cluster` property to the job that contains various job level directives that are related to the cluster manager.  In this particular case we would add `target` under `cluster` to specify details about targeting.

I think in the first implementation, lets just concern ourselves with **required** targets ... and ignore k8s support for preferred targets.

So for now, a job would look like:


```json
{
    "name": "gen-tenk-noop",
    "lifecycle": "persistent",
    "workers": 1,
    "cluster":{
        "target": [
            {"key": "zone", "value": "west"}
        ]},
    "operations": [
        {
            "_op": "elasticsearch_data_generator",
            "json_schema": "",
            "size": 10000,
            "start": "",
            "end": "",
            "format": "",
            "stress_test": false,
            "date_key": "created",
            "set_id": "",
            "id_start_key": ""
        },
        {
            "_op": "noop"
        }
    ],
    "analytics": true
}
```

which would result in the following addition to the jobs worker deployments:

```yaml
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: zone
            operator: In
            values:
            - west
```

If we chose to implement preferred affinities, we can do this later.

```json
{
    "name": "gen-tenk-noop",
    "lifecycle": "persistent",
    "workers": 1,
    "cluster":{
        "target": [
            {"key": "zone", "value": "west"},
            {"key": "rack", "value": "r10", "required": false}
        ]},
    "operations": [
        {
            "_op": "elasticsearch_data_generator",
            "json_schema": "",
            "size": 10000,
            "start": "",
            "end": "",
            "format": "",
            "stress_test": false,
            "date_key": "created",
            "set_id": "",
            "id_start_key": ""
        },
        {
            "_op": "noop"
        }
    ],
    "analytics": true
}
```


```
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: zone
            operator: In
            values:
            - west
[...]

Issue #793, opened 2018-08-21
Add support for job specific resource requests and limits


godber:
K8s supports both resource requests and resource limits.  We should have a way of allowing users to specify these on a per job basis.  We should also probably be making at least some resource request by default in the case where users do not specify them for the job.

godber:
For managing worker resource requests and limits we will continue to use the `cluster` key on the job and extend it with the `resource.request` and `resource.limit` keys.  These will both take `cpu` and `memory` subkeys which map to k8s `vcore` and `memory` resources.  So a job would look like this:

```json
{
    "name": "gen-tenk-noop",
    "lifecycle": "persistent",
    "workers": 1,
    "cluster":{
        "resource": {
            "request": {"cpu": 1, "memory": 2147483648},
            "limit": {"cpu": 2, "memory": 4294967296}
        }},
    "operations": [
        {
            "_op": "elasticsearch_data_generator",
            "json_schema": "",
            "size": 10000,
            "start": "",
            "end": "",
            "format": "",
            "stress_test": false,
            "date_key": "created",
            "set_id": "",
            "id_start_key": ""
        },
        {
            "_op": "noop"
        }
    ],
    "analytics": true
}
```

Which would map to a resource snippet as follows on the worker deployment:

```yaml
    resources:
      requests:
        memory: 2147483648
        cpu: 1
      limits:
        memory: 4294967296
        cpu: 2
```

## Questions

**Q** If a job omits these, does the cluster master insert defaults or do we let workers run without resource settings?

**A** Perhaps if there are cluster defaults defined, then we apply them.  Otherwise there are no defaults.

**Q** What to do if `limits` are provided but `requests` are not?  It's probably OK to have `requests` without limits. 

**Q** What validation should be enforced?  Must `limits` be greater than `requests`?

peterdemartini:
If kubernetes works with a limit but no minimum, we should allow that. 

Yes, I think validations should enforce that limits are greater than the minimum. Additionally there should be a lower validation limit for both values, for example Teraslice requires at `60M` RAM. 

Also it would be easier for user’s to deal with human readable bytes, like `1G` or `2048M`

godber:
The section linked to here describes how scheduling decisions are made based on resources

https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#how-pods-with-resource-requests-are-scheduled

The section following the one above describes how BOTH `requests` and `limits` are applied for both CPU and memory.  Mostly:

[...]

Issue #785, opened 2018-08-16
Add jobName and clusterName labels to pods


godber:
The pods created for a teraslice job should havel labels that contain the `jobName` as well as the `clusterName`.

godber:
Done in #802

Issue #782, opened 2018-08-15
Add support for custom volumes on k8s deployments


godber:
Some environments may want to attach k8s persistent volumes to Teraslice worker pods when they are created.  The deployment should have a way of reading the desired volumes and paths from the teraslice config and adding them to the deployment used to create the workers.

Other options:

* completely override the deployment used
* teraslice jobs could technically define volumes to use

godber:
A volume would be specified by a job as follows

```json
{
    "name": "gen-tenk-noop",
    "lifecycle": "persistent",
    "workers": 1,
    "volumes": [
        {"teraslice-data1": "/data"}
    ],
    "operations": [
        {
            "_op": "elasticsearch_data_generator",
            "json_schema": "",
            "size": 10000,
            "start": "",
            "end": "",
            "format": "",
            "stress_test": false,
            "date_key": "created",
            "set_id": "",
            "id_start_key": ""
        },
        {
            "_op": "noop"
        }
    ],
    "analytics": true
}
```

which would result in a k8s deployment snippet like:

```yaml
      containers
          ...
          volumeMounts:
            - name: teraslice-data1
              mountPath: /data
      ...
      volumes:
        - name: teraslice-data1
          persistentVolumeClaim:
            claimName: teraslice-data1
```

godber:
Volumes should also be configurable at the cluster level, these volumes will be mounted on all pods.  Possibly implemented separately from this issue.

godber:
I had originally suggested the following as the job level config for suggesting mounts:

```
    "volumes": [
        {"teraslice-data1": "/data"}
    ],
```
but I think this reads better
```
    "volumes": [
        {"name": "teraslice-data1", "path": "/data"}
    ],
```

godber:
Done on #802

PR #802, opened 2018-08-24
Issue 794 worker deployment improvements

godber:
This PR includes the following improvements:

* #785 - add jobName and clusterName labels
* #782 - add support for volumes, by adding the following to a job definition

```
    "volumes": [
        {"name": "teraslice-data1", "path": "/data"}
    ],
```

* #793 - add support for resource requests and limits, by adding the following to a job definition

```
    "resources": {
        "minimum": {"cpu": 1, "memory": 2147483648},
        "limit": {"cpu": 2, "memory": 4294967296}
    },
```

* #794 - add support for affinity, by adding the following to a job definition

```
    "node_labels": [
        {"key": "zone", "value": "west"}
    ],
```


godber:
I'll update the docs with the new options ... oh, there is probably a job schema I have to update too ...

godber:
Changed to WIP, it's just occurred to me that I've only tweaked deployment but that won't apply to the execution_controller job.  Thats annoying.

godber:
So, thinking about the execution controller job ... 

* it probably doesn't need volumes
* if it needs resource limits, it probably needs it's own rather than the same as the workers, perhaps I should hardcode a request and ignore limit for now
* affinity it probably needs

kstaken:
Execution controller will absolutely need volumes.

godber:
Well, if it needs two of the three, maybe I should just apply the same resource requirements and just run the execution controller Job object through the same code and call it good.

kstaken:
Yeah to start it's probably fine to just share the same resource requirements.

godber:
Ok, I've manually confirmed that the volume, resource and label restrictions have their basic functionality.  I'll add some details in the k8s readme about the new job options.  I also should probably track down the job schema validation stuff and update that.

godber:
I have added my new schema components but not added tests for them.

godber:
So close ... my resources validation function is not behaving how I expected ..

```
Job submission failed, error: Error validating job: Error: Validation failed for jobConfig: example-data-generator-job-resource - Cannot read property 'minimum' of undefined
```

labels and volumes worked though.

godber:
This is all sorted out now.  The WIP tag is attached only as long as it takes us to finish discussing the proper naming of things.

The readme here has a summary of how the newly added job options are presented to users:

https://github.com/terascope/teraslice/blob/58b59cb6bd40a75e4c532d7a3c1dd4e3c2f19911/docs/k8s-clustering.md#teraslice-job-properties


Outstanding questions:

* The use of `resources.minimum` seems inappropriate (given that it actually gets used as a limit to some degree (weird design choice on their part).  Perhaps `resources.request` is fine even though that's what k8s uses.
* Kimbro was supposed to sleep on `node_labels` which is what I had chosen, these are intended to set affinity ... to be selectors ... `node_labels` was my choice, but maybe `node_selector` is good.  I am too wishy washy on this one.

PR #805, opened 2018-08-27
set a transition of stopping before its stopped resolves #783

### LINKED ISSUE #783: job status set to stopped before all workers exit
When stopping jobs on the k8s qa instance of teraslice (v0.38.1) I found that the job status will be set to stopped before all the workers have exited.  The job was reading from kafka with 50 workers, so stopping the all the workers takes about 30 seconds.

jsnoble:
simple fix. It changes the status to `stopping`, waits for it to stop, then it checks to see if its already in a terminal status before changing it to `stopped`

### DIFF
--- a/packages/teraslice/lib/cluster/services/api.js
+++ b/packages/teraslice/lib/cluster/services/api.js
@@ -141,19 +141,10 @@ module.exports = function module(context, app) {
 
         const handleApiError = handleError(res, logger, 500, `Could not stop execution for job: ${jobId}`);
 
-        jobsService.getLatestExecution(jobId)
-            .then((execution) => {
-                const isTerminal = _isTerminalStatus(execution);
-                if (isTerminal) return res.status(200).json({ status: execution._status });
-                return executionService.stopExecution(execution.ex_id, timeout)
-                    .then(() => {
-                        if (blocking === true || blocking === 'true') {
-                            return _waitForStop(execution.ex_id)
-                                .then(() => res.status(200).json({ status: 'stopped' }));
-                        }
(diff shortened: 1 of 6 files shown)


kstaken:
We had discussed implementing this so that both the native and kubernetes backends would implement stop as a non-blocking operation. Then the API layer would add blocking back in for that use case. 

You had said you were going to drive this based on monitoring cluster state. Did that turn out to not be possible?

Doing it the way it was discussed has two important elements
1. There is only one implementation of the polling logic.
2. It is now possible to implement a non-blocking stop operation in the API using a parameter.

jsnoble:
I thought we all decided that we did not want a non blocking stop as it is inherently dangerous. It would allow jobs to be started before they were shutdown, too many potential errors. As of right now, it is the messaging system and the timeouts that control native clustering because we require a response back from the node_master. After the timeout it will force kill the processes. I opted to put the logic in the k8s part so that calling stopExecution will behave the same in both circumstances instead of cluster blocking and k8's does not block. If we really wanted to allow non blocking stop job capabilities I would have to alter native stop job to not use transactional messaging and put the cluster state monitoring in the execution service layer (which we usually leave that to the cluster service to micro mange things like that) so that it could work with both.

kstaken:
This looks good. Only issue is that calling stop twice on the same job results in an error being logged which is probably unnecessary.

```
Unable to set execution 4e06dd1a-62be-43a9-a27a-ab245e60f80f status code to stopping Error: Cannot update terminal job status of \"stopped\" to \"stopping\"\n
```

PR #819, opened 2018-09-05
Fix teraslice-worker execution-controller stop issue

peterdemartini:


godber:
Note that I have added the installation of the kafka connector/reader/writer to the Dockerfile.

godber:
The disadvantage of leaving this Dockerfile change in, is that it will result in longer Docker build times in all cases, which can be annoying during development.  Maybe we shouldn't include it.  Also it may not achieve what I want.

peterdemartini:
@godber to speed this up, the extra connects modules after `buynan` install so that you can take advantage of the docker image layers. Also, make sure you install it using specific git hash so that it can be cached for installs, otherwise yarn will have to fetch the repo each time, i.e. `yarn add -W terascope/teraslice_kafka_reader#690fbf0`. It might be faster to make them a single `yarn add` command to since you add many modules at once.

godber:
With DEBUG on, it just prints `shutdown is waiting for execution to finish...`

godber:
refs #820

PR #828, opened 2018-09-18
Mock api tests

jsnoble:
updates op test harness. This change should be backwards compatible


packages/teraslice-op-test-harness/index.js
```diff
@@ -17,55 +17,196 @@ const simpleData = [
...
+            instance[mtd] = instance[mtd].bind(instance);
         });
-    });
 }
 
-module.exports = (processor) => {
-    /* A minimal context object */
-    const context = new TestContext('teraslice-op-test-harness');
-    const events = context.apis.foundation.getSystemEvents();
+class Operation {
```
peterdemartini: This file needs to broken out into different files, it usually a bad smell if you have multiple classes per file

packages/teraslice-op-test-harness/index.js
```diff
@@ -17,55 +17,196 @@ const simpleData = [
...
+            executionContext = { config: _.cloneDeep(executionConfig) };
+            this.executionContext = executionContext;
+        }
+        this.retryData = retryData;
+
+        if (newContext) {
+            context = Object.assign({}, _context, newContext);
+            this.context = context;
+        }
+        if (client) {
```
peterdemartini: This check should be smarter; they should only be able to retrieve the client that matches `getConnection` specification. This might change the way a developer specifies a client for the test. It might make sense to have a `->addClient()` method to this class so you can specify additional information and it might make easier to do that instead of having to specify it on each `->init()`

packages/teraslice-op-test-harness/index.js
```diff
@@ -17,55 +17,196 @@ const simpleData = [
...
+    }
+
+    // TODO: Im not sure this should be here, maybe this should all be in init
+    async init({
+        opConfig: newOpConfig = null,
+        executionConfig: newExecutionConfig = null,
+        context: newContext = null,
+        retryData = [],
+        client = null,
+        type = 'slicer'
```
peterdemartini: It would be nice to have convenience methods for this called `->initSlicer`, `->initReader`, and `initProcessor` that infer the type. That has the added benefit of only exposing options available to that type, like the `slicer` is the only one that cares about the `retryData`.

PR #834, opened 2018-09-21
Update Operation APIs and Examples

peterdemartini:
I made some more changes to the new Operation APIs and examples. I am looking for feedback on these changes.

I created examples that used the operation APIs in a backwards compatible way, which means that operations can be developed using this style with the current build. But it does require that a `@terascope/job-components` is required in the asset bundle.

After discussing these changes, the following things need to be completed:

- [x] Make DataEntity factory methods static methods, and remove them from `OperationCore`.
- [x] Make base class for a `Schema` and make the default `convict`.
- [x] Expose a shim for legacy operations from within `job-components`.
- [x] Make OperationAPIs created only when requested, and blow up if called again.
- [x] Expose a "get" OperationAPI method.
- [x] Add a temporary manual registry of `assets` to `apis` which will be used to look up OperationAPIs. This will be a placeholder for shims to support APIs without having to build asset peer dependencies yet.
- [x] Change `makeDataEntity` to `make`


kstaken:
How does the processor get access to the module if it's not included in the asset bundle?

peterdemartini:
I am going to remove WIP from this PR. One thing to note is that I decided to just use a temporary simple mapping of names to APIs, and the APIs are added to the registry when using a processor or reader shim.

Issue #841, opened 2018-09-26
Failed k8s jobs or deployments should result in ex being failed.


godber:
When the cluster master submits the execution_controler k8s job and worker k8s deployment, if there's an error during that process, the cluster master should set the ex to failed, add an error to the ex and then make sure there's no job or deployment left around.

godber:
The workaround for this, until it is resolved, is to view the logs on the master.  Logging in the k8s master works well ... 

```
kubectl -n ts-dev1 logs teraslice-prod1-master-6e41f283d5-vtphr | bunyan
```

godber:
Probably need to do something different when:

* [Creating the slicer](https://github.com/terascope/teraslice/blob/master/packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/index.js#L154-L167)
* [Creating the worker](https://github.com/terascope/teraslice/blob/master/packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/index.js#L202-L208)

Issue #827, opened 2018-09-14
Add cluster name constraint


godber:
There should be constraints on cluster names applied at the config level and added to the documentation, the value is currently only constrained to be a string:

https://github.com/terascope/teraslice/blob/v0.40.2/packages/teraslice/lib/config/schemas/system.js#L68-L72

Cluster names are used in elasticsearch index names so they have to be valid for that usage.

godber:
ES index name restrictions:

https://stackoverflow.com/questions/41585392/what-are-the-rules-for-index-names-in-elastic-search

Also we use cluster name as a k8s label.

PR #842, opened 2018-09-27
K8s improvements 2

godber:
A few more k8s improvements.


packages/job-components/src/formats.ts
```diff
@@ -51,6 +51,37 @@ export const formats : Format[] = [
...
+                || _.startsWith(val, '+')) {
+                throw new Error(`value: ${val} should not start with _, -, or +`);
+            }
+
+            if (val === '.' || val === '..') {
+                throw new Error(`value: ${val} should not equal . or ..`);
+            }
+
+            const badChar = new RegExp('[#*?"<>|/]');
+            if (badChar.test(val) || val.includes('\\')) {
```
godber: Handling `\` as an invalid character was a hassle since it's the escape character.  I have a test for it, which passes, but I still have doubts since I have a bad feel for what contexts `\` works as the escape character.  I expected `RegExp('[#*?"<>|/\\]')` to work, but it did not.

packages/teraslice/lib/cluster/services/execution.js
```diff
@@ -398,13 +398,21 @@ module.exports = function module(context, { clusterMasterServer }) {
...
-                                logger.error(`Workers failed to be allocated, they will be enqueued, error: ${parseError(err)}`);
+                                if (isNative) {
+                                    // this is to catch errors of allocateWorkers
+                                    // if allocation fails, they are enqueued
+                                    logger.error(`Workers failed to be allocated, they will be enqueued, error: ${parseError(err)}`);
+                                } else {
+                                    // if k8s allocateWorkers fails, the execution fails
+                                    logger.error(`Failured to create k8s worker deployment ${execution.ex_id}, error: ${parseError(err)}`);
+                                    const errMetaData = executionMetaData(null, parseError(err));
+                                    return setExecutionStatus(execution.ex_id, 'failed', errMetaData);
```
godber: If both @jsnoble and @peterdemartini can think about the changes here and let me know if they thing the native case will still work, that would be great.  Specifically, I have some doubts about my call to `return` here ... is that what keeps Workers from being queued as the native comment says?  Regardless, in k8s the job ends up in `failed`, the workers and execution_controllers removed and I get an error message on the `ex` like I want.  I **think** it's right.

godber:
The changes fix:

* #841 - failed k8s deployments should set `_failureReason` on `ex`
* #827 - add clusterName constraint - `elasticsearch_Name` format

packages/teraslice/lib/cluster/services/execution.js
jsnoble: So there is a little bit of an issue with the code, maybe not. It seems that you are duplicating the error logic of marking it with error etc etc. If you returned a Promise.reject() it would then go to the outer catch which marks the job as failed with the associated error meta data.  Other than that it seems fine

godber: Does that imply that what it did in the allocateSlicer catch here:

https://github.com/terascope/teraslice/blob/v0.42.0/packages/teraslice/lib/cluster/services/execution.js#L407

which was before I modified it, should also be rejecting a promise rather than returning?

godber: Disregard my previous comment.

Trying `Promise.reject(err)` in the inner `catch` block (for `allocateWorkers`) ... it doesn't seem to be handled correctly.  The job erroneously ends up in the running state, only to exit later when workers fail to connect.  And the error on the `ex` is wrong.  If the worst flaw in this patch is code duplication, then at least it works right.

```
diff --git a/packages/teraslice/lib/cluster/services/execution.js b/packages/teraslice/lib/cluster/services/execution.js
index 41d2129..81aaeac 100644
--- a/packages/teraslice/lib/cluster/services/execution.js
+++ b/packages/teraslice/lib/cluster/services/execution.js
@@ -405,8 +405,7 @@ module.exports = function module(context, { clusterMasterServer }) {
                                 } else {
                                     // if k8s allocateWorkers fails, the execution fails
                                     logger.error(`Failured to create k8s worker deployment ${execution.ex_id}, error: ${parseError(err)}`);
-                                    const errMetaData = executionMetaData(null, parseError(err));
-                                    return setExecutionStatus(execution.ex_id, 'failed', errMetaData);
+                                    Promise.reject(err);
                                 }
                             })))
                     .catch((err) => {
```

packages/job-components/src/formats.ts
peterdemartini: I believe it has to be `RegExp('[#*?"<>|/\\\]')`

packages/teraslice/lib/cluster/services/execution.js
peterdemartini: According to the diff, you added a `Promise.reject(err)` but never returned it, so it won't be propagated up.

I think we should avoid adding more conditional clustering logic to the execution service if possible

godber:
Yeah, I was missing a `return` ... so returning the promise.reject fails the job as desired.

The best solution I have for the `if (isNative)` branch is to just let the job fail indirectly as it does in the native cluster case.  So instead of immediately getting an error on the `ex` saying the k8s worker deployment failed the job will go into `running` state, will never get workers, then fail with this on the `ex`:


```
"_failureReason": "slicer for ex 4fa2d559-032c-4287-9688-ff1725ca7071 had an error, shutting down execution, caused by Error: No workers have connected to slicer in the allotted time: 60000 ms
...
```

This is because `_executionAllocator` suppresses any errors raised by `allocateWorker`

https://github.com/terascope/teraslice/blob/v0.42.0/packages/teraslice/lib/cluster/services/execution.js#L399-L404

I assume this is deliberate.  But I can't "raise any harder" in the k8s cluster master.  So we'll either have to live with the behavior described above, accept the if statement in `_executionAllocator`  or rethink why the native cluster master requires these errors suppressed here.

PR #850, opened 2018-10-04
v0.42.3 execution controller and messaging performance improvements

peterdemartini:


peterdemartini:
I found a memory leak within the elasticsearch client, see https://github.com/elastic/elasticsearch-js/issues/707. I couldn't quickly fix or find the issue; we will have to monitor the severity of the leak, but for now I suggest we move forward with merging these changes.

PR #857, opened 2018-10-16
Changes to Job Components

### LINKED ISSUE #846: Add support for Observers
In the new job APIs, we should support the concept of Observers, which are autoloaded and never called directly from another operation.

In addition to base slice events, an observer should be to observe the data after an operation completes, or fails.

### LINKED ISSUE #849: Add native support the new Job APIs in Teraslice
Currently, the new job APIs are only usable via shims, the new Job APIs should be supported natively in Teraslice. To maintain compatibility we either need to create a shim for legacy operations to operate like new operations, or to provide an API that both backwards and forwards compatible.

The following things are required for this:

- Support the new Slice Events
- Support the new Slicer Events
- Support the different Operation Subclasses and their lifecycles
- Support the new folder structure within when loading an "Operation"
- Support the new validation methods

### LINKED ISSUE #854: job-components: filter returning null
I've been experimenting with a filter based on the new job-components and it looks like it requires the return value to be `null` in order to exclude the record. If you return `false`, which I expected to work, it appears to short circuit the evaluation of the slice or something as no further data is handed to the processor. Is there some particular reason this doesn't just exclude based on a falsey result?

peterdemartini:
- Support loading and validating the new Operations (in a backwards and forwards compatible way)
- Add shims for wrapping old style operations in a new way
- Move the Schema validation onto the Schema class (breaking-change)
- Separate the `Processor` class to `EachProcessor`, `MapProcessor` and `FilterProcessor` (breaking-change)
- Use decorators for  the `DataEntity` to properly lock down the methods on the class
- Use a `WeakMap` for the metadata on DataEntity in-order to privately store metadata, with added benefit or memory improvements
- Add the ability to pass in metadata to the constructor of `DataEntity`
- Add an `Observer` subclass
- Remove `teraslice-types`, since there was no reason for it to exist, everything in `teraslice-types` could have been placed in `job-components`

Resolves #846 
Resolves #854


peterdemartini:
That module just didn't make sense any more, because the teraslice-types was made before the job-components package and which contained mostly job component type definitions. It was becoming a hassle and confusing to deal, which was becoming more and more apparent the more job components stuff was added. In addition to that teraslice-types was a confusing package and pretty much useless outside of the teraslice repo

PR #864, opened 2018-10-18
Add ExecutionContext to Job Components

peterdemartini:
- bump `job-components`
- add `ExecutionContext` to `job-components`
- create two variations of `ExecutionController`, one used by the `slicer` and one by the worker
- reduce the size of `job-components` from 18mb to 13mb by removing `lodash`
- add support for `slicerQueueLength`
- reorganize some of the interfaces and files in `job-components`
- a few other fixes

No breaking changes were introduced in this PR


peterdemartini:
# Notable Changes

Since these changes include design changes, it is essential to
simplify the core changes to get proper feedback and suggestions.

- Thanks, @godber for suggesting this

## Execution Context

Currently, the ExecutionContext looks like this:

```ts
interface ExecutionContext {
    config: ExecutionConfig;

    // this was used on the execution controller
    slicer: Function;

    // this was used on the execution controller
    queueLength: 10000|number;

    // this was used on the execution controller
    dynamicQueueLength: boolean;

    // this was used on the worker
    queue: Function[];

    // this was used on the worker
    reader: Function|null;

    // this was never used
    reporter: null;
}
```

I have separated the ExecutionContext into two classes,
one for the Slicer and one for the Worker.


```ts
interface SlicerExecutionContext {
   // ExecutionConfig a unique configuration instance for a running Job
   readonly config: ExecutionConfig;

   // SlicerContext extends the Terafoundation Context
   // to includes the type definitions for the APIs available to a Execution Controller
   readonly context: SlicerContext;

   // A list of assetIds available to the job.
   // This will be replaced by `resolvedAssets`
   readonly assetIds: string[];

   // The instance of a "Slicer"
   readonly slicer: SlicerCore;

   // Called to initialize all of the registered operations available to the Execution Controller
   async initialize(recoveryData: object[]): void;

   /// Called to clean up all of the registered operations available to the Execution Controller
   async shutdown(): void;

   // Returns a list of any registered Operation that has been
   // initialized.
   getOperations(): Operations[];
}

export interface SlicerContextApis extends ContextApis {
    // Includes an API for getting a client from Terafoundation
    op_runner: OpRunnerAPI;
    // Includes an API for getting a opConfig from the Job
    job_runner: JobRunnerAPI;
}

export interface SlicerContext extends Context {
    apis: SlicerContextApis;
}
```

```ts
interface WorkerExecutionContext {
    // ExecutionConfig a unique configuration instance for a running Job
   readonly config: ExecutionConfig;

   // WorkerContext extends the Terafoundation Context
   // to includes the type definitions for the APIs available to a Worker
   readonly context: WorkerContext;

   // A list of assetIds available to the job.
   // This will be replaced by `resolvedAssets`
   readonly assetIds: string[];

   // The instance of a "Fetcher"
   readonly fetcher: FetcherCore;

   // A Set of a Processors available to Job.
   // This does not include the Fetcher since they have
   // different APIs.
   readonly processors: Set<ProcessorCore>;

   // Called to initialize all of the registered operations available to the Worker
   async initialize(): void;

   // Called to clean up all of the registered operations available to the Worker
   async shutdown(): void;

   // Returns a list of any registered Operation that has been
   // initialized.
   getOperations(): Operations[];
}

export interface WorkerContextApis extends ContextApis {
    // Includes an API for getting a client from Terafoundation
    op_runner: OpRunnerAPI;
    // Includes an API for getting a opConfig from the Job
    job_runner: JobRunnerAPI;
    // An API for registering and loading the new Job APIs
    executionContext: ExecutionContextAPIs;
}

export interface WorkerContext extends Context {
    apis: WorkerContextApis;
}
```

## Slicer Queue Length

**NOTE:** This functionality has not previously been added to the new Job APIs.

Currently, the only to specify the max number of available slices that can be queued is by defining a `slicerQueueLength` function on the "Reader". If this method was not specified the default was set to `10000`. If this method returned `QUEUE_MINIMUM_SIZE` the queue length was
limited to the number of workers connected. If the method returned a `number` greater than `0` it would use that.

This has a couple of drawbacks:

- This is only called during initialization
    - For example, the Slicer could never dynamically
      adjust the minimum queue size based its data source or internal
logic.
- The `QUEUE_MINIMUM_SIZE` is pretty limited
    - For example, the Slicer cannot set it to the number of workers times 2,
      which would work well in the kafka connector senario.

To remedy this, I have added three new methods the "Slicer":

```ts
// Simplified interface
class Slicer {
    // Used to determine the maximum number of slices queued.
    maxQueueLength() {
        // example for current functionallity
        if (legacyQueueLength === 'QUEUE_MINIMUM_SIZE') {
            return this.workersConnected;
        }
        // example for kafka
        if (kafka) {
            return this.workersConnected * 2;
        }
        // actual implementation defaults to 10000
        return 10000;
    }

    // In order to give the Slicer knowledge about
    // the number of workers connected
    // and other information (see below).
    // This is the default implementation
    // and does not need to specified
    onExecutionStats(stats: ExecutionStats) {
        return this.stats;
    }

    // This is the default implementation
    // and does not need to specified
    protected get workersConnected() {
        return this.stats.workers.connected;
    }
}

interface ExecutionStats {
    workers: ExecutionWorkerStats;
    slices: ExecutionSliceStats;
}

interface ExecutionWorkerStats {
    connected: number;
    available: number;
}

interface ExecutionSliceStats {
    processed: number;
    failed: number;
}
```

peterdemartini:
I got bored waiting for a review and created some benchmarks:

The benchmark suite basically contains tests and round using a simple set of test cases
Every test will create using the Fetcher 10000 records and then run it through the following operations Map -> ForEach -> Filter -> ForEach.

The variation of between test cases is to test the performance impact of DataEntities and using the Operation specific methods or the generic Teraslice handle method. The handle method for each operation is a basic generic interface for Teraslice to interact with, and each Operation has to implement they own version of it in order to handle the different APIs they expose.

```txt
---------------------------- Simple Job ----------------------------
calling methods without DataEntities   74.58 op/s ±  7.69%   (69 samples)
calling methods with DataEntities   36.35 op/s ±  3.55%   (56 samples)
calling handle with precreated DataEntities   33.80 op/s ±  4.34%   (53 samples)
calling handle with automatic DataEntities   36.91 op/s ±  2.87%   (56 samples)
------------ Best: calling methods without DataEntities ------------
```

After running the benchmarks I found and made one improvement, see https://github.com/terascope/teraslice/pull/864/commits/77a8fb262931b37d055556c1398bf19d2743ff3b

```txt
---------------------------- Simple Job (after) ----------------------------
calling methods without DataEntities   82.36 op/s ±  4.76%   (69 samples)
calling methods with DataEntities   41.30 op/s ±  4.34%   (62 samples)
calling handle with precreated DataEntities   39.57 op/s ±  3.55%   (60 samples)
calling handle with automatic DataEntities   42.94 op/s ±  2.94%   (63 samples)
------------ Best: calling methods without DataEntities ------------
```

PR #881, opened 2018-11-01
added minimal support for new procssor type operations

jsnoble:


kstaken:
Is there updated documentation on how to use this now?

PR #882, opened 2018-11-02
Improvements for teraslice-test-op-harness and parallel-slicer

peterdemartini:
* Easier use in tests
* Fix parallel slicer to handle slicers race to finish
* Return the new Constructors on the legacy shims to make it more compatible with test-op-harness


peterdemartini:
@jsnoble yes, but it is `{ fullSlice: true }`. I will add a documentation for it.

Issue #877, opened 2018-10-30
ex/errors endpoint


jeffmontagna:
Add an `/ex/errors` endpoint to the api.  Currently the only way to get the errors for an ex_id is to hit the jobs endpoint,  i.e. `GET jobs/{job_id}/errors/{ex_id}`.  Functionality should be just like `/jobs/{job_id}/errors` or `/jobs/{job_id}/errors/{ex_id}`

Issue #874, opened 2018-10-29
teraslice-cli issues


kstaken:
Getting these into a ticket so that we can review. Some of these should probably be broken out into separate tickets.

--cluster-manager-type -- this option can be removed as this information is now exposed on the root URL of Teraslice and can be introspected.

Options on sub commands -- when you get the options on sub commands many of them do not apply to the current context and should not be displayed.

Examples
     * some examples still hard code the command name
     * some commands are missing examples

Not enough non-option arguments: got 0, need at least 1 
    * this message is pretty unclear on what it wants. Especially on the commands where examples are missing.

cluster_sh -- There is confusion over cluster_sh and object_ids. In most cases where the commands reference cluster_sh they really mean object_id.

earl ex list -- should not be restricted to running/failing by default, nor should any other command. Filtering by status should be an option and if there is a common set of statuses then maybe special flags can be added for those cases.

earl jobs run -- the mechanism for specifying the cluster when providing the jobs file is not correct
     * You can do `earl jobs run localhost:job.json`
     * But then you can not do `earl jobs stop localhost:job.json` which is completely inconsistent
     * I think it's a bad idea to muddle the cluster_sh, object_id with the job.json handling. 
     * This implies that when you're working with a job.json then you have to specify the cluster independently. We should discuss.

earl jobs run -- Doesn't add metadata to the job. Open for discussion if it should. tjm did.

earl jobs stop -- should not prompt for confirmation when stopping a single job.
     
earl jobs --
     - Commands that take .json files should not require the file extension
     - The documentation is still written from the perspective of 'all' jobs

earl jobs find -- no idea how to use this command. I question if it should exist until we have a stable base.

earl assets list -- this command doesn't exist but should

earl assets deploy -- does not appear to add the meta data to assets.json

earl aliases update -- it appears re-running add has the same behavior as update. 

earl ex errors localhost --
```
UncaughtException:  TypeError: tsuFunctions.errors is not a function
    at Object.exports.handler (/Users/kstaken/projects/opensource/teraslice/packages/teraslice-cli/cmds/ex/errors.js:37:25)
    at Object.runCommand (/Users/kstaken/projects/opensource/teraslice/node_modules/yargs/lib/command.js:238:44)
    at Object.parseArgs [as _parseArgs] (/Users/kstaken/projects/opensource/teraslice/node_modules/yargs/yargs.js:1059:30)
[...]

PR #888, opened 2018-11-07
Fixed serveral bugs from #874

jeffmontagna:
- cluster_sh shouldn't allow files only object_ids.  
    - Removed option to have files in cluster_sh.  
    - Added more error checking for invalid entries 
- `assets` 
    - `list` -- this command doesn't exist but should.  Added command.
    - `deploy` -- does not appear to add the meta data to assets.json.  Updated to now update asset.json
- `ex errors`
    - needs ex/errors endpoint #877
    - added function, output placeholder for now
- `jobs` 
    - Commands that take .json files should not require the file extension. Removed `.json` requirement.
    - `init` command not implemented.  Added command
    - `restart cluster --all` error message when stopping prevents starting of jobs
    - `start` error after starting job, `Cannot read property 'workers_active' of undefined`
    -  Fixed issue with restart in which the stop loop would exit prematurely and try to start up jobs.  Replaced `for` loops with a map which resulted in faster job restarting because and removed the premature exit.


kstaken:
In the examples there is this which doesn't work. Not sure if it is supposed to.

```
earl jobs list http://cluster1.net:80
```

peterdemartini:
I recommend you rebase or merge master in

PR #892, opened 2018-11-13
Fix loading of compiled TypeScript operations and start job-test-harness

peterdemartini:
Running `export default class Example {}` will compile to this `exports.default = Example`, which causes our operation loader to fail. There are three options to fix this, figure out how to fix typescript compilation to make it work right, use `export = class Example {}` or add support for loading `exports.default` into the `OperationLoader`. I chose the latter because the first two require a specific configuration of Typescript which will push the problem to the users and other our projects.

Additionally, this PR introduces a job-test-harness package which currently does very little but it is a start. For the job-test-harness, I needed to a few test operations in the job-components since they will be needed for running a job in "tests". So added a `builtin` directory to job-components with a few test operations like, `test-reader`, `noop` and `delay`.


jsnoble:
Peter and I discussed that in the future the job and op test harness should be together in a top level non-scoped library so it can share more

PR #895, opened 2018-11-15
Faster DataEntity creation

peterdemartini:
Creating large DataEntities was slow because of using shallow cloning the object onto the DataEntity. I switched DataEntity.make to create a [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) instead of using Object.assign. Benchmarks prove that this is roughly 600x-1000x faster, which is very significant and likely well worth another performance impacts the previous DataEntity had.

**NOTE:** Unfortunately I have been unable to figure out how to make `new DataEntity` to work with a Proxy, but this has a very small impact since it is rarely used. I added comments to make this clear and will defer the change to a later date.

Ref https://github.com/terascope/kafka-assets/issues/11

### DIFF
--- a/packages/job-components/bench/data-encoding-suite.js
+++ b/packages/job-components/bench/data-encoding-suite.js
@@ -27,15 +27,13 @@ const run = async () => Suite('DataEncoding')
     })
     .run({
         async: true,
-        initCount: 2,
-        maxTime: 5,
+        initCount: 1,
+        maxTime: 3,
     });
 
 if (require.main === module) {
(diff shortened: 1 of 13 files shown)


peterdemartini:
Removing WIP from the PR

kstaken:
Is this something we still want to move forward with?

peterdemartini:
I don’t see why not, it is faster with very little downsides.

Issue #907, opened 2018-11-27
teraslice-cli simplify lib/config 


jeffmontagna:
The config object used for teraslice-cli is overly complex.  The conditionals for the different commands should removed to simplify the config.  The config should be made more universal regardless on which command is used.

godber:
That is, the set of properties on the `cliConfig` object should be the same for all commands instead of changing for each command.  The values set will be `null` in some cases, but the properties should still be there.

godber:
Now tracking this branch for this issue: https://github.com/terascope/teraslice/pull/913

kstaken:
Is this still an issue?

godber:
Done

PR #911, opened 2018-11-29
Teraslice cli simplify lib config

jeffmontagna:
We've made a number of changes

* Introduced a new configuration object (`TerasliceCliConfig`) intended to replace the old objects
* Added tests for `TerasliceCliConfig`
* Changed to a layered approach for including and sharing command line options
* Reworked the `aliases` command to use this new structure

Future pull requests will update the remaining commands.  We are not updating the version yet to avoid building and releasing a new package on npm.

refs #907


godber:
I have manually tested all of the alias commands.  And verified that all of the command/action combinations run without exploding.  I have lengthy output I could share, ask me if you want it.

PR #912, opened 2018-11-30
added test to load all cmds

jeffmontagna:
Added some tests to check that all cli commands can be loaded.


godber:
Note, I updated this branch too.

godber:
I am just going to kill this PR.  We've got some tests for all commands at this point.  Strangely, the tests did not pass ... but TravisCI didn't seem to recognize the error.

PR #920, opened 2018-12-07
Improved e2e tests

peterdemartini:
While testing the kafka asset bundle in the e2e tests, I got annoyed waiting and debugging issues. So this PR includes a few improvements to make things faster and improve the debugging of e2e tests.

- Remove the 10000 record example index to go with less-is-more
- Disable a test in worker-allocation that is unstable and slow
- Remove support autoload cache in-favor of pruning the older assets to make sure only one exists
- Make `spawnAssetLoader` quicky check if the assets exist before trying to spawn
- Improve worker-shutdown logging to make it more clear what is happening and how long it is taking
- Fix scheduler logging (minor)
- Remove dist before building to prevent old files hanging around
- Upgrade TypeScript
- Fix logging in `node_master`

P.S. While waiting for the e2e tests run locally I started to convert the Promise madness in the tests to use async await.


kstaken:
Showing conflicts.

PR #923, opened 2018-12-10
Fix test-op-harness and test client usage

peterdemartini:
When running in elasticsearch-assets there was an issue when changing an existing client between test runs. This is fixed now.


peterdemartini:
The existing documentation is still there

PR #924, opened 2018-12-10
v0.44.4 fix issue docker image useage with kafka

peterdemartini:


peterdemartini:
@godber I added a Dockerfile to e2e that inherited from `terascope/teraslice:latest` and didn't cache as much stuff to speed up the docker builds. Which caused the e2e tests to fail in travis but not locally, I removed it in order to prevent this type of problem from happening again.

Issue #904, opened 2018-11-26
Observers should be top level on the Job


peterdemartini:
The job definition should be able to reference Observers and possibly Operation API so they can be loaded and validated automatically.

Issue #902, opened 2018-11-26
Support for testing out-of-band Observers and APIs in teraslice-test-harness


peterdemartini:
The teraslice-test-harness should support testing Observers and APIs from modules out of band

Issue #450, opened 2017-07-07
Message/record level error handling? - dead letter channel


erik-stephens:
Looks like error handling is done only at the slice level.  I'm looking for some sort of [error channel or dead letter channel](http://www.enterpriseintegrationpatterns.com/patterns/messaging/DeadLetterChannel.html) to write exceptional msgs/records to.  I don't want the entire slice to fail due to a small fraction of failures.  Use cases I'm thinking of:

1. Garbage in.  Retrying the slice won't help.  Gracefully redirecting the garbage somewhere will better allow clients gauge how impactful it is and what to do about it.

1. Transient external exceptions (eg network, disk, or external service outage).  I wouldn't want to hold up or retry an entire slice if only a small fraction of items failed to process successfully.  Sending them to another channel for retry later would be ideal for large slices.

1. Job bugs - similar to the use case above.  Retrying the slice when there is an app/job level bug will be useless.  Better to capture the unhandled cases somewhere until the job is patched to handle them.

jsnoble:
1) The best way might be to have a schema outside teraslice for the data, so malformed data never hits the pipeline in the first place. We need to set up a contract of what it looks like especially since we need mappings for them to do any sort of queries on it. If not then your reader/slicer would have to check it yourself and determine what to return.
2) If you check out how we have our elasticsearch sender code, and the elasticsearch api code. We have to handle that ourselves. The sender bulk send at times sends up partial bulk, we have to check what wasnt sent and use exponential backoff to send up the rest
3) If this happens everything fails, or the failures shows up in the logs.

What this really means is that the common core readers/processors  that we provide have to be tested and battle hardened. The docs need to explain various circumstances an op needs to account for but its up to the people making their own custom op to make it bullet proof. This is at least of things stands now. Im not fully sure what we could do to help ease others code choices. We have the capabilities for them to implement a recover abilitiy to try to redo any failed slices but they have to make it so their slicer can handle that as well.

kstaken:
We should definitely incorporate an API for this as part of the improved processor APIs and streaming.

kstaken:
This will be implemented in an asset bundle but I'm going to hold this ticket open until we know where it will live.

PR #925, opened 2018-12-11
Better Job / Operation Error Handling

peterdemartini:
This PR adds support for a basic Dead Letter Queue API and when a `onSliceRetry` event fails, the processor can throw an error to stop the retry from happening.

This action will specify what to do when failing to parse or transform a record. ​​​
​​​​​The following built-in actions are supported: ​​​
​​​​​  - "throw": throw the original error ​​​​​
​​​​​  - "log": log the error and the data ​​​​​
​​​​​  - "none": (default) skip the error entirely

​​​​​If none of the actions are specified it will try and use a registered Dead Letter Queue API under that name. The API must be already be created by an operation before it can used.​

Using the dead letter queue:

Option 1:

```ts
class CustomProcessor extends BatchProcessor {
    async onBatch(batch: DataEntity[]) {
        return batch.map((data) => {
            try {
                // do transformation here
            } catch(err) {
                this.rejectRecord(data, err);
            }
        })
    }
}
```

Option 2:

```ts
class CustomProcessor extends BatchProcessor {
    async onBatch(batch: DataEntity[]) {
        return batch.map(this._tryRecord(() => {
            // do transformation here
        }));
    }
}
```

To create and use a custom Dead Letter Queue API:

```js
{
    "name": "Example Job",
    "operations": [
        {
            "_op": "example",
            "_dead_letter_action": "example",
            "_encoding": "json"
        },
        {
            "_op": "noop"
        }
    ]
}
```

```ts
class ExampleDeadLetterAPI extends OperationAPI {
    async createAPI() {
        return (record: any, err: Error) => {
            // handle storing the record for later usage
        };
    }
}
```

```ts
class ExampleFetcher extends Fetcher {
    async initialize() {
        await this.createAPI('example', { ...config });
        return super.initialize();
    }

    fetch() {
        return [
            'hello'
        ].map(this._tryRecord((data) => DataEntity.fromBuffer(data, this.opConfig)))
    }
}
```

Resolves: #904
Resolves: #902 
Resolves: #450

### DIFF
--- a/docs/ops-reference.md
+++ b/docs/ops-reference.md
@@ -59,7 +59,6 @@ Note that the job configuration is divided into top level job configuration, and
 
 #### Job level configuration options ####
 
-
 | Configuration | Description | Type |  Notes |
 | --------- | -------- | ------ | ------ |
 | name | Name for the given job | String | optional |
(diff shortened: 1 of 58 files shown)


peterdemartini:
@kstaken This is what I was thinking about for the specification of APIs. When the Execution Context is initialized, it will register the APIs to be loaded like the other processors. For an operation to use the Kafka dead letter queue, it will specify `"_dead_letter_action": "kafka_dead_letter"`. 

```js
{
    // An array of APIs to load and any configurations they require. Constructed similarly to operations.
    "apis": [
        {
            // required like `_op`
            "_name": "kafka_dead_letter",
            // works like `_op` and is validated when the job submitted
            "connection": "default"
            // topic to dump data to
            "topic": "example-topic",
            // ...           
            // and other kafka producer configuration
        }
    ]
}
```

It this should work the same for observers, which we currently don't support well.

kstaken:
OK, this seems promising.

peterdemartini:
I am removing WIP from this PR as I believe this should work well

Issue #941, opened 2019-01-04
Stuck Workers: Failure to update failed state


godber:
There have been a number of times when teraslice jobs end up with stuck workers (workers with continuously increasing lag, never making progress).  These cases often show errors in the worker logs like this:

```
23:35:45.208Z  INFO worker: Resolving with 100000 results (module=operation, opName=teraslice_kafka_reader, jobName=myJob, jobId=7cd07716-0ab6-4a7a-8089-9818bb7c00bf, exId=a104af7d-d274-4952-abbd-d66892e35693)
23:37:53.850Z ERROR worker: (module=elasticsearch_backend)
    invoking elasticsearch-api _runRequest resulted in a runtime error: Error: Request Timeout after 120000ms
        at /app/source/node_modules/elasticsearch/src/lib/transport.js:354:15
        at Timeout.<anonymous> (/app/source/node_modules/elasticsearch/src/lib/transport.js:383:7)
        at ontimeout (timers.js:498:11)
        at tryOnTimeout (timers.js:323:5)
        at Timer.listOnTimeout (timers.js:290:5)
23:39:53.853Z ERROR worker: (module=elasticsearch_backend)
    invoking elasticsearch-api _runRequest resulted in a runtime error: Error: Request Timeout after 120000ms
        at /app/source/node_modules/elasticsearch/src/lib/transport.js:354:15
        at Timeout.<anonymous> (/app/source/node_modules/elasticsearch/src/lib/transport.js:383:7)
        at ontimeout (timers.js:498:11)
        at tryOnTimeout (timers.js:323:5)
        at Timer.listOnTimeout (timers.js:290:5)
23:39:53.854Z  INFO worker: analytics for slice: slice_id : 4c464058-5b5a-4ecc-9682-20c262f350b2 slicer_id : 0 slicer_order : 5372454 request : 1 _created : 2019-01-03T23:35:12.357Z time : 11659,967,15,2758,4892 memory : 39029928,105036920,3184784,100464152,9553160 size : 100000,100000,86738,133618,133618  (ex_id=a104af7d-d274-4952-abbd-d66892e35693, job_id=7cd07716-0ab6-4a7a-8089-9818bb7c00bf, module=slice, assignment=worker, slice_id=4c464058-5b5a-4ecc-9682-20c262f350b2)
[...]

PR #943, opened 2019-01-04
Fixes for better error handling and prevention for stuck workers

peterdemartini:
This PR includes the following changes:

- Fix Execution Watch Dog in Execution Controller
- Add the ability for a slice retry to "fail loudly" by setting `error.fatalError = true` when `onSliceRetry` is called.
- Avoid marking a slice as "failed" when the state update to "complete" fails
- Avoid the worker's availability getting out-of-sync with the Execution Controller by resending the availability message every 60 seconds when no slices have received on a worker.
- Fix `worker_id` in logger
- Fix bunyan piping error but it still exits early
- Fix stop issue when shutting execution controller
- Updates packages

This PR should help #941 from getting in a stuck state but it won't solve the underling problem


peterdemartini:
I am going to remove WIP from this PR as it has important fixes that I found while debugging the stuck worker's problem. My current theory is the stuck workers have more to do with Kafka that this ES errors.

@godber would you mind reviewing this?

Issue #930, opened 2018-12-13
ElasticSearch Index API/Manager


peterdemartini:
**Requirements:**

- It should have an API abstraction, similar to the `@terascope/elasticsearch-api` package.
- It should expose APIs to replace the majority of the elasticsearch backend in Teraslice.
- It should have an index mapping management API, that can do the following:
    * Detect a mapping change and determine if it can do a seamless update.
    * Provide an API for doing the seamless upgrade.
    * Provide an API for migrating an index (this will likely not be done automatically).

**Misc Goals:**

- It should be well-tested with at least 90% test coverage.
- It should use TypeScript.
- It should use Classes.
- It should be battle tested.
- It should reside in its own package with the teraslice repo.

**Future Goals:**

- Add CLI support for in doing an index migration.

kstaken:
Versioning is the critical aspect of this and I think there a couple ways to think about it.

1. Index based versioning
2. Record based versioning

So far all our data engineering processes have been based around index based versioning and given the restrictions of ES it self that may still be the best approach but I'd like to at least give some thought to record based versioning. That may not help in regard to index mappings but it could be useful from a code compatibility perspective.

kstaken:
This is covered by ElasticsearchStore.

PR #949, opened 2019-01-09
Add new ElasticSearch Store

peterdemartini:
Add new ElasticSearch Store, see #930 

Remaining things to implement for the first release, in no particular order:

 - [x] Add support for Mapping Templates
 - [x] Add support for TimeSeries Indexes
 - [x] Add Index version detection and auto-migration or failure (a very basic implementation)
 - [x] Handle the case where the Index Shards haven't been allocated
 - [x] Handle the case where the Index is not available
 - [x] Build a resilient retry system
 - [x] Use DataEntities for metadata
 - [x] Add support for Data Schema's using JSON Schema
-  [x] Handle Strict Mode for Data Schema and Index


peterdemartini:
Removing WIP from this PR as I think it is good for an initial version

PR #953, opened 2019-01-15
v0.48.0 Update Terafoundation / WebHDFS

peterdemartini:


peterdemartini:
@godber removing rc from the version

PR #977, opened 2019-01-18
refactoring of tjm commands

ciorg:
* created a job-src class
* refactored all tjm commands and added a few more
* added tests for job-src class
* revamped README


packages/teraslice-cli/lib/job-src.js
```diff
@@ -0,0 +1,49 @@
...
+    addMetaData(id, clusterUrl) {
+        // currently a job is only associated with one cluster at a time
+        // but could change in the future
+        _.set(this.content, '__metadata.cli.cluster', clusterUrl);
+        _.set(this.content, '__metadata.cli.version', this.version);
+        _.set(this.content, '__metadata.cli.job_id', id);
+        _.set(this.content, '__metadata.cli.updated', new Date().toISOString());
+    }
+
+    metaDataCheck() {
```
godber: You could implement this as a getter named `hasMetaData` ... `get hasMetaData(...` and the code that uses this would read:

```javascript
if (job.hasMetaData) {
         const regCluster = _.get(job.content, '__metadata.cli.cluster');
         reply.fatal(`job has already been registered on ${regCluster}`);
     }
```

packages/teraslice-cli/lib/job-src.js
```diff
@@ -0,0 +1,49 @@
...
+     * @param {string} name name of job file.
+     */
+    constructor(srcPath, name) {
+        this.jobPath = path.join(srcPath, name);
+        if (!fs.pathExistsSync(this.jobPath)) {
+            reply.fatal(`Cannot find ${this.jobPath}, check your path and file name and try again`);
+        }
+        this.content = require(this.jobPath);
+        // Job file must contain name, number of workers and at least 2 operations
+        if (!(
```
godber: This validation code could be refactored out of the constructor and just called here.

godber: Also, ideally this validation would come from somewhere else in teraslice ... somewhere shared ... for instance whatever code validates the jobs when submitted ... if factored properly could be used here.  Odds are that code is not conveniently re-usable.  Though maybe this validation should differ ... for reasons I don't fully grasp.

godber:
One thing I learned in reviewing this ... in the past I've avoided using:

```javascript
path.join(getDir(), 'foo/baz/bar');
```

because I thought including path separators in your strings wouldn't work as expected on Windows.  But I manually validated that this time, and it does actually search strings and replace the `/` with `\\` on Windows.

packages/teraslice-cli/lib/job-src.js
ciorg: this seems like a good idea

packages/teraslice-cli/lib/job-src.js
ciorg: yeah, it should be reusable from somewhere...I have no idea where though.  If I need this again I think it would be a good idea to pull it out.

peterdemartini: I recommend not using `require` to load a json. It can have odd side effects, like mutations are applied globally and it can require JavaScript files which would work on reads but not writes.

godber:
Note, I clicked "Update Branch" on this.  If you have to come back to this branch after this point, make sure you pull down changes.  Once the tests have passed though, this can be merged as far as I am concerned.

PR #979, opened 2019-01-22
Output cli validation fixes

jsnoble:
- able to use regex as /d.*ta/ as well as  d.*ta
- can specify `output: false` to remove intermediate fields
-  can specify `output: false` on validations to act as a NOT
-  fixed cli -D issues
- migrated required_extractions to output_phase
- added uppercase and lowercase normalizations
- added jsonparse transform


kstaken:
Does giving elasticsearch output to -d require another option or something? It still doesn't appear to work as expected. Just generates empty results for a rule set that works if the file is converted to ldjson.

kstaken:
When I have a sequence of rules similar to the following I get an error.

```
{ "source_field": "somefield", "start": "value=", "end": "EOP", "target_field": "hashoutput", "tag": "hash_field", "post_process": "base64decode" }

{ "follow": "hash_field", "post_process": "urldecode", "tag": "urldecoded" }
{ "follow": "urldecoded", "post_process": "jsonparse", "tag": "parsed" }
```

I get the following error.

```
Error: could not instantiate phase manager: could not find target_field for JsonParse validation or it is improperly formatted, config: {"follow":"urldecoded","post_process":"jsonparse","tag":"parsed"}
```

To get it to work I have to add `target_field` and `source_field` to the rule. The problem is that the `urldecoded` rule right before the rule it's complaining about has neither of those fields and works fine. This is very inconsistent and unexpected.

Also the message refers to "JsonParse validation" but this is a post_processor.

packages/ts-transforms/bin/ts-transform.js
```diff
@@ -2,4 +2,10 @@
 
 'use strict';
 
-require('../dist/command');
+try {
+    require('../dist/command.js');
+} catch (err) {
+    // eslint-disable-next-line
+    console.log('error while attempting to invoke cli command', err.toString());
```
peterdemartini: This should use `console.error`

packages/ts-transforms/src/command.ts
```diff
@@ -1,17 +1,19 @@
...
 import { DataEntity, debugLogger } from '@terascope/utils';
 import _ from 'lodash';
 import got from 'got';
 import { PhaseManager } from './index';
-import { WatcherConfig } from './interfaces';
+import { PhaseConfig } from './interfaces';
+import validator from 'validator';
 
 const logger = debugLogger('ts-transform-cli');
 const dir = process.cwd();
```
peterdemartini: This should be `__dirname` not `process.cwd()` since that would change or break depending on your current directory.

peterdemartini: Oops, I was referring to the package.json version code below.

PR #982, opened 2019-01-25
v0.49.0-rc1 Improved Error Handling in Teraslice and ElasticSearch

### LINKED ISSUE #986: Add additional context to connector related error messages
At a minimum it would be good to have a clear indication of the connection that the message applies to. Ideally as a new attribute on the log entry.

### LINKED ISSUE #987: Analytics background bulk requests can crash workers
Analytics storage can throw an unhandled exception when the bulk request times out. This can result in the worker crashing. Analytics storage should never adversely impact the worker.

 Default bulk request retry policy should be applied. We don't want to lose analytics if at all possible but if the retry cycle fails then we should just log a clear indication and continue.

peterdemartini:
Improvements ElasticSearch state error handling to ensure all slice events are called properly.

Resolves #986 
Resolves #987 
Ref #941 - this should fix the stuck worker problem, but it might be too hard to tell
Ref #989 - only solves it for the Worker

**NOTE** This updates some dev dependencies


packages/utils/src/logger.ts
```diff
@@ -58,14 +59,22 @@ export function debugLogger(testName: string, param?: debugParam, otherName?: st
...
     ];
 
     for (const level of levels) {
         const fLevel = `[${level.toUpperCase()}]`;
         const debug = debugFn(name);
+
         logger[level] = (...args: any[]) => {
+            if (level === 'trace' && !logTrace) return false;
+            if (level === 'fatal') {
+                console.error(name, ...args);
```
erik-stephens: Does this mean we'll see un-bunyan'd log output in production?

packages/teraslice/test/workers/worker/slice-spec.js
```diff
@@ -320,9 +320,9 @@ describe('Slice', () => {
                 await testContext.cleanup();
             });
 
-            it('should throw an error if given invalid state', async () => {
+            it('should throw an error if given invalid state', () => {
                 const data = { should: 'break' };
-                return expect(slice._logAnalytics(data)).rejects.toThrowError(/Failure to update analytics/);
+                return slice._logAnalytics(data);
```
erik-stephens: This change seems at odds with the test description.

peterdemartini: Ah yes the test description needs to be changed since it won’t throw and error any more.

packages/utils/src/logger.ts
peterdemartini: No, in most cases, this is used as replacement for a bunyan logger within our tests.

PR #991, opened 2019-01-31
fix index.js in teraslice-cli package.json

godber:
We fat fingered something here that breaks package install.


godber:
I just pushed this through without review.

PR #992, opened 2019-02-01
Typescript Improvements

peterdemartini:
TypeScript compilation was getting slow since we are adding more projects and the configuration was becoming more difficult to manage. This PR simplifies the configuration and reduces the build times.


package.json
```diff
@@ -6,9 +6,9 @@
         "setup": "yarn bootstrap && yarn build",
         "bootstrap:prod": "lerna bootstrap -- --production=true --link-duplicates --ignore-optional",
         "bootstrap": "lerna bootstrap --force-local",
-        "build": "lerna run build",
-        "build:prod": "lerna run build:prod",
-        "build:watch": "lerna run build:watch --stream --prefix",
+        "clean": "tsc --build --clean",
+        "build": "lerna run build:prepare && tsc --build --pretty",
```
jsnoble: where is build:prepare? i dont see it in the scripts

jsnoble: we talked and resolved this quesiton

Issue #918, opened 2018-12-04
Consider renaming k8s deployments with teraslice job names


godber:
The teraslice execution controller and workers all have the execution ID in their names.  It might be better to put the job name in the job and deployment names instead.  The labels would still contain jobId and exId.  Though this might have negative consequences.  I haven't fully thought through it.

erik-stephens:
Might want to abbreviate "teraslice-worker-..." to something like "ts-wkr-".  I thought we were limited to 64 characters, but looks like [up to 253 is supported](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/), so maybe not that important.

Also, if going to bake an id from the job, I think job_id makes more sense than ex_id since we almost always operate against job ids and there shouldn't be two instances of the same job running.

erik-stephens:
Out of scope for this issue but wanted to point out another point where it would be useful to bake worker details into the name: kafka consumer client ids.  Being able to look at something like kafka manager and quickly see which workers are lagging would be useful.

godber:
Seems for some reason 63 is still the max for deployment names ... when I try to exceed that I get:

```
Error: Deployment.apps \"ts-wkr-example-data-generator-job-d7fe2074-d0c6-4168-9300-065a1c04d944\" is invalid: spec.template.spec.containers[0].name: Invalid value: \"ts-wkr-example-data-generator-job-d7fe2074-d0c6-4168-9300-065a1c04d944\": must be no more than 63 characters\n    at K8s.post
```

godber:
@erik-stephens how do you feel about:

```
ts-exc-example-data-generator-job-d9e36896-f0b9-p2hdc             1/1     Running   0          20s
ts-wkr-example-data-generator-job-d9e36896-f0b9-bb6d4cc66-dcdd2   1/1     Running   0          19s
ts-wkr-example-data-generator-job-d9e36896-f0b9-bb6d4cc66-wchl8   1/1     Running   0          19s 
```

That's this template:

```
        const name = `ts-exc-${jobNameLabel.substring(0, 43)}-${execution.job_id.substring(0, 13)}`;
```

erik-stephens:
GET OUT OF MY HEAD! /s  That's exactly what I was hoping for in the face of that limitation.

godber:
I include more of the job name and less of the job Id on the assumption that the job name is more likely to collide.

Issue #914, opened 2018-12-03
Teraslice in k8s should set default memory and CPU limits


godber:
The Dockerfile currently enforces a 2GB node heap limit:

https://github.com/terascope/teraslice/blob/v0.44.2/Dockerfile#L39

Though k8s does not have any limits set by default.  Since the 2GB is effectively an implicit limit, we should just accept that fact and make k8s also set memory and cpu limits by default.

godber:
@kstaken you want the default node heap setting to be about 4GB right?

In recent release, it's 2GB.

kstaken:
Since we can change it easily a smaller default is fine. We only set it that large before because it was hard to change.

godber:
Ok, so the plan for this is as follows.

* I am adding `cpu` and `memory` to cluster config.  These will be optional but recommended.
* These can be overridden on a job by job basis (already implemented).
* Internally we will decide what are appropriate cluster wide values.

Keep in mind that both the k8s request and limit are set to the same value and that node's `max-old-space-size` is set to `0.9 * memory` and if `memory` is unset it is `2048`.

PR #998, opened 2019-02-05
K8s improvements feb 2019

godber:
Making some k8s improvements.  Definitely a work in progress at the moment.

* Updated for new earl commands


packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/index.js
```diff
@@ -172,35 +171,15 @@ module.exports = function kubernetesClusterBackend(context, clusterMasterServer)
      * Creates k8s deployment that executes Teraslice workers for specified
      * Execution.
      * @param  {Object} execution  Object that contains information of Execution
-     * @param  {number} numWorkers number of workers to allocate
      * @return {Promise}           [description]
      */
-    function allocateWorkers(execution, numWorkers) {
```
godber: I removed the second argument `numWorkers` here ... accidentally at first.  As it turns out, it's not necessary.  The job starts with the requested number of workers and workers can be added and removed ... and set with total.

packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/k8sResource.js
```diff
@@ -0,0 +1,187 @@
...
+                    name: volume.name,
+                    mountPath: volume.path
+                });
+            });
+        }
+    }
+
+    _setResources() {
+        // The settings on the executions override the cluster configs
+        const cpu = this.execution.cpu || this.terasliceConfig.cpu || -1;
```
godber: @peterdemartini Is there a better way to do this sort of thing in js or is this OK?

godber:
This PR resolves:

* #914 
* #918

packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/k8sResource.js
peterdemartini: I think that it is fine, it’s readable. However it’s vulnerable to errors in the case when the template doesn’t have those deeply nested properties or they are not an array, but I doubt that will be a problem as long as there are proper tests.

godber: Yeah, both `execution` and `terasliceConfig` are relying on their respective schemas to ensure that they exist and are `undefined`.

peterdemartini: Actually in-order to make `this.execution.cpu || this.terasliceConfig.cpu || -1;` work `0` (a falsy value), you'd have to use something like `_.get(this.execution, 'cpu', _.get(this.terasliceConfig, 'cpu', -1))`

packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/k8sResource.js
```diff
@@ -0,0 +1,187 @@
...
+            });
+            this.resource.spec.template.spec.containers[0].volumeMounts.push({
+                name: this.terasliceConfig.assets_volume,
+                mountPath: this.terasliceConfig.assets_directory
+            });
+        }
+    }
+
+    _setVolumes() {
+        if (_.has(this.execution, 'volumes') && (this.execution.volumes != null)) {
```
peterdemartini: This can be simplified to 
```suggestion
        if (this.execution.volumes != null) {
```

packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/k8sResource.js
```diff
@@ -0,0 +1,187 @@
...
+    _setImagePullSecret() {
+        if (this.terasliceConfig.kubernetes_image_pull_secret !== '') {
+            this.resource.spec.template.spec.imagePullSecrets = [
+                { name: this.terasliceConfig.kubernetes_image_pull_secret }
+            ];
+        }
+    }
+
+    _setAssetsVolume() {
+        if ((this.terasliceConfig.assets_directory !== '') && (this.terasliceConfig.assets_volume !== '')) {
```
peterdemartini: This can be simplified to:

```suggestion
        if (this.terasliceConfig.assets_directory && this.terasliceConfig.assets_volume) {
```

packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/k8sResource.js
```diff
@@ -0,0 +1,187 @@
...
+        const templateKeys = ['{{', '}}'];
+
+        return (config) => {
+            const templated = barbe(templateData, templateKeys, config);
+            return JSON.parse(templated);
+        };
+    }
+
+    _setImagePullSecret() {
+        if (this.terasliceConfig.kubernetes_image_pull_secret !== '') {
```
peterdemartini: You catch more cases by using a truthy check, so this can be simplified to
```suggestion
        if (this.terasliceConfig.kubernetes_image_pull_secret) {
```

packages/teraslice/test/lib/cluster/services/cluster/backends/kubernetes/k8sResource-spec.js
```diff
@@ -0,0 +1,401 @@
...
+            const kr = new K8sResource(
+                'deployments', 'worker', terasliceConfig, execution
+            );
+
+            expect(kr.resource.kind).toBe('Deployment');
+            expect(kr.resource.spec.replicas).toBe(2);
+            expect(kr.resource.metadata.name).toBe('ts-wkr-example-data-generator-job-7ba9afb0-417a');
+
+            // The following properties should be absent in the default case
+            expect(kr.resource.spec.template.spec).not.toHaveProperty('affinity');
```
peterdemartini: `expect().toHaveProperty()` works with dot notation, so this can be simplified to 
```suggestion
            expect(kr).not.toHaveProperty('resource.spec.template.spec.affinity');
```

Issue #1000, opened 2019-02-05
refactor cli workers command


jeffmontagna:
Refractor the cli workers command to use the new config class.

jeffmontagna:
Worker subcommand was refactored in PRs #1001 and #1014.

PR #1001, opened 2019-02-05
refactored worker cmd to use new config class

jeffmontagna:
refs #1000


packages/teraslice-cli/lib/utils.js
```diff
@@ -1,9 +1,30 @@
 'use strict';
 
+const _ = require('lodash');
+
 function getTerasliceClient(cliConfig) {
     return require('teraslice-client-js')({ host: cliConfig.clusterUrl });
 }
 
+async function getTerasliceClusterType(terasliceClient) {
```
godber: This function along with my cluster state comment above makes me wonder whether we need a higher level Teraslice class.  It could abstract away the Client util above, the `.clusterType` have some `.worker` type methods or getters.

PR #1013, opened 2019-02-07
Multivalue fix

jsnoble:
- this should allow multipe `multivalue` settings on post_process extractions with mutate: true


kstaken:
This needs a rebase and tests appear to be failing.

PR #1022, opened 2019-02-13
Transform readme fixes

jsnoble:


docs/packages/ts-transforms/operations.md
```diff
@@ -27,7 +27,7 @@ Example:
 
 ```
 
-- other_match_required : `Booelan`(optional) = If set to true, then it sets a conditional for this field that the final object must have other keys that do not have the `other_match_required` flag to appear in the resulting object in order for this field to be set. This is used to pass along certain fields/metadata if another field passes extraction/validation etc.
+- other_match_required : `Booelan`(optional) = If set to true, then it sets a conditional for this field that the final object must have other keys that do not have the `other_match_required` flag to appear in the resulting object in order for this field to be set. This is used to pass along certain fields/metadata if another field passes extraction/validation etc. You can also specify a `selector` on the same configuration with `other_match_required` to make sure that logic is only applied to a particular rule pipline.
```
ciorg: I have a hard time following this paragraph.  The example below helps, but this is a tricky concept to explain.   My suggestion is to simplify the first sentence and rely on the example to do most of the explaining.

docs/packages/ts-transforms/overview.md
```diff
@@ -25,9 +25,19 @@ yarn global add ts-transforms
...
+* [operations reference](./operations.md)
+
+## Plugins
+This library provides a wide array of manipulations/validations etc but you may need a custom operation. You may do so by making a plugin and injecting it.
+* [Plugin reference](./plugins.md)
+
 # Usage
 
-Example file: `rulesFile.txt`
+There are two different execution: `Matcher` and `Transform`. The former returns the raw matching documents while the later provides additional ETL capabilities
```
ciorg: execution should be plural

docs/packages/ts-transforms/overview.md
```diff
@@ -195,42 +205,32 @@ This phase will go through all the configurations and apply all the extractions
 ```
 
 ### Post Process Phase
-This phase is for any additional processing that needs to occur after extraction. Each post_process configuration is affecting the `target_field` of the chained configuration if you use the `tag` and `follow` tags. You can chain multiple times if needed.
+This phase is for any additional processing that needs to occur after extraction. Each `post_process` configuration is affecting the `target_field` of the chained configuration if you use the `tag` and `follow` tags. You can chain multiple times if needed. This phase also include `validation` operations and can be freely chained to each other.
+
```
ciorg: includes should be plural

docs/packages/ts-transforms/overview.md
```diff
@@ -195,42 +205,32 @@ This phase will go through all the configurations and apply all the extractions
...
+- **tag** = marks the config and the target_field with an ID so other configurations can chain off of it
+- **follows** = marks the config that it is chaining off the tag id
```
ciorg: follows should be follow

docs/packages/ts-transforms/overview.md
ciorg: approved with a few suggestions

PR #1024, opened 2019-02-15
Teraserver Data Access Plugin

peterdemartini:
Add a new GraphQL API for managing data-access in [teraserver](https://github.com/terascope/teraserver)


peterdemartini:
I believe I have a functioning of the graphql api built for data-access, and I verified the plugin works. However, I still need to verify this will work to restrict queries for two different users. Once that is done, I will remove WIP.

peterdemartini:
There are the remaining tasks to complete that I am aware of. However I recommend we do those in a separate in PR in order and move forward with reviewing and merging this.

- [ ] add permission rules to manage the ACL
- [ ] add user session support
- [ ] add support for the search counter
- [ ] add time series/history index support

Issue #1003, opened 2019-02-06
Refactor teraslice-cli jobs subcommand


godber:
Refactor teraslice-cli jobs subcommand

godber:
Done

PR #1030, opened 2019-02-27
refactored teraslice-cli jobs commands

jeffmontagna:
Refactored teraslice-cli jobs
- Moved jobs logic into a job class
- Adding checking before start/stoping jobs if job is already running or stopped before changing status
- Moved all display and prompts into display class
- Removed annotations

refs #1003


godber:
I've read through this now.  Obviously it would be good to improve test coverage but everything (or nearly everything) has a test so we won't have to get over that hurdle to extend coverage.  I'll play with this in the morning and merge then.

packages/teraslice-cli/cmds/jobs/run.js
```diff
@@ -1,29 +1,34 @@
...
+const Config = require('../../lib/config');
+const Jobs = require('../../lib/jobs');
+const YargsOptions = require('../../lib/yargs-options');
+
+const yargsOptions = new YargsOptions();
 
-exports.command = 'run <cluster_sh> [job_id]';
-exports.desc = 'Run a job\n';
+exports.command = 'run <cluster-alias> [id]';
+exports.desc = 'run job specificed by id \n';
```
godber: change `specificed` to `specified`

PR #1031, opened 2019-03-01
Remove state and analytics store refresh

### LINKED ISSUE #988: Adjust shard count on state cluster indices
Currently indices are created with 5 shards and 1 replica. This is the ES default.

At a minimum we should probably increase the default to 15 shards. 

Ideally ... both shard count and replicas would be configurable on a per index basis at the cluster master level.

### LINKED ISSUE #990: Eliminate need for refresh=true on state record creation
This is something that has always been a scalability concern in regard to our use of state storage. I don't think we actually hit this limit yet but as our workloads scale we surely will at some point. This was originally added to ensure correctness on jobs that process slices extremely quickly. 

We should be able to handle this scenario by handling the 404 not found with retries at the point of state update.

That will introduce additional latency into the process but I think we can optimize around that through queue management in the execution controller.

peterdemartini:
- Removes state and analytics refreshes when creating and updating records to improve performance and prevent the cluster for being overloaded with large jobs.
- Add the ability to configure the index shards and replicas for analytics, assets, execution, jobs, and state, see the example below.

```yaml
terafoundation:
  connectors:
    elasticsearch:
      default:
        host:
          - "127.0.0.1:9200"

teraslice:
  master: true
  master_hostname: "127.0.0.1"
  name: "teracluster"
  index_settings:
    analytics:
      number_of_shards: 8
      number_of_replicas: 1
    assets:
      number_of_shards: 5
      number_of_replicas: 2
    jobs:
      number_of_shards: 5
      number_of_replicas: 2
    execution:
      number_of_shards: 5
      number_of_replicas: 2
    state:
      number_of_shards: 15
      number_of_replicas: 1
```

Resolves #990 
Resolves #988

### DIFF
--- a/packages/teraslice/lib/config/schemas/system.js
+++ b/packages/teraslice/lib/config/schemas/system.js
@@ -92,6 +92,58 @@ const schema = {
             }
         }
     },
+    index_settings: {
+        analytics: {
+            number_of_shards: {
+                doc: 'The number of shards for the analytics index',
+                default: 5,
+            },
+            number_of_replicas: {
+                doc: 'The number of replicas for the analytics index',
+                default: 1,
+            },
+        },
+        assets: {
+            number_of_shards: {
+                doc: 'The number of shards for the assets index',
+                default: 5,
+            },
+            number_of_replicas: {
+                doc: 'The number of replicas for the assets index',
+                default: 1,
+            },
+        },
+        jobs: {
+            number_of_shards: {
+                doc: 'The number of shards for the jobs index',
+                default: 5,
+            },
+            number_of_replicas: {
+                doc: 'The number of replicas for the jobs index',
+                default: 1,
+            },
+        },
+        execution: {
+            number_of_shards: {
+                doc: 'The number of shards for the execution index',
+                default: 5,
+            },
+            number_of_replicas: {
+                doc: 'The number of replicas for the execution index',
+                default: 1,
+            },
+        },
+        state: {
+            number_of_shards: {
+                doc: 'The number of shards for the state index',
+                default: 5,
+            },
+            number_of_replicas: {
+                doc: 'The number of replicas for the state index',
+                default: 1,
+            },
+        },
+    },
     action_timeout: {
         doc: 'time in milliseconds for waiting for a action ( pause/stop job, etc) to complete before throwing an error',
         default: 300000,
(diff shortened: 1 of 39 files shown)


packages/teraslice/lib/config/schemas/system.js
```diff
@@ -92,6 +92,58 @@ const schema = {
             }
         }
     },
+    index_settings: {
+        analytics: {
+            number_of_shards: {
+                doc: 'The number of shards for the analytics index',
+                default: 5,
```
godber: From: https://github.com/terascope/teraslice/issues/988

> At a minimum we should probably increase the default to 15 shards.

I read this that the default should be 15.

kstaken:
It's unclear to me how the worker handles the scenario where a state record doesn't exist yet.

peterdemartini:
It just retries a Not Found error in updateState function in the state store. Although now that I think of it, it should probably not be treated the same as a time out error, since it shouldn’t retry nearly as many times. I can address that on Monday.

kstaken:
When you look at that can you also verify the backoff behavior is solid? This problem we're having this week seems to have teraslice really pounding the cluster in a way that doesn't make sense. When the state cluster dies it can't even recover until all the teraslice workers are stopped. We've run jobs with more workers in the past and not seen it hammer state so hard.

packages/teraslice/lib/config/schemas/system.js
peterdemartini: I read that as, I want to be able to configure the shards to be 15 for production cluster. I can set the default to 15 but that seems like all clusters?

kstaken: Yeah, this change should also address the underlying reason for that ticket so the current default of 5 should be fine.

godber: 5 is a lot in the small cluster/dev case (e.g. why more than one if you only have one node).  The cost of making it 15 by default is virtually nothing.

peterdemartini: I see your point, but my gut feeling was the configuration should lean towards using minimal defaults vs production. I am fine setting it 15 as long that is acceptable for small clusters.

peterdemartini:
I verified that the retry logic works right based on @kstaken's feedback. This PR is ready for review and we need to figure out the appropriate default shards based on @godber feedback.

Issue #1032, opened 2019-03-04
update kubernetes-client


godber:
We're behind on the kubernetes client we should update it.

Issue #1002, opened 2019-02-05
Allow teraslice job to override the teraslice version in k8s


godber:
As long as the communications between the workers/execution controllers and the masters don't change, we should be able to select the teraslice version on the job level.  Right now the version is set to be the same as the master's version, but it doesn't have to be.  If the version is unspecified, it should use the master version, but the value set on the job should override that.

Issue #981, opened 2019-01-25
Teraslice worker versions should report NA in k8s clustering


godber:
In k8s mode, I think teraslice currently reports the master version as the `teraslice_version` for each worker node.  It should probably report `N/A`.  Perhaps each worker listed under `active` should report it's own stats.

```
    "10.6.6.6": {
        "node_id": "10.6.6.6",
        "hostname": "10.6.6.6",
        "pid": "N/A",
        "node_version": "v8.12.0",
        "teraslice_version": "0.44.4",
        "total": "N/A",
        "state": "connected",
        "available": "N/A",
        "active": [...
```

godber:
Rather than reporting `NA`, it could report the worker's version (doing so would be ideal).  I am not sure what's involved in that, probably a lot of fiddly messaging that may not be worth the effort.  Unless it's stored in the deployment, oh, it probably is.

peterdemartini:
I think the reason why we used `N/A` was not that it was difficult to get information but the "node" could potentially have many executions running on it, with different teraslice versions.

godber:
Ok, for now, I am going to do this:

* Set `node_version` and `teraslice_version` to `N/A` at the worker/node level since it doesn't really have meaning in the k8s case (there's not node manager, a single node can run workers with versions that differ)
* Add an `image` property that reports the docker image used, including tag, which in our case provides version and build information.

I think this solution makes a decent compromise, maybe the `/txt` output should show `image` instead of `node_version` and `teraslice_version` in the k8s master case.  We haven't introduced any differences at the `/txt` api based on cluster type, but I'll see if doing so is convenient.

godber:
`/txt/workers` will now read more usefully:

```
######  Workers  ######

assignment  job_id  ex_id  node_id    pod_name                           image
----------  ------  -----  ---------  ---------------------------------  ------------------
master      N/A     N/A    10.0.2.15  teraslice-master-794bdf85f6-dl49k  teraslice-k8sdev:1
```

PR #1033, opened 2019-03-05
K8s improvments march 2019

godber:
This PR addresses the following things

* #1032 - update `kubernetes-client`
* #1002 - allow jobs to specify teraslice image name
* #981 - report pod image in cluster state, improved `/txt/workers`


godber:
Since I got these done this week, I'm going to move these changes to the `v0.50.0` milestone since that's not out yet.  @peterdemartini if you can do a quick review ... if it passes without requests for changes, please merge it.  If it doesn't, I'll be back monday to address changes and it can go back to `v0.51.0`.

PR #1034, opened 2019-03-06
Better docker image and elasticsearch store improvements

### LINKED ISSUE #933: Update docker image to use node:10
Node 10 has performance gains and node 8 will go on maintenance LTS in Jan 2019. We should start to use it in the docker image and for development. Although we will need to continue to support node 8, so we should avoid using node 10 specific features.

peterdemartini:
- Fixes to elasticsearch store availability and intervals
- Use multi-stage docker image, remove base image and use node 10
- Use terafoundation_kafka_connector with ssl support

Resolves #933


peterdemartini:
Waiting https://github.com/terascope/kafka-assets/pull/20 so it can included in the image

PR #1035, opened 2019-03-07
fix job name special character issue in k8s

godber:
Fortunately @macgyver603 caught the k8s jobname invalid DNS character problem (# wouldn't work in job names) a couple days ago.  I remembered and fix it here.

This happened because I added job names into the pod names ... which in turn end up being DNS names.

So this should go into `v0.50.0` or Charlie will be sad.


godber:
The thing I liked most about this PR is I got to use `test.each()`.

godber:
You guys are going to want this if you try and pull a testable build together today.

Issue #1039, opened 2019-03-11
Create pkg binary of teraslice-cli


godber:
We should have a binary distribution of `teraslice-cli` build using [pkg](https://github.com/zeit/pkg).

godber:
So, the associated PR is all thats necessary to run teraslice or teraslice-cli with pkg.  There's nothing to  that step really.  The question is how to build the generation of the `pkg` binary in the context of our travisci build scripts.  I will talk with Peter about the best strategy here.

godber:
I have binaries I can share with people, as soon as I think of a way to do so.

godber:
Running teraslice:

```
./teraslice-macos -c ../../examples/config/teraslice-master.yaml | bunyan
```

Running `teraslice-cli`

```
./teraslice-cli-macos aliases list
cluster           host
----------------  --------------------------------------
local             http://localhost:5678
```

godber:
Closing this for now, `pkg` is dead, replaced by native node functionality, and I'd really like a unified binary of the cli and teraslice itself.

PR #1041, opened 2019-03-11
Use pkg teraslice cli

godber:
Make teraslice and teraslice-cli executable binaries with pkg.

refs #1039


godber:
Let's merge this too, it may not be everything we need to build binary packages, but the change to `teraslice-cli` is definitely needed ... and that example job could stand updating as well.

kstaken:
This needs a rebase

godber:
Made same change here: https://github.com/terascope/teraslice/pull/1083/

PR #1056, opened 2019-03-15
Transforms performance

jsnoble:
This cleans up extraction-phase performance issues.

- Breaking change: this disallows mutate:true in extraction phase


kstaken:
This doesn't appear to be executing extraction rules that have no selector.

kstaken:
More specifically it doesn't appear to be executing rules with no selector and with no other_match_required attribute. The rules where other_match_required is true do get executed.

jsnoble:
- this now throws errors for compact configs
- this now requires multi_value:false for extractions to get array => singular value extractions

PR #1057, opened 2019-03-15
Add DataType, and many other changes to the data-access-plugin

### LINKED ISSUE #1052: Add default access to a space
If a role has access to a space, it should automatically gain access to the default view of a space. It might make sense to automatically create a default view for a space so there is only one way of doing things.

Additionally the role currently has "spaces" reference on its data model, however, it may make more sense to move that the default view.

peterdemartini:
We discussed on Thursday to move things around significantly on the data-access models. This PR has a good portion of those changes and gets things back to a working state. This does not include any of ACL management permissions, multi-tenant support, etc.

**IMPORTANT** if you've installed this before, remove the indices and recreate everything since it has changed.

I have updated the documentation, see [`./docs/packages/data-access-plugin/overview.md`](https://github.com/terascope/teraslice/blob/data-access-improvements/docs/packages/data-access-plugin/overview.md)

Resolves #1052


kstaken:
This doesn't appear to work if you don't define at least one view. Views should only be required if restrictions are needed.

There's an inconsistency on some of the naming

 datatype `typeConfig` vs spaces `search_config`

Given other naming of properties it seems snake case should be the preference.

I'm not sure `search_config.connection` is working. I'm still investigating but I think I have everything configured correctly and the endpoint appears to work but I get no results.

kstaken:
On `search_config.connection` as far as I can tell I have it correct and the connection attribute is saved to the record in elasticsearch but it appears to ignore it on execution.

kstaken:
Scratch my comments on the connection. I got it working.

kstaken:
Overall this feels much better! I'll work through more scenarios next week but preliminary testing is pretty good.

peterdemartini:
@kstaken I fixed those issues

PR #1058, opened 2019-03-17
Improve k8s targets

godber:
Extending Teraslice job targets to add support for k8s preferred affinity and tolerations.

### DIFF
--- a/docs/clustering/k8s.md
+++ b/docs/clustering/k8s.md
@@ -199,7 +199,6 @@ using `requiredDuringSchedulingIgnoredDuringExecution`.
 
 For each entry in `targets`, if the `constraint` is omitted or set to `require`,
 the pod spec will include a
-
 `nodeAffinity` of
 `requiredDuringSchedulingIgnoredDuringExecution` constraint like the one shown
 below:
--- a/docs/clustering/k8s.md
+++ b/docs/clustering/k8s.md
@@ -188,7 +188,6 @@ by applying a taint, you could use `required` to target the label, then use
 ],
 ```
 
-The
 You can specify one or more targets that will
 use [Kubernetes Node Affinity](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity)
 to force nodes onto specific nodes.  The targets specified here will be required
--- a/docs/clustering/k8s.md
+++ b/docs/clustering/k8s.md
@@ -170,7 +170,7 @@ thing:
 
 Using `"constraint": "preferred"` establishes a looser constraint on the label:
 
-```
+```json
 "targets": [
     {"key": "zone", "value": "west", "constraint": "preferred"}
 ],
--- a/docs/clustering/k8s.md
+++ b/docs/clustering/k8s.md
@@ -162,7 +162,7 @@ specify a target with just a `key` and `value`:
 Using the `"constraint": "required"` property, as shown below, achieves the same
 thing:
 
-```
+```json
 "targets": [
     {"key": "zone", "value": "west", "constraint": "required"}
 ],
--- a/docs/clustering/k8s.md
+++ b/docs/clustering/k8s.md
@@ -204,7 +204,7 @@ the pod spec will include a
 `requiredDuringSchedulingIgnoredDuringExecution` constraint like the one shown
 below:
 
-```
+```yaml
         nodeSelectorTerms:
         - matchExpressions:
           - key: <KEY>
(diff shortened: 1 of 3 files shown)


godber:
I think this does what I want at this point ... I will implement a few more mixed target type tests and update docs later today.

godber:
Ok, this seems to work as desired.  I'll leave the WIP on until I update the docs, but it can be reviewed now.  And if we're in a hurry to get this into production we can merge and do docs later.  I just want to describe exactly what this does when there are multiple targets of different `constraint` types, since the behavior differs.

@mkhpalm if you can take a look at the tests (tomorrow is fine) from lines 517-533:

https://github.com/terascope/teraslice/pull/1058/files#diff-89734a715b498f0041dace4b5eac9be0R517

That will give you the idea of what you get when setting multiple targets:

https://github.com/terascope/teraslice/pull/1058/files#diff-89734a715b498f0041dace4b5eac9be0R500

I think that makes sense for me.  Maybe it's weird that `required` and `preferred` do something different when the could be made to do the same thing ... `preferred` is slightly different because it has two lists.  Then `tolerations` are different of course.

I'll update the docs tomorrow.

@peterdemartini if you can do a quick review in the AM that would be good.

docs/clustering/k8s.md
```diff
@@ -120,23 +120,90 @@ will result in the Kubernetes `resources.requests.memory` and
...
+are insufficient resources, they may be scheduled on nodes without this label.
+* `accepted` - Pods may be scheduled to nodes with the kubernetes taint
+provided by this label.  This uses a Kubernetes tolerance.
+
+### Examples
+
+If you want to force pods to run on nodes with a given label, you can simply
+specify a target with just a `key` and `value`:
+
+```
```
peterdemartini: I recommend specifying the language for the code blocks because it makes easier to read.

```suggestion
```json
```

docs/clustering/k8s.md
```diff
@@ -120,23 +120,90 @@ will result in the Kubernetes `resources.requests.memory` and
...
+"targets": [
+    {"key": "zone", "value": "west"}
+],
+```
+
+Using the `"constraint": "required"` property, as shown below, achieves the same
+thing:
+
+```
```
peterdemartini:
```suggestion
```json
```

docs/clustering/k8s.md
```diff
@@ -120,23 +120,90 @@ will result in the Kubernetes `resources.requests.memory` and
...
+"targets": [
+    {"key": "zone", "value": "west", "constraint": "required"}
+],
+```
+
+Using `"constraint": "preferred"` establishes a looser constraint on the label:
+
+```
```
peterdemartini:
```suggestion
```json
```

docs/clustering/k8s.md
```diff
@@ -120,23 +120,90 @@ will result in the Kubernetes `resources.requests.memory` and
...
+    {"key": "zone", "value": "west", "constraint": "preferred"}
+],
+```
+
+If you wanted your Teraslice workers to target a set of nodes with a given label
+but also wanted to guarantee the availability of these nodes for this workload
+by applying a taint, you could use `required` to target the label, then use
+`accepted` to tolerate that taint as follows:
+
+```
```
peterdemartini:
```suggestion
```json
```

docs/clustering/k8s.md
```diff
@@ -120,23 +120,90 @@ will result in the Kubernetes `resources.requests.memory` and
...
+"targets": [
+    {"key": "zone", "value": "west", "constraint": "required"}
+    {"key": "gen4", "value": "NoSchedule", "constraint": "accepted"}
+],
+```
+
+The
```
peterdemartini: I don't the "The" should be here

```suggestion
```

docs/clustering/k8s.md
```diff
@@ -120,23 +120,90 @@ will result in the Kubernetes `resources.requests.memory` and
...
+to force nodes onto specific nodes.  The targets specified here will be required
+using `requiredDuringSchedulingIgnoredDuringExecution`.
+
+[Kubernetes Taints and Tolerations](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/)
+
+### Pod Spec Details
+
+For each entry in `targets`, if the `constraint` is omitted or set to `require`,
+the pod spec will include a
+
```
peterdemartini: The line break shouldn't be here
```suggestion
```

docs/clustering/k8s.md
```diff
@@ -120,23 +120,90 @@ will result in the Kubernetes `resources.requests.memory` and
...
+`nodeAffinity` of
+`requiredDuringSchedulingIgnoredDuringExecution` constraint like the one shown
+below:
 
 ```
```
peterdemartini:
```suggestion
```yaml
```

godber:
Ok, I've updated the docs.  This is ready to go, @peterdemartini if you want to read the k8s.md from like line 190 down to the end of the section.

PR #1059, opened 2019-03-18
Bump version of teraslice to v0.51.0 for release.

godber:
This just bumps the version of teraslice for when we want to do the `v0.51.0` release.


godber:
Internal tests on the `daily-2019.03.18-7567586b` have passed so I am merging this and doing a release.

Issue #1049, opened 2019-03-14
Add access control to data access management


peterdemartini:
To manage the users, roles, views, and spaces data-access needs permission restrictions. This includes creating the concept of an "admin" role. The admin role is the only role to give write access to managing data-access models, additionally, the admin role can see API tokens. A non-admin user should have query access but can only write to its own user record.

peterdemartini:
After discussing this, it will best to create the concept of a user type, which can be one of the following:

- `USER` - A basic client user with read/write access to its user, and read access to the spaces, views and roles it can access.  A `USER` cannot see or query properties on the view, data type, or space that expose fields that expose the underlying data model of the `data_type`. We might also consider giving a `USER` read access the users within its client.
- `ADMIN` - A admin client user with read/write access to the users, roles, spaces, views, data types within its client. However, an `ADMIN` cannot create a `SUPERADMIN`; it can only create a `USER` and `ADMIN` users under the same client.
- `SUPERADMIN` - A user with read/write access to everything.

PR #1064, opened 2019-03-25
Data access permissions

peterdemartini:
Add data-access management access control / permissions, see the updated [docs](https://github.com/terascope/teraslice/blob/data-access-permissions/docs/packages/data-access-plugin/overview.md) for a list the user's permissions.

Resolves: #1049


peterdemartini:
I am going to remove WIP from this PR since we should be able to move on and test the changes for now. I deferred the following changes to another PR:

- a user should only be able to see things his role has access to
- ensure client ids match when associating records
- improve graphql usage and api

kstaken:
Is there a script to bootstrap access for the first time? It looks like the records I created previously are broken due to client_id but you now need a token to make any request so if I reset not sure how to get started again.

I noticed in the docs it says for USER: "List the users for its client, it cannot see the api_token on any user except itself." A USER shouldn't be able to see anything about other users.

peterdemartini:
> Is there a script to bootstrap access for the first time? It looks like the records I created previously are broken due to client_id but you now need a token to make any request so if I reset not sure how to get started again.

I realize the authentication is broken when loading the playground. Also creating a script is really not ideal, because it needs to be ran from within teraserver for the connector and plugin configuration. Instead I changed the way the bootstrap mode works:

```yaml
# Enabling this flag will cause a `SUPERADMIN` user to be created when this plugin is initialized and there are no other users. The `SUPERADMIN`'s credentials are (username: admin, password: admin). Make sure to only run one worker when using this mode.
bootstrap_mode: true
```

> I noticed in the docs it says for USER: "List the users for its client, it cannot see the api_token on any user except itself." A USER shouldn't be able to see anything about other users.

I can make that change, however should the "user" be able to see something about the other users, maybe just the `firstname`?

peterdemartini:
I am going to merge without the ci checks since travis isn't building

PR #1068, opened 2019-03-26
Tjm stop error

ciorg:
* fixed bug where if the stop command was called and the jobs was already stopped then tjm would throw an error
* created tjm-util to reduce code duplication
* removed un-used tjm modules
* added tests for tjm-util
* bumped version


packages/teraslice-cli/cmds/tjm/restart.js
```diff
@@ -20,27 +20,7 @@ exports.handler = async (argv) => {
...
-        const start = await client.jobs.wrap(job.jobId).start();
-        if (!start.job_id === job.jobId) {
-            reply.fatal(`Could not start ${job.name} on ${job.clusterUrl}`);
-        }
-        reply.green(`Started ${job.name} on ${job.clusterUrl}`);
-    } catch (e) {
-        reply.fatal(e.message);
-    }
+    const tjmUtil = new TjmUtil(client, job);
+    await tjmUtil.stop();
```
godber: There are probably cases where `.stop()` doesn't do what what you want and proceeding on to start in those cases might not be ideal.

packages/teraslice-cli/lib/tjm-util.js
```diff
@@ -0,0 +1,48 @@
...
+                reply.yellow(logMessage);
+                return;
+            }
+
+            reply.fatal(e);
+        }
+    }
+}
+
+module.exports = TjmUtil;
```
godber: The addition of this module is definitely an improvement.  I am concerned with the possible overlap between this and `job.stop()` and `job.start()` but reviewing the job code ... there does seem to be a lot more going on there.  So it's probably worth ignoring that concern now.

packages/teraslice-cli/cmds/tjm/restart.js
ciorg: I was thinking that the .stop logic will catch the errors, but it may prove that I need more error handling in command function

PR #1075, opened 2019-04-01
Data Access Plugin Fixes

### LINKED ISSUE #1071: Undefined in createSpace spaces array throws
I got the following error when my script passed a spaces array to createSpace containing an undefined value. Obviously that's an issue in my script but this error is coming from the lucene parser. It seems this should maybe be caught earlier so the user gets a more helpful indication of the problem.

```
[GraphQL error]: Message: error occured while attempting to parse lucene query: id: () , error: Expected "!", "&&", "(", "+", "-", ".", "/", "<", "<=", "=<", "=>", ">", ">=", "AND", "NOT", "OR", "[", "\"", "\\", "{", "||", [^: \t\r\n\x0C{}()"+-/\^~[\]], or whitespace but ")" found., Location: [object Object], Path: createSpace
(node:34474) UnhandledPromiseRejectionWarning: Error: GraphQL error: error occured while attempting to parse lucene query: id: () , error: Expected "!", "&&", "(", "+", "-", ".", "/", "<", "<=", "=<", "=>", ">", ">=", "AND", "NOT", "OR", "[", "\"", "\\", "{", "||", [^: \t\r\n\x0C{}()"+-/\^~[\]], or whitespace but ")" found.

```

### LINKED ISSUE #1072: Confusing error: Unable to determine field
I believe this error comes from a space with require_query: true when you don't provide a query. This won't be helpful to an end user. Should follow the error from the existing impl.

```
{"error":"Search failure: Unable to determine field"}
```

### LINKED ISSUE #1073: Errors leak implementation details
An error like this is not useful to an enduser and reveals details about the implementation that shouldn't be revealed. This should just return a generic 404 Not Found type error or something like "The requested endpoint does not exist"

This should be considered in all cases where an error may display to an end user. In this scenario we err on the side of providing less detailed errors tailored to helping the user. If an error can not provide specific help to the user then it should be generic and not reveal unnecessary implementation details.

```
{"error":"Failure to access /api/v2/:space: Unable to find Space by 'id:\"value\" OR endpoint:\"value\"'"}
```

### LINKED ISSUE #1078: xlucene - query doesn't return results
I'm starting to look at end user queries and comparing behavior between v1 APIs and v2 APIs. The following query form did not resolve correctly. 

In v1 it returned `55` results, in v2 `0`.

```
date:[2019-04-01T01:00:00Z TO *] AND field:value AND otherfield:(1 OR 2 OR 5 OR 15 OR 33 OR 28) AND NOT (otherfield:15 AND sometype:thevalue) AND NOT anotherfield:value
```

This was the first query I've looked at.

peterdemartini:
- More restrictive/descriptive error messages
- Better handling of an empty query in search
- Fix findAll when given a falsy value
- Fix handling of AND NOT statement
- Test fixes
- Update deps

Resolved #1078
Resolves #1072
Resolves #1071 
Resolves #1073


kstaken:
This now has an issue with the legacy APIs that I layered on top. Those had specific errors that they were throwing which are now being caught up in the generic "Error during query execution" handling. I'm just throwing `new  Error`. Is there something else that should be used there to preserve the context specific error?

peterdemartini:
> This now has an issue with the legacy APIs that I layered on top. Those had specific errors that they were throwing which are now being caught up in the generic "Error during query execution" handling. I'm just throwing `new Error`. Is there something else that should be used there to preserve the context specific error?

Yes, to make the error "safe", do the following:

```js
const error = new Error('Uh Oh');
error.context = {
   safe: true,
};
```

FYI: I am adding some more tests right now so hold of on approving this for a moment

peterdemartini:
You can use the `TSError` from the `@terascope/utils` (see example below), which makes it easier to use. We don't use subtypes because I have been using `TSError` since it is more flexible, however eventually we should make subtypes off of the `TSError` once we formalize the subtypes.

```js
const { TSError } = require('@terascope/utils');

new TSError('Failure', {
   statusCode: 422,
   context: {
       safe: true
    }
});
```

peterdemartini:
I am done adding more tests and I think this is ready to go

kstaken:
When you don't have access to a space now you get:

```
{"error":"Failure to access space"}
```

The term `space` is not something that should be surfacing to search users as it won't mean anything. I'd really like these errors to just follow those from the v1 API.

peterdemartini:
@kstaken would the error `Search forbidden` be better?

kstaken:
In v1 you simply got: ```{"error":"Access Denied"}```

There's no reason to be changing these messages. I would really like all the search user messaging to just be compatible if possible.

peterdemartini:
I will try my best but it is more difficult that it was previously because this all of the code doesn't exist single search file, these errors can be throw in many underlying libraries which aren't specific to the search plugin and should not throw vague errors.

With that said I have been setting the status code on the errors so we can map status code `403` to `Access Denied` in the search plugin.

kstaken:
Yes it seems we're just talking about compatibility / messaging at the search plugin level not down through the whole stack.

PR #1089, opened 2019-04-11
re-enable asset specs

godber:
I didn't manage to get these to fail locally, so I've re-enabled the tests and added a remove of a file.  Maybe something will turn up in CI.


godber:
I have these passing in CI.  They were never failing locally.  I don't have a good explanation for why they started failing.  The two changes I made were not motivated by "oh, this should fix it" they were more like "oh, this is a different way of doing it".  Being more explicit about managing an output file is good ... but why the `yarn --cwd` thing stopped working for us ... I do not know.

Of course, I've realized that I need to support `npm` here as the default and use `yarn` optionally.  I'll do that later though.

godber:
Maybe it was a relative directory thing that changed between the old and new ways of running the tests.

PR #1090, opened 2019-04-12
Attempt to pkgify teraslice and the cli

godber:
So the changes here add the following capabilities:

* added a pkg build script, callable by `yarn build:pkg`
* the `teraslice` pkg run as far as I know
    * think about ideal asset location, currently `./assets`
    * assets were uploaded, data generator job to null ran
* the `teraslice-cli` pkg works
    * with the exception of all `assets` sub commands.
        * If I omit `assets init` then the other commands will work
        * The init command appears to not work because it uses yeoman, which in turn uses `editions` and is a known issue: https://github.com/zeit/pkg/issues/519. Options are:
            * implement `init` without yeoman.
            * Use yeoman the way as a stand alone, they way it is meant to be used. (We would implement separate yeoman generator packages and the user would use yeoman directly)
            * Address the `pkg` bug above.
            * "Patch" yeoman to not use `editions`.

Questions (other than the obvious ones from above)

* What do we do with pkgs?  Upload to github as a package as part of our release cycle?


godber:
Another option would be to re-implement the templating with `hygen`:

http://www.hygen.io/quick-start/

I've implemented one of their examples and built it with `pkg` and it functions correctly.

godber:
I have removed the WIP in this PR.  @peterdemartini, if you are good with the changes that are more global (addition of `build:pkg` and build script) then this can be merged.  The scripts aren't meant to be the final solution, just to be a starting place for when we want to tackle this.  Maybe they're not useful enough and should be removed.

The lower level changes to the asset cli code should be accepted because they are functionally equivalent and resolve `pkg` errors.  They are the real point of merging this PR instead of just deleting it.

PR #1091, opened 2019-04-15
Transform cli fixes

jsnoble:
- added -i options for cli to ignore bad data
- extractions do not always use expensive _.get
-some speed refactors for post_process phase and output phase

### DIFF
--- a/packages/ts-transforms/src/command.ts
+++ b/packages/ts-transforms/src/command.ts
@@ -198,7 +198,7 @@ async function initCommand() {
         const results = manager.run(data);
         // tslint:disable-next-line
         if (command.perf) console.timeEnd('execution-time');
-        // process.stdout.write(`${JSON.stringify(results, null, 4)} \n`);
+        process.stdout.write(`${JSON.stringify(results, null, 4)} \n`);
     } catch (err) {
         console.error(err);
         process.exit(1);
--- a/packages/ts-transforms/src/command.ts
+++ b/packages/ts-transforms/src/command.ts
@@ -123,8 +123,11 @@ function parseData(data: string): object[] | null {
             try {
                 results.push(JSON.parse(line));
             } catch (err) {
-                if (!ignoreErrors) {
-                    console.error(`Failed to parse line ${i + 1} data:"${line}"`);
+                const errorMsg = `Failed to parse line ${i + 1} -- "${line}"`;
+                if (ignoreErrors === true) {
+                    console.error(errorMsg);
+                } else {
+                    throw new Error(errorMsg);
                 }
             }
         }
--- a/packages/ts-transforms/src/command.ts
+++ b/packages/ts-transforms/src/command.ts
@@ -124,8 +124,7 @@ function parseData(data: string): object[] | null {
                 results.push(JSON.parse(line));
             } catch (err) {
                 if (!ignoreErrors) {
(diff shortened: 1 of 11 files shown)


packages/ts-transforms/src/command.ts
```diff
@@ -121,8 +123,10 @@ function parseData(data: string): object[] | null {
             try {
                 results.push(JSON.parse(line));
             } catch (err) {
-                logger.error(`Failed to parse "${line}"`);
-                return null;
+                if (!ignoreErrors) {
```
peterdemartini: If `ignoreErrors === true` is set we should still log it

jsnoble: it was never getting logged at all. When -i is set, by default it ignore's parsing errors. That was the intention. If its not set, it throws

packages/ts-transforms/src/operations/lib/transforms/extraction.ts
```diff
@@ -113,25 +119,23 @@ export default class Extraction {
             // @ts-ignore
             if (config.regex) config.regex = formatRegex(config.regex);
             if (config.end === 'EOP') config.end = '&';
+            if (config.source_field.includes('.')) config.sourceFieldWildcard = true;
```
peterdemartini: We should also check for square brackets, i.e. `example["some-field"]`

packages/ts-transforms/src/command.ts
peterdemartini: It was being log using `logger.error(...);`, but since it was returning `null` it would short circuit the parsing. Regardless the error shouldn't go into the abyss even if the `ignoreErrors` is set to `true`

packages/ts-transforms/src/operations/lib/transforms/extraction.ts
jsnoble: That would be out of spec. We only allow fields to be either the singular, or chained by a "."  Its best sometimes to allow only one way of doing things.

packages/ts-transforms/src/command.ts
jsnoble: That may be the case, but since it would always short circuit then essentially it was doing nothing. Running the cli command normally will produce the appropriate error and point to the offending data. For development purposes (which this is mainly for) if you are working with dirty data that might have issues then you can use the -i on purpose to ignore errors so that it actually can run in the first place and second to not get cluttered with errors and get the response. -i is an unsafe mode so to speak, it's purpose to my knowledge would be to throw them to the abyss on purpose. -i is not intended to be used normally

peterdemartini: My policy with errors is they should rarely be ignored completely. My interpretation (and expectation) about a feature like this is to allow the CLI to skip over parsing errors but not fail early, but that doesn't mean they shouldn't outputted. If the errors are silently ignored, a developer can't be informed that there was problem parsing and may blame the rules or something else.

But I guess I am not really aware of the requirements for this, maybe @kstaken could provide better input than I.

packages/ts-transforms/src/command.ts
```diff
@@ -121,8 +123,10 @@ function parseData(data: string): object[] | null {
...
+                    throw new Error(`Failed to parse "${line}"`);
```
peterdemartini: For the parse error we should probably keep track of the line number so they can find the problem easier.

PR #1093, opened 2019-04-17
GraphQL API improvements

### LINKED ISSUE #1053: Add a CLI for data access management
There should be a CLI tool for data access management, the first use case is for an admin user.

peterdemartini:
**Notable Changes:**

- Use existing `teraserver` sessions for user
- Add script for creating `SUPERADMIN` user (see [docs](https://github.com/terascope/teraslice/blob/graphql-api-improvements/docs/packages/data-access-plugin/overview.md#create-initial-superadmin))
- Add the ability to login using `authenticate` in the GraphQL API instead of using headers only
- Add a more relational and functional API for managing the data models.
- Add the ability to pass in more from, size, and sort when search data access models.

**Breaking Changes:**

- Bootstrap Mode has been removed is replaced by a CLI script for creating a `SUPERADMIN` user.
- Query methods are no longer prefixed with `find`, e.g. `findUsers` is now `users` and `findDataType` is now `dataType`
- Foreign IDs are now resolved to the foreign type, e.g.

```js
// query like this won't work anymore
findSpaces(query: "*") {
   views
}
// old output: => {
   views: ["example-id"]
}

// it has return specific fields from the views
spaces(query: "*") {
   views {
       id
   }
}
// new output: => {
   views: [
      { id: "example-id" }
   ]
}
```

**Other changes:**

- Adds a SimpleContext to terafoundation that will works in "single worker context"
- Fix data-access-plugin test using of `_doc` so it won't break elasticsearch 5 usage.
- Code reorganization/refactor in data-access and data-access-plugin.
- Add more utils to `@terascope/utils` and use them across other packages
- Update dependencies

Resolves #1053


kstaken:
It looks like there are issues with the query parameter when no specific criteria is provided.

```
query {
  views {
		name
  }  
}

Implicit fields are restricted, please specify the field
```

```
query {
  views(query: "*") {
		name
  }  
}

Implicit fields are restricted, please specify the field
```

```
  users(query: "") {
		firstname
  }  

Failure to parse xlucene query \" AND id: undefined\", caused by SyntaxError: Expected term, or whitespace but \":\" found.
```

kstaken:
I must have something wrong. 

```
query {
  authenticate(token: "TOKEN") {
    username
  }
  users(query: "username:admin") {
		firstname
  }  
}
```

Results in:
```
{
  "data": {
    "authenticate": {
      "username": "admin"
    },
    "users": []
  }
}
```

kstaken:
This is working better but I tried accessing users as a non admin and still got this.

```
query {
  authenticate(token: "NON_ADMIN_TOKEN") {
    username
  }
  users(query: "*") {
    firstname
    lastname
    username
    api_token
  }  
}

Implicit fields are restricted, please specify the field
```

Searching for it's specific username works and searching for other usernames allows no results as expected.

peterdemartini:
@kstaken this should be fixed now

PR #1095, opened 2019-04-23
better job stopping and no assumptions with asset replace

ciorg:
2 fixes to the cli
* tjm stop checks status before calling stop on a job
* asset --replace doesn't assume the latest asset is the one that needs replacing and checks for the asset version


godber:
These changes look good ignoring my ongoing concern that both the tjm and jobs sub commands maintain their own job management logic.  See https://github.com/terascope/teraslice/blob/master/packages/teraslice-cli/lib/jobs.js#L180-L243

Has anyone considered whether this is necessary or not?  @jeffmontagna & @ciorg 

I'm going to merge this regardless.

ciorg:
I thought those were my bad assumptions that you copied over from the original tjm.  

Jeff and I need to, at some point in the near future, go through the jobs/ tjm code and find the common threads.

PR #1097, opened 2019-05-07
Graphql api for spaces

jsnoble:
This is just a starting point of dynamic role based schema. There is more work to be done but this should be in a good spot to play with until further requirements/questions are taken care of.

This requires that a type_config be set with the types of all the fields of the data (ts-transform date,ip,geo where appropriate).

NOTE: you must specify the api_token in the url `http://localhost:8000/api/v2/spaces?token=3c3e5157f83ecd7edcc1e3edc567cb24863956ea`. Make sure that the endpoint listed in side the playground has the token specified as well.

Types
- ip
- geo
- date
- Elasticsearch numbers (long, integer, short, byte, double, float, half_float)
- Elasticsearch strings (keyword, text)
- boolean
- string
- int

Example Query:
```
query {
  testdata(query:"ip:152.223.244.212") {
    bytes,
    ip,
    created,
    location,
    space2(join: ["created", "location"]) {
      id
    }
  }
}


// Results
{
  "data": {
    "testdata": [
      {
        "bytes": 2852394,
        "ip": "152.223.244.212",
        "created": "2019-04-26T15:00:23.207Z",
        "location": "0.05102, -41.82129",
        "space2": [
          {
            "id": "96669a45-3e2a-4dbe-a34e-3aeb97d1419b"
          }
        ]
      }
    ]
  }
}
```
Queries of different types:
```
 const query1 = `
                query {
                    ${space1}(query: "ipv6:ab88:805e:55db:0750:b143:61ce:e07a:7180" , size: 1){
                        bytes
                        url
                    }
                }
            `;
         // you can also specify query: "ipv6:\\"ab88:805e:55db:0750:b143:61ce:e07a:7180\\""

            const query2 = `
                query {
                    ${space2}(query: "created:>=2019-04-26T08:00:00.000-07:00", size: 2){
                        bytes,
                        bool
                    }
                }
            `;

            const query3 = `
                query {
                    ${space2}(query: "location:(_geo_box_top_left_:\\"83.906320,-100.058902\\" _geo_box_bottom_right_:\\"80.813646,-97.758421\\")", size: 2){
                        bytes,
                        location
                    }
                }
            `;
```


packages/data-access/src/interfaces.ts
```diff
@@ -73,6 +73,11 @@ export interface DataAccessConfig {
     */
     space_id: string;
 
+    /**
+     * The endpoint of the space
+    */
+    endpoint: string;
```
peterdemartini: We should probably name this `space_endpoint` to keep the naming convention

packages/data-access-plugin/src/spaces/resolvers/index.ts
```diff
@@ -0,0 +1,105 @@
...
+        };
+        const queryAccess = new QueryAccess(accessData, logger);
+
+        endpoints[view.endpoint] = async function resolverFn(root: any, args: any, ctx: any, info: GraphQLResolveInfo) {
+            const _sourceInclude = getSelectionKeys(info);
+            const { size, sort, from, join } = args;
+            const queryParams = { index: view.search_config!.index, from, sort, size, _sourceInclude };
+            let { query: q } = args;
+
+            if (root == null && q == null) throw new UserInputError('a query must be provided for the root query');
```
peterdemartini:
```suggestion
            if (root == null && q == null) throw new UserInputError('Invalid request, expected query to nested');
```

packages/data-access-plugin/src/spaces/resolvers/index.ts
```diff
@@ -0,0 +1,105 @@
...
+            if (root && join == null) throw new UserInputError('a join must be provided when querying a space against another space');
```
peterdemartini:
```suggestion
            if (root && join == null) throw new UserInputError('Invalid query, expected join when querying against another space');
```

packages/data-access-plugin/src/spaces/resolvers/index.ts
```diff
@@ -0,0 +1,105 @@
...
+
+            if (join) {
+                if (!Array.isArray(join)) throw new UserInputError('a join must be an array of values');
```
peterdemartini:
```suggestion
                if (!Array.isArray(join)) throw new UserInputError('Invalid join, must be an array of values');
```

packages/data-access-plugin/src/spaces/dynamic-schema.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    endpointList.forEach(endpoint => results.push(` ${endpoint}(join: [String], query: String, size: Int, from: Int, sort: String): [${endpoint}] `));
+
+    return results.join(' ');
+}
+
+function getMappingValue(type: string, id: string) {
+    const results = allTypeMappings[type];
+    if (results == null) throw new ts.TSError(`could not convert mapping type: ${type} from data_type id: ${id}`);
```
peterdemartini:
```suggestion
    if (results == null) throw new ts.TSError(`Invalid mapping type "${type}" for data type "${name}"`);
```

It would also read better to use `name` instead of `id`

kstaken:
Peter has a review requesting changes. Has that been resolved?

Issue #11, opened 2015-08-13
HDFS Sender


kstaken:
Need a sender implementation that can write to a file / set of files in HDFS.

PR #1099, opened 2019-05-15
Add basic UI for data-access

peterdemartini:
- Adds a basic UI with the ability to authenticate and with CRUD support for Users and Roles.
- Fix issue with parsing parens in xlucene.
- Fixes to data-access and data-access-plugin in order to get UI and teraserver to work right.
- Add history index support


kstaken:
When creating a user email is optional but if you leave it out you get:

```
GraphQL error: .email should match format "email"
```

kstaken:
When you create a new user and return to the users list the user does not show up.

kstaken:
When you delete a user from the users list the user is removed but the count is not correct. Also if page size is 10 and the removed user is #11 pushing it on a new page, removing the user gives you a list with no users in it.

kstaken:
The new token button doesn't seem to work for me. Clicking it has no obvious effect. I do see a request to the server but the token doesn't change. What is the expected behavior?

Also probably better to simply have the button say "New Token" as the usage of just the "refresh" icon is unclear. Also in general any time there is an icon only, a tooltip should exist to say what the button does.

kstaken:
It looks like the cancel button on the forms only activates if you click on the text rather than the button it self. The submit button doesn't behave that way.

kstaken:
If you try to edit a 'SUPERADMIN' it complains that the client_id has to be greater than 0 but the whole point of superadmin is to not be assigned to a particular client. Should client_id just be 0 in that case?

kstaken:
Scratch that last comment. I see it is allowing 0 when type is 'SUPERADMIN'

kstaken:
it looks like when client_id is 0 and you edit the user, the client_id value doesn't show in the form and then attempting to submit the form with no changes fails silently.

peterdemartini:
@kstaken I believe this PR is ready for review again

kstaken:
When you generate a new token it looks like the hide button stops functioning, is that intentional?

peterdemartini:
> When you generate a new token it looks like the hide button stops functioning, is that intentional?

It was intentional to how to automatically show the token when generated but not to break the ability to toggle it.

I made it automatically show the token because it wasn't clear a new one was generated.

kstaken:
Yeah showing the token is good but it won't let you toggle to hide it again which seems weird.

peterdemartini:
I fixed it in https://github.com/terascope/teraslice/pull/1099/commits/694b741a2b347b009df66645b5309e56609e8527

Issue #1102, opened 2019-05-20
Flush on shutdown support for processors


peterdemartini:
Currently we have jobs that hold large amounts of in-memory state and that is lost when the job is stopped since there is no mechanism for flushing the data outside of a "slice". Teraslice needs an additional steps added before the worker is shutdown:

1. Finish the current slice and stop accepting new slices.
1. Trigger a new pre-shutdown notice.
1. Send an empty slice though pipeline, skipping the fetcher.
1. Trigger the existing worker shutdown after the slice is complete.

This has potential problems of overloading processors, Out-of-Memory issues, and exceeding the worker shutdown timeout when it shouldn't.

peterdemartini:
Should this prevent retries?

kstaken:
You're outline is good but I think there's a terminology problem here. You're referring to "finish the current slice" as the first step however this whole process is really about finishing the current slice. It should all occur under a single slice_id and that slice_id should not be marked completed until the entire process is completed. I think this is what you're intending but there is an overloading of the term `slice` which is problematic. #1 is just finishing an execution iteration which is something that to this point has been aligned with the slice concept but now is a bit muddled.

You're asking about retries during the finalization phase? It does seem framework level retries would be out of the picture there.

peterdemartini:
I see. What would happen in the scenario when the worker doesn't have a current slice? Does anything special happen when the current slice is in the retry process or is failed?

Also re-using the slice might break the analytics for that slice (probably not a big deal).

kstaken:
Hmm, that's a good point. If the shutdown event happens between slices you could still lose data as I'd expect the worker would simply exit. 

We definitely want to see correct analytics.

Maybe this finalization should be tracked as a separate slice somehow. That may make all these scenarios clearer.

peterdemartini:
In scenario where the worker doesn't have an "active" slice, the last used known slice should be used and duplicate analytics records will be created. The analytics record will need an additional property to indicate that it was a "flushed" record.

PR #1103, opened 2019-05-22
Flush on shutdown

peterdemartini:
- Add the ability to flush in memory state when shutting down
- Add `onFlushStart` and `onFlushEnd` events to the operation lifecycle
- Emit `slice:flush:start` and `slice:flush:end` when respective events are called
- Fix logger flush issue on shutdown
- Fix potential issue when sending bulk request when shutting down
- Formatting and misc fixes
- More more logic into `WorkerExecutionContext` to prevent duplicate implementations in `teraslice-test-harness`
- Update packages and type fixes in `elasticsearch-store` from updated `rambda`
- Test fixes when running from the root of the project

Resolves https://github.com/terascope/teraslice/issues/1102


peterdemartini:
This PR is ready to be reviewed

PR #1100, opened 2019-05-16
Data types

jsnoble:
- [x] needs more individual type tests
- [x] need more specialized/commonly used types beside the base ones
- [x] more integration with existing services

PR #1107, opened 2019-05-31
Add Views and Spaces to Data Access UI

peterdemartini:
## Goals

- Add CRUD UI for Views
- Add CRUD UI for Spaces
- Polish UI and Framework improvements

## Changes

- [x] Add Spaces CRUD models
- [x] Views CRUD models
- [x] Disable changing data type once created on space / view
- [x] Bump utils, data-types, ui libraries, elasticsearch-store, and data access libraries
- [x] Fix model partial updates
- [x] Change space configuration to have a type and a configuration
- [x] Make sure the Client ID is consistent across related models
- [x] Pull in fields from type config when making views
- [x] Validate Space endpoint and other model name fields
- [x] Add the ability to the space on a View (GraphQL API)
- [x] Add better reload support for the UI framework
- [x] Fixes to loading UI framework changes
- [x] Make tslint validate shadowed variables like eslint.
- [x] Add a couple more utilities to `@terascope/utils`
- [x] Fix es error handling context in `@terascope/utils`
- [x] Update dependencies
- [x] Use new DataTypes see https://github.com/terascope/teraslice/pull/1100
- [x] Add a basic / safe mapping update functionality to `elasticsearch-store`
- [x] Rename the property `type_config` to `config` on the data type index to avoid a migration
- [x] Better typescript definition for the `data-types` library
- [x] More restrictive error messages in API
- [x] `yarn setup` will now make sure that dist and build directories are cleaned up first
- [x] Add form buttons on top and bottom of screen
- [x] Add the ability per page count dropdown to the data table

## Extra
- [ ] Add/edit/preview a view from space edit form
- [ ] Add the ability to derive from other data types


peterdemartini:
In order to prevent migrating the `data types` index, I had to rename the property `type_config` to `config` since the data structure changed and elasticsearch didn't really like that.

peterdemartini:
@kstaken this PR is ready to review

kstaken:
Getting a build failure now:

```
[@terascope/ui-components] external "UIComponents" 42 bytes {main} [built]
    + 887 hidden modules

ERROR in /Users/kstaken/projects/opensource/teraslice/packages/ui-data-access/src/ModelForm/Form.tsx
./src/ModelForm/Form.tsx
[tsl] ERROR in /Users/kstaken/projects/opensource/teraslice/packages/ui-data-access/src/ModelForm/Form.tsx(131,34)
      TS2322: Type '{ children: ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component<any, any, any>)> | null; ... 9 more ...; deleteRecord: () => Promise<...>; }' is not assignable to type 'IntrinsicAttributes & Props & { children?: ReactNode; }'.
  Property 'redirectPath' does not exist on type 'IntrinsicAttributes & Props & { children?: ReactNode; }'.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

lerna ERR! yarn run build stderr:
error Command failed with exit code 2.

lerna ERR! yarn run build exited 1 in '@terascope/ui-data-access'
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

peterdemartini:
@kstaken  I am done making additional changes to this PR and I suggest we merge it sooner than later unless you need something specific changed

kstaken:
The delete button in the forms shouldn't have the same placement as the other buttons. Maybe at the bottom left would be a good place to put it for now. We want the option to exist but not be too prominent.

Also the `Submit` button on edit forms would probably be better as `Save`.

kstaken:
Got this error trying to delete a view.

```
Elasticsearch Error: [illegal_argument_exception] [script] unknown field [source], parser not found
```

kstaken:
The view was assigned to a space. Removing the assignment to the space allowed the view to be deleted.

kstaken:
On `Create space` since the selection of the data type affects available views the form should put data type selection above role and view selection.

kstaken:
I created a new data type through the UI then went to create a space for access to the data type and the data type I created does not show up in the list of data type options.

kstaken:
It shows up under `Data Types` but even reloading the app it doesn't show up in the list of options when creating a space or view.

kstaken:
Looking at the raw records the only difference I see is that the UI created data type has an empty `inherit_from` array but the others were created before that was added so that attribute is missing.

kstaken:
Attempting to `Create Space` validation give you the error below before you attempt to save which disables the submit button but then filling in the field does not clear the error and the button remains disabled.

```
Search Configuration is missing index
```

kstaken:
On `Create View` the labels for include / exclude fields could use some love. 

Currently

```
Restricted Fields (inclusive)
A whitelist of fields that can be views and searched
```

Let's make this

```
Allowed Fields
Fields that should be allowed in searches and returned results. If any fields are listed here, all other fields will be excluded automatically.
```

And the corresponding

```
Excluded Fields
Fields that should be excluded from searches and returned results. If fields are listed here, all other fields will be allowed automatically.
```

Also, does the model allow setting both of these at the same time?

kstaken:
Attempting to search as SUPERADMIN still throws this error and the UI doesn't allow me to assign a role to that user.

```
{"error":"Access Denied","debug":{"timestamp":"2019-06-14T19:37:33.901Z","message":"User \"admin\" is not assigned to a role","stack":"TSError: User \"admin\" is not assigned to a role\n    at ACLManager.getViewForSpace (/Users/kstaken/projects/opensource/teraslice/packages/data-access/dist/src/acl-manager.js:371:19)\n    at spaceErrorHandler (/Users/kstaken/projects/opensource/teraslice/packages/data-access-plugin/dist/src/search/index.js:43:50)\n    at /Users/kstaken/projects/opensource/teraslice/packages/data-
```

kstaken:
Hitting a search endpoint without providing the `q` parameter results in an `Access denied` error even when the user has access to the space. Adding a query works but this endpoint is defined to not require a query. The `Access Denied` error in this scenario definitely doesn't make sense. I fear this was confusion about what I was saying related to ES being unavailable. I certainly didn't mean to extend that to override other errors.

peterdemartini:
I fixed all but one of the problems since I was unable to replicate the issue. When I created a data type and then immediately created a space, it showed up. Is it possible the `client_id`'s didn't match?

kstaken:
`client_ids` match. Only difference I could find was the `inherit_from` on the raw record. Tried reloading, restarting teraserver just won't give me that data type as an option. I'll try regenerating the other types and see if it makes a difference.

kstaken:
If I have a space that is defined to not require a query I get 

```
{"error":"Query restricted"}
```

if no query is provided. Those endpoints should just execute a query with no criteria.

PR #1115, opened 2019-06-14
Data type cli

jsnoble:
this provides a basic cli for using data-types

-h shows commands, parameters and examples

### DIFF
--- a/packages/data-types/bin/data-types.js
+++ b/packages/data-types/bin/data-types.js
@@ -1,0 +1,15 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const path = require('path');
+
+try {
+    // this path.join is only used for pkg asset injection
+    path.join(__dirname, '../package.json');
+    require('../dist/src/command');
+} catch (err) {
+    // eslint-disable-next-line
+    console.error('error while attempting to invoke cli command', err.toString());
+    process.exit(1);
+}
--- a/packages/data-types/src/command.ts
+++ b/packages/data-types/src/command.ts
@@ -68,7 +68,9 @@ function wrapper(cb: CB) {
 
 function parseSettings(settings: string) {
     const results = {};
+    if (settings == null) return results;
     const data = Array.isArray(settings) ? settings : [settings];
+
     data.forEach((str) => {
         const setting = str.split(':');
         if (settings.length !== 2) throw new Error(`setting "${str}" is not fomatted correctly, please follow "key:value" format`);
--- a/packages/data-types/src/command.ts
+++ b/packages/data-types/src/command.ts
@@ -39,9 +39,9 @@ yargs
         {},
         wrapper(getXluceneValues)
     )
-    .example('yarn data-types es-mapping --name=event -s index.number_of_shards:20 -s index.number_of_replicas:1', '')
-    .example('\nyarn data-types gql --name=event', '')
-    .example('\nyarn data-types xlucene', '')
+    .example('$0 es-mapping --name=event -s index.number_of_shards:20 -s index.number_of_replicas:1', '')
+    .example('$0 gql --name=event', '')
+    .example('$0 xlucene', '')
     .help('h')
     .alias('h', 'help')
     .version(version)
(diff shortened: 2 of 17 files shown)


packages/data-types/bin/data-types.js
```diff
@@ -0,0 +1,15 @@
...
+
+const path = require('path');
+
+try {
+    // this path.join is only used for pkg asset injection
+    path.join(__dirname, '../package.json');
+    require('../dist/src/command');
+} catch (err) {
+    // eslint-disable-next-line
+    console.error('error while attempting to invoke cli command', err.toString());
```
peterdemartini: This file has some odd behavior, first `path.join` will only throw an error if a non-string is passed and since the result is never being used, it is useless. Second I think try catching here is an odd choice, if there an expected error you should let it be thrown with a full stack track and by node, if you are trying to detect if the `../dist/src/command` exists, then you should use `fs.existsSync` first and throw an error that says the that library hasn't been built properly.

packages/data-types/src/command.ts
```diff
@@ -0,0 +1,165 @@
...
+import { DataType, DataTypeConfig } from './index';
+
+// @ts-ignore
+const logger = debugLogger('ts-transform-cli');
+// change pathing due to /dist/src issues
+const packagePath = path.join(__dirname, '../../package.json');
+const { version } = JSON.parse(fs.readFileSync(packagePath, 'utf8'));
+
+yargs
+    .command('es-mapping', 'create an elasticsearch mapping from the provided data type',
```
peterdemartini: I think naming this `elasticsearch` or just `es` would be better

packages/data-types/src/command.ts
```diff
@@ -0,0 +1,165 @@
+import yargs from 'yargs';
+import path from 'path';
+import fs from 'fs';
+import { debugLogger, get } from '@terascope/utils';
+import { DataType, DataTypeConfig } from './index';
+
+// @ts-ignore
+const logger = debugLogger('ts-transform-cli');
```
peterdemartini: Don't copy and paste
```suggestion
const logger = debugLogger('ts-transform-cli');
```

packages/data-types/bin/data-types.js
jsnoble: Most, if not all of the reasons it was done this way was to make this pkg(able). When we were working on ts-transforms we spent several tickets/issues to get this to work with the way pkg was being used. 'path.join' is only for 'pkg' to be able to notice that this is an additional asset to load in the environment. The error as well provided the error desired when dealing with pkg.

packages/data-types/src/command.ts
jsnoble: I dont like the name either but I think just putting in 'elasticsearch' or 'es' would be to vague. There is no direct understanding from the name that this command produces a mapping and not anything else related to elasticsearch

peterdemartini: That should be fine I guess

packages/data-types/src/command.ts
jsnoble: fixed

packages/data-types/src/command.ts
```diff
@@ -0,0 +1,165 @@
...
+            demandOption: true
+        }
+    },
+    wrapper(getGraphQlSchema)
+    )
+    .command('xlucene', 'create an elasticsearch mapping from the provided data type',
+        {},
+        wrapper(getXluceneValues)
+    )
+    .example('yarn data-types es-mapping --name=event -s index.number_of_shards:20 -s index.number_of_replicas:1', '')
```
peterdemartini: It is not safe to assume the script name will be `yarn data-types` use `$0` instead.
```suggestion
    .example('$0 es-mapping --name=event -s index.number_of_shards:20 -s index.number_of_replicas:1', '')
```

packages/data-types/src/command.ts
```diff
@@ -0,0 +1,165 @@
...
+    .example('\nyarn data-types gql --name=event', '')
```
peterdemartini: Same as above, also there is no reason to add a newline in front of this.

```suggestion
    .example('$0 gql --name=event', '')
```

packages/data-types/src/command.ts
```diff
@@ -0,0 +1,165 @@
...
+    .example('\nyarn data-types xlucene', '')
```
peterdemartini: Same as above
```suggestion
    .example('$0 xlucene', '')
```

packages/data-types/src/command.ts
jsnoble: fixed

packages/data-types/src/command.ts
jsnoble: fixed

PR #1130, opened 2019-06-18
Improved e2e tests

peterdemartini:
Hopefully this will help improve the stability of the e2e tests. This PR includes:

- Pass in logger to teraslice-messaging so errors can be properly logged
- Impoved error messages and logging (so if errors happens in the tests we can see them)
- Change size of example data to 100 from 10 to avoid race conditions
- Add resource constraints to docker images

### DIFF
--- a/packages/teraslice-messaging/src/messenger/client.ts
+++ b/packages/teraslice-messaging/src/messenger/client.ts
@@ -4,7 +4,7 @@ import * as i from './interfaces';
 import { Core } from './core';
 import { newMsgId } from '../utils';
 
-const logger = debugLogger('teraslice-messaging:client');
+const _logger = debugLogger('teraslice-messaging:client');
 
 export class Client extends Core {
     readonly socket: SocketIOClient.Socket;
@@ -18,8 +18,16 @@ export class Client extends Core {
     protected serverShutdown: boolean;
 
     constructor(opts: i.ClientOptions) {
-        super(opts);
-        const { hostUrl, clientId, clientType, serverName, socketOptions = {}, connectTimeout } = opts;
+        const { hostUrl, clientId, clientType, serverName, socketOptions = {}, connectTimeout, logger, ...coreOpts } = opts;
+
+        super({
+            logger: logger
+                ? logger.child({
+                    module: 'messaging:client',
+                })
+                : _logger,
+            ...coreOpts,
+        });
 
         if (!isString(hostUrl)) {
             throw new Error('Messenger.Client requires a valid hostUrl');
@@ -103,12 +111,12 @@ export class Client extends Core {
         });
 
         this.socket.on('reconnecting', () => {
-            logger.trace(`client ${this.clientId} is reconnecting`);
+            this.logger.trace(`client ${this.clientId} is reconnecting`);
             this.ready = false;
         });
 
         this.socket.on('reconnect', () => {
-            logger.trace(`client ${this.clientId} reconnected`);
+            this.logger.trace(`client ${this.clientId} reconnected`);
             this.ready = true;
             this.emit('ready');
 
@@ -118,18 +126,18 @@ export class Client extends Core {
                 try {
                     await this.sendAvailable();
                 } catch (err) {
-                    logger.warn('update availablilty on reconnect error', err);
+                    this.logger.warn('update availablilty on reconnect error', err);
                 }
             });
         });
 
         this.socket.on('disconnect', () => {
-            logger.debug(`client ${this.clientId} disconnected`);
+            this.logger.debug(`client ${this.clientId} disconnected`);
             this.ready = false;
         });
 
         this.socket.on('shutdown', () => {
-            logger.info(`server ${this.serverName} shutdown`);
+            this.logger.debug(`server ${this.serverName} shutdown`);
             this.ready = false;
             this.serverShutdown = true;
             this.emit('server:shutdown');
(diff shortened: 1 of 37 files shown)


packages/teraslice-messaging/src/messenger/client.ts
```diff
@@ -118,18 +126,18 @@ export class Client extends Core {
...
 
         this.socket.on('disconnect', () => {
-            logger.debug(`client ${this.clientId} disconnected`);
+            this.logger.debug(`client ${this.clientId} disconnected`);
             this.ready = false;
         });
 
         this.socket.on('shutdown', () => {
-            logger.info(`server ${this.serverName} shutdown`);
+            this.logger.debug(`server ${this.serverName} shutdown`);
```
jsnoble: does this really need to change from info to debug? Its really helpful to know when something actually shutsdown, or is getting the shutdown command. When we have problems, we need logs to know the state something is going through.

packages/teraslice-messaging/src/messenger/client.ts
```diff
@@ -143,15 +151,15 @@ export class Client extends Core {
         });
 
         this.socket.on('connect', () => {
-            logger.debug(`client ${this.clientId} connected`);
+            this.logger.trace(`client ${this.clientId} connected`);
```
jsnoble: The biggest things that could break or have issues would be connection and shutdown. Does this really need to be trace? If something is not working (in tests etc) we would be running in debug mode. Does this event really not matter that much?

packages/teraslice-messaging/src/messenger/server.ts
```diff
@@ -335,18 +345,18 @@ export class Server extends Core {
 
     protected updateClientState(clientId: string, state: i.ClientState): boolean {
         if (this._clients[clientId] == null) {
-            logger.warn(`${clientId} does not exist and cannot be updated`);
+            this.logger.debug(`${clientId} does not exist and cannot be updated`);
             return false;
```
jsnoble: Are these really worthy of debug? seems like they are important things. In production, we only have 'info' logs to work off of and if something goes awry we lose more context about what happened

packages/teraslice-messaging/src/messenger/server.ts
```diff
@@ -422,9 +432,9 @@ export class Server extends Core {
 
         socket.on('disconnect', (error?: Error | string) => {
             if (error) {
-                logger.error(`client ${clientId} disconnected with error`, error);
+                this.logger.info(`client ${clientId} disconnected`, { error });
```
jsnoble: why should an error log be changed to an info log? If it just disconnected that fine, but if its because of an error it should be an error

packages/teraslice/lib/cluster/services/cluster/backends/native/messaging.js
```diff
@@ -151,7 +151,9 @@ module.exports = function messaging(context, logger) {
 
         const selfHasEvent = _.some(selfMessages, type => type[eventName] != null);
         if (!selfHasEvent) {
-            throw new Error(`"${self}" cannot register for event, "${eventName}", in messaging module`);
+            throw new Error(
+                `"${self}" cannot register for event, "${eventName}", in messaging module`
```
jsnoble: there are several logs like this that seem to change it for no reason. Why is this linting deciding to newline this out?

packages/teraslice-messaging/src/messenger/client.ts
peterdemartini: Previously this wasn't logged at all.

packages/teraslice-messaging/src/messenger/server.ts
peterdemartini: Most logging messages in this package should be debug or trace since this could generate a lot of unwanted noise. The reason why it switched from `warn` to `debug` was because the logger was designed for debugging this package and now it is using a logger from teraslice and it is not really a warning.

packages/teraslice-messaging/src/messenger/server.ts
peterdemartini: That is because the error is just a message that says it was removed from the namespace

peterdemartini:
> It seemed like you have recently changed your linting styles and it seems to have made PR's a lot more noisy. Some of these linting changes don't make much sense for me.
> 
> In some places it does wierd things
> 
> ```
>  new Error('some message')
> // is now 
>  new Error(
> 'some message'
> )
> 
> // why?
> 
> // for promises in some places its inconsistent
> asyncFn().then().then()
> // and other places
> asyncFn()
>    .then()
>    .then()
> ```
> 
> What I am trying to say that if we need linting rules to change, put them in the project so editors can pick them up when we write things. And hopefully the linting rules act consistently

What you are referring to is not linting but formatting, I have been using prettier in my editor because it helps move fast without wasting time reformatting changes. The reason why it formats some errors on Error on three lines is because originally the like exceeded the maximum line length of 100 and it tries to make it as short as possible. There is a similar reason for formatting the promise chains that way, it does that so it will fill the line as much as possible. I agree it looks ugly but I think that is good because it encourages me to refactor my code so it is not insanely nested.

But you are right some of that was extra noise, I reverted some of unnecessary the formatting changes.

We really do need to lock down formatting and linting because I don't think any of us like having these conversations and we all have different opinions on how it should be done.

jsnoble:
100% agreed

PR #1132, opened 2019-06-19
Fix empty query

### LINKED ISSUE #1116: Defining a Space to not require a query still requires a query.
Setting `require_query: false` on a space should allow a search with no query criteria. It appears to be ignored and still requires a query to be provided or you receive the following error from the search endpoint.

```
{"error":"Query restricted"}
```

peterdemartini:
Fix xlucene's translation of an empty query. I also change the code so we use the empty query over a * query.

Fixes #1116

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -20,6 +20,7 @@ assets
 autoload
 packages/*/assets
 packages/*/autoload
+packages/*/bench
 
 # exclude packages that are not needed inside the docker image
 packages/chunked-file-reader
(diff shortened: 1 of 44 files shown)


kstaken:
I'm still not getting any results when I hit a search endpoint where the query is supposed to be optional.

kstaken:
This is showing out of date.

kstaken:
Is this able to merge or are you still trying to fix the tests?

peterdemartini:
Still trying to fix the tests. The e2e test are being problematic.

PR #1169, opened 2019-06-28
Add DATAADMIN user type 

peterdemartini:
This PR adds support for the `DATAADMIN` user type.


kstaken:
I get

```
User doesn't have permission to write to that client
```

Any time I try to save something. DATAADMIN User was created with client_id:1 and all data records are for the same client.

peterdemartini:
> I get
> 
> ```
> User doesn't have permission to write to that client
> ```
> 
> Any time I try to save something. DATAADMIN User was created with client_id:1 and all data records are for the same client.

This should be fixed now

Issue #950, opened 2019-01-10
Conventional Metadata Properties on DataEntity


peterdemartini:
@kstaken recommended we add conventional properties for the `DataEntity` metadata. Currently, the creation of a `DataEntity` will add `_created` property.

I purpose we add the following **recommended** properties:

```ts
interface DataEntityMetadata {
    /** The time at which the DataEntity was created **/
    readonly _createTime: number;

    // conventional properties...

    /** The time at which the data was ingested into the source data */
    _ingestTime?: Date;

    /** The time at which the data was consumed by the reader */
    _processTime?: Date;

    /**
    * The time associated with this data entity,
    * usually off of a specific field on source data or message
    */
    _eventTime?: Date;

    /** A unique key for the data which will be can be used to key the data */
    _key?: string;
}
```

Since storing multiple Date objects on each metadata property will decrease performance, it will be best if we use a getter and setter to abstract the Date object creation until needed.

kstaken:
Are there still places this needs to be applied? Has it been more generally documented outside this ticket?

peterdemartini:
It’s been partially applied to the elasticsearch assets and the kafka assets. They both need to have configuration to pull the evenTime from the incoming data. We still need documentation for this and we need to formalize the date format/usage.

kstaken:
This just needs access APIs to further solidify the convention.

PR #1170, opened 2019-06-28
State storage

jsnoble:
introduces the state_storage library

### DIFF
--- a/packages/teraslice-state-storage/package.json
+++ b/packages/teraslice-state-storage/package.json
@@ -30,7 +30,7 @@
     },
     "dependencies": {
         "@terascope/elasticsearch-api": "^2.0.3",
-        "@terascope/job-components": "^0.19.0",
+        "@terascope/job-components": "0.20.1",
         "@types/lru-cache": "^5.1.0",
         "bluebird": "^3.5.1",
         "lru-cache": "^5.1.1"
--- a/packages/teraslice-state-storage/package.json
+++ b/packages/teraslice-state-storage/package.json
@@ -1,0 +1,39 @@
+{
+    "name": "@terascope/teraslice-state-storage",
+    "description": "State storage operation api for teraslice",
+    "version": "0.1.0",
+    "publishConfig": {
+        "access": "public"
+    },
+    "homepage": "https://github.com/terascope/teraslice/tree/master/packages/teraslice-state-storage#readme",
+    "repository": "git@github.com:terascope/teraslice.git",
+    "author": "Terascope, LLC <info@terascope.io>",
+    "license": "MIT",
+    "bugs": {
+        "url": "https://github.com/terascope/teraslice/issues"
+    },
+    "files": [
+        "dist/src/**/*"
+    ],
+    "srcMain": "src/index.ts",
+    "main": "dist/src/index.js",
+    "typings": "dist/src/index.d.ts",
+    "scripts": {
+        "lint": "tslint -p tsconfig.json -t verbose -e '**/*.json'",
+        "lint:fix": "yarn lint --fix",
+        "prepublishOnly": "yarn build",
+        "build": "tsc --build --pretty",
+        "build:watch": "yarn build --watch",
+        "test": "jest",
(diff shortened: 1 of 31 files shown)


packages/teraslice-state-storage/src/elasticsearch-state-storage/index.ts
```diff
@@ -0,0 +1,197 @@
...
+        this.cache.shutdown();
+    }
+
+}
+
+function has(data: object, key: any) {
+    return key in data;
+}
+
+function chunk<T>(dataArray: T[], size:number) {
```
peterdemartini: Can we move `chunk` and `has` to `@terascope/utils` see `packages/utils/src/utils.ts`?

packages/teraslice-state-storage/src/elasticsearch-state-storage/index.ts
```diff
@@ -0,0 +1,197 @@
...
+
+        return bulkRequest;
+    }
+
+    private _esBulkUpdate(docArray: DataEntity[]) {
+        const bulkRequest = this._esBulkUpdatePrep(docArray);
+
+        return bPromise.map(chunk<ESBulkQuery>(bulkRequest, this.chunkSize),
+            chunkedData => this.es.bulkSend(chunkedData)
+            );
```
peterdemartini: This is weirdly formatted

packages/teraslice-state-storage/src/elasticsearch-state-storage/index.ts
```diff
@@ -0,0 +1,197 @@
...
+        this.chunkSize = config.chunk_size;
+        this.persist = config.persist;
+        this.persistField = config.persist_field || config.id_field;
+        this.cache = new CachedStateStorage(config);
+        this.es = esApi(client, logger);
+    }
+
+    private getIdentifier(doc: DataEntity) {
+        const id =  doc.getMetadata(this.IDField);
+        if (id == null) throw new TSError(`There is no field "${this.IDField}" set in the metadata for doc: ${JSON.stringify(doc)}`);
```
peterdemartini: It would be good to catch an empty string as well.

```suggestion
        if (!id) throw new TSError(`There is no field "${this.IDField}" set in the metadata`, {
             context: { doc }
        });
```

packages/teraslice-state-storage/src/cached-state-storage/index.ts
```diff
@@ -0,0 +1,70 @@
...
+        this.IDField = config.id_field;
+        this.cache = new LRU({
+            max: config.cache_size,
+            maxAge: config.max_age
+        });
+    }
+
+    private getIdentifier(doc: DataEntity) {
+        const id =  doc.getMetadata(this.IDField);
+        if (id == null) throw new TSError(`There is no field "${this.IDField}" set in the metadata for doc: ${JSON.stringify(doc)}`);
```
peterdemartini: This should also catch empty strings and we should avoid stringifying the doc and exposing too much info (use the TSError context).
```suggestion
        if (!id) throw new TSError(`There is no field "${this.IDField}" set in the metadata`, {
            context: { doc }
        });
```

packages/teraslice-state-storage/package.json
```diff
@@ -0,0 +1,39 @@
...
+        "build": "tsc --build --pretty",
+        "build:watch": "yarn build --watch",
+        "test": "jest",
+        "test:watch": "jest --coverage=false --notify --watch --onlyChanged",
+        "test:debug": "env DEBUG='*teraslice*' jest --detectOpenHandles --coverage=false --runInBand"
+    },
+    "dependencies": {
+        "@terascope/elasticsearch-api": "^2.0.3",
+        "@terascope/job-components": "0.20.1",
+        "@types/lru-cache": "^5.1.0",
```
peterdemartini: The `@types/` libraries should most always be dev dependencies

peterdemartini:
@kstaken should we be using `_key` not `_id` in this case? Ref #950

kstaken:
Yes of course. @jsnoble 100% my fault.

PR #1180, opened 2019-07-01
Fix data access issues

### LINKED ISSUE #1172: Space will allow you to assign two views for the same role
When you create a space it's possible to assign two views for the same role to the same space.  Should probably warn on adding of the second view and maybe provide the name of the conflicting view already assigned to the space.

### LINKED ISSUE #1173: View -> Spaces interaction is problematic in views context
When you create a view you're given a complete list of roles to choose from. Once you save and then go back to edit the same view, without adding it to a space, your selected role will not show and the list of roles will be empty so you can't change anything.

Then if you add it to a space the list of roles assigned to the space will show and if your assigned role is also assigned to the space it will show up.

I think this interaction between views and spaces is incorrect. It seems to assume a 1 to 1 relationship between views and spaces but the relationship is many to many. The same view can be associated with many spaces and those spaces can have different top level role assignments so making a 1 to 1 link between the view and a space in `views` context will always be problematic.

It would certainly be good to see all associated spaces for a view but it needs to be a list and I'm not sure the role list can be linked as it appears to be currently.

peterdemartini:
Resolves #1172 
Resolves #1173

### DIFF
--- a/packages/data-access-plugin/package.json
+++ b/packages/data-access-plugin/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@terascope/data-access-plugin",
     "description": "A teraserver plugin for managing data access and searching spaces",
-    "version": "0.10.0",
+    "version": "0.10.1",
     "publishConfig": {
         "access": "public"
     },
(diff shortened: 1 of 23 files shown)


kstaken:
I'm getting this error building on node:8. Have we moved off node:8 as target?

```
error global-agent@2.0.0: The engine "node" is incompatible with this module. Expected version ">=10.0". Got "8.11.2"
error Found incompatible module

```

peterdemartini:
Seems like we may have accidentally. I will double check.

kstaken:
Would it be possible to tweak this message

```
Unable to assign View "Composite View 2" to Space with a conflicting Roles
```

to be more like this. 

```
Unable to assign View "Composite View 2" as it has Role conflicts with existing Views.
```

Ideally it would provide more detail on the exact conflicts but that may be tricky if there are multiple conflicts.

PR #1196, opened 2019-07-08
exposed underlying cache and more of its apis

jsnoble:


packages/teraslice-state-storage/src/elasticsearch-state-storage/index.ts
```diff
@@ -110,10 +110,9 @@ export default class ESCachedStateStorage {
         return cached;
     }
 
-    async mget(docArray: DataEntity[]) {
+    async mget(docArray: DataEntity[], cb = (doc: DataEntity) => doc) {
```
peterdemartini: The variable name `cb` should is a naming convention for when an async function completes, not for mapping functions. It would be more clear by naming it `mapFn` or just `fn`.

jsnoble: I made the change

PR #1201, opened 2019-07-09
GraphQL Spaces search fixes

jsnoble:


peterdemartini:
The title on this is very misleading, this fixes a field name regex in `ui-data-access`.

packages/data-access-plugin/src/query-point/index.ts
```diff
@@ -45,7 +45,7 @@ export default class SpacesPlugin {
     async shutdown() {}
 
     registerRoutes() {
-        const managerUri = '/api/v2/spaces';
+        const managerUri = '/api/v2/qp';
```
peterdemartini:
```suggestion
        const queryPointURI = '/api/v2/qp';
```

packages/data-access-plugin/src/query-point/index.ts
```diff
@@ -45,7 +45,7 @@ export default class SpacesPlugin {
...
         this.logger.info(`Registering data-access-plugin space manager at ${managerUri}`);
```
peterdemartini:
```suggestion
        this.logger.info(`Registering data-access-plugin query-point at ${queryPointURI}`);
```

PR #1266, opened 2019-07-24
Ts transforms updates

jsnoble:
To use multiple capturing regex groups you need to specify multivalue:true on the ts-transforms config.


peterdemartini:
Do you have a sample rules file and data set I can use to test this?

PR #1300, opened 2019-08-05
allow more config options for csv parser in chunked-file-reader

macgyver603:


packages/chunked-file-reader/lib/formatters.js
```diff
@@ -58,8 +58,9 @@ function csv(incomingData, logger, opConfig, metadata, slice) {
         headers: opConfig.fields,
         trim: true,
         noheader: true,
-        ignoreEmpty: true,
-        output: 'json'
+        ignoreEmpty: opConfig.ignore_empty || false,
+        output: 'json',
+        ...opConfig.extra_args
```
peterdemartini: We should avoid using the splat for `...` operator in javascript because it isn't compatible with node 8. Using `Object.assign` instead.

macgyver603: fixed

PR #1301, opened 2019-08-05
v0.54.7 Add performance metrics

peterdemartini:
Using https://github.com/newrelic/node-native-metrics I added the ability to collect performance metrics.


packages/teraslice/lib/workers/metrics/index.js
```diff
@@ -0,0 +1,132 @@
...
+            max_time: loopMetrics.usage.max,
+            sum_of_squares: loopMetrics.usage.sumOfSquares,
+            count: loopMetrics.usage.count
+        });
+    }
+
+    _emitMetric(type, data) {
+        const msg = {
+            type,
+            timestamp: makeISODate(),
```
godber: Do we end up with two dates in the log message with this?

peterdemartini:
@godber do you think having the enabled flag make it more confusing to read or would it be more clear to optional construct it? `this.metrics = performanceMetrics ? new Metrics(...) : null`

packages/teraslice/lib/workers/metrics/index.js
peterdemartini: Well each log message has a timestamp but if we were to change how this was reported it may be useful.

godber:
> @godber do you think having the enabled flag make it more confusing to read or would it be more clear to optional construct it? `this.metrics = performanceMetrics ? new Metrics(...) : null`

I think `enabled` is a little indirect.  But since it's not exposed to the end user, it might not be worth fussing over.

Issue #1299, opened 2019-08-05
Move more scripts to ts-scripts


peterdemartini:
The following scripts should be moved to ts-scripts in order to our monorepo tools more portable.

- [x] `scripts/publish-packages.sh`
- [x] `scripts/docker-release.sh`
- [ ] `scripts/publish-documentation.sh`
- [ ] `scripts/build-pkg.sh`

We should also add a generate changelog script

peterdemartini:
This can be closed I think

PR #1304, opened 2019-08-06
v0.54.9 - Fixes and ts-scripts improvements

### LINKED ISSUE #1282: Elasticsearch request error when adding field types to data type
when adding new field types to a datatype I get:
```
Request Error
Elasticsearch Error: [mapper_parsing_exception] failed to parse
```
I was adding 3 new fields, 2 are marked as arrays, though I don't think that has anything to do with the error.  I have saved field types as arrays and not seen the error.  

I have not noticed a pattern to this error, it seems random.  I will add to this ticket if I see it again.

Screen Shot 2019-07-30 at 1 10 42 PM

peterdemartini:
- Moves npm and docker build scripts to `ts-scripts`
- Makes ts-scripts more generic and flexible
- Fixes `ui-data-access` Issues
- Use `utils` error parsing vs the depecrated `error_parser` parsing in teraslice (just one file change)
- Fix error handling issue in `elasticsearch-api->_createIndex`
- Dependency and consistency fixes
- Improved performance metrics using `gc-stats` and `event-loop-stats`

Ref #1299 
Resolves #1282


godber:
Reviewing

packages/teraslice/package.json
```diff
@@ -74,7 +72,12 @@
     },
     "displayName": "Teraslice",
     "terascope": {
+        "main": true,
         "enableTypedoc": false,
         "testSuite": "elasticsearch"
+    },
+    "optionalDependencies": {
```
godber: I'm not sure it makes sense to make these optional dependencies.  Doing so makes this effectively a build time choice, but the feature is triggered with a job option.  That would mean users could try to use it, only to discover that support was not provided at build time.

I might feel differently if these packages are particularly huge or something.  If you leave these as optional dependencies, you might want to change your error message in your catch statements.  `Failure to construct gc-stats` should be more along the lines of `Failure to construct gc-stats, gs-stats module may not be installed`.

packages/teraslice/test/workers/helpers/configs.js
```diff
@@ -72,7 +72,8 @@ const newConfig = (options = {}) => {
         lifecycle,
         max_retries: maxRetries,
         operations,
-        performance_metrics: false,
+        // this will give it a 20% it will be enabled
+        performance_metrics: Math.random() < 0.2,
```
godber: This introduces non-determinism into your tests.  I question if it's worth it.

packages/teraslice/package.json
peterdemartini: Actually our dockerfile ignores optional dependencies so they should be regular dependencies. Fixing it now

PR #1311, opened 2019-08-08
Ts transforms notification

jsnoble:


packages/ts-transforms/src/loader/rules-loader.ts
```diff
@@ -26,14 +37,34 @@ export default class RulesLoader {
...
+        if (ruleFile) {
+            inputStream = fs.createReadStream(ruleFile);
+        } else if (notifications != null && notifications.length > 0) {
+            const notificationsRules = [notifications];
+
+            inputStream = new Readable({
+                read() {
+                    const rule = notificationsRules.pop();
+                    if (!rule) {
+                        this.push(null);
```
peterdemartini: You should return here?
```suggestion
                        this.push(null);
                        return;
```

packages/ts-transforms/src/loader/rules-loader.ts
```diff
@@ -26,14 +37,34 @@ export default class RulesLoader {
         return JSON.parse(strConfig);
     }
 
-    private async fileLoader(ruleFile: string): Promise<OperationConfigInput[]> {
+    private async dataLoader(ruleFile: string|false, notifications?: string): Promise<OperationConfigInput[]> {
         const parseConfig = this.parseConfig.bind(this);
```
peterdemartini: Is there a reason why this being bound to this?

packages/ts-transforms/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "ts-transforms",
-    "version": "0.21.1",
+    "version": "0.21.2",
```
peterdemartini: This should be a minor release

packages/ts-transforms/src/loader/rules-loader.ts
```diff
@@ -26,14 +37,34 @@ export default class RulesLoader {
...
         const results: OperationConfigInput[] = [];
         const errorResults: string[] = [];
         let hasError = false;
 
+        let inputStream;
+
+        if (ruleFile) {
+            inputStream = fs.createReadStream(ruleFile);
+        } else if (notifications != null && notifications.length > 0) {
+            const notificationsRules = [notifications];
```
peterdemartini: Why make this an array if there is only one? Additionally it makes it even more weird to call it notifications.

packages/ts-transforms/test/matcher-spec.ts
```diff
@@ -74,4 +75,47 @@ describe('matcher', () => {
...
+            'alasdf{789q;lk]erasp98d7@'
+        ];
+
+        const config1: WatcherConfig = {
+            types: { _created: 'date' },
+            notification_rules: rules1.join('\n')
+        };
+
+        try {
+            await opTest.init(config1);
```
peterdemartini: If this passes, the test should fail, but currently this will pass since there is expectation.

```suggestion
            await expect(opTest.init(config1)).toReject();
```

packages/ts-transforms/src/loader/rules-loader.ts
```diff
@@ -26,14 +37,34 @@ export default class RulesLoader {
         return JSON.parse(strConfig);
     }
 
-    private async fileLoader(ruleFile: string): Promise<OperationConfigInput[]> {
+    private async dataLoader(ruleFile: string|false, notifications?: string): Promise<OperationConfigInput[]> {
```
peterdemartini: Overloading this function like this is weird, mainly because you can specify both parameters and if you want to specify the notifications parameter you have to disable the first. It seems like maybe this should broken into three function a `fileLoader`, `notificationsLoader` and a `dataLoader`. Ideally the `dataLoader` would have the common logic and take in a `Reader`.

packages/ts-transforms/test/matcher-spec.ts
```diff
@@ -74,4 +75,47 @@ describe('matcher', () => {
...
+        } catch (err) {
+            expect(startsWith(err.message, 'Failure to parse xlucene query')).toEqual(true);
```
peterdemartini: Using `jest-extended` will make this error more readable if it fails.

```suggestion
            expect(err.message).toStartWith('Failure to parse xlucene query');
```

packages/ts-transforms/src/loader/rules-loader.ts
jsnoble: notifications is a \n delimitated list of xlucene rules, it is set into an array to more easily make a readable stream. The stream must push 'null' to signal its done reading.

PR #1332, opened 2019-08-13
Query Point Updates

jsnoble:
I am posting this here until further feedback on complexity, the default should limit to 1,000,000.

```
data_access: {
                namespace: `some_name_space`,
                complexity_limit: {number here},
            },
```
The complexity is calculated by the size multiplied by the size of each subsequent join

```
query myQuery{
  test_space1(query: "*", size: 10) {
    bytes,
    test_space2(join:["bytes"]) {
      bool,
      bytes
      ip,
      test_space3(join: ["bytes"], size: 100) {
        wasFound
      }
    }
  }
}
```
If size is not set, it default to elasticsearch size which is 10;
This query should have a complexity score of 10,000  (10 * 10 * 100).


kstaken:
I think this is a reasonable first attempt and will definitely address my initial concern. 

```
The query was rejected due to its complexity score: 10000000000, please reduce the size on the queries/joins
```

The message is pretty good. I'd suggest adding commas to the score and a little tweak to the second part.

```
The query was rejected due to its complexity score: 10,000,000,000. Please reduce the "size" parameter on the queries/joins.
```

kstaken:
Are you controlling concurrency on the join requests?

kstaken:
Does this support joining on multiple fields? It looks like if it does it's ORing the fields when it should AND but I'm not sure that's what's going on.

kstaken:
On geospatial we moved queries into xlucene but this will still likely need explicit support for geo sorting which appears to be missing. 

We'll also need to figure out how to join on geospatial fields, this will likely need some thought as the prototype didn't support this in a general manner. I believe with the typing we should have what we need to do this automatically just based on providing the field name in the `join` list.

jsnoble:
that issue has already been addressed

kstaken:
Trying to build the ui using `yarn setup && yarn build:all` on this branch continually fails with this error:

```
WARNING in ../xlucene-evaluator/dist/src/parser/engine.js
Module not found: Error: Can't resolve '../../peg/peg-engine.js' in '/Users/kstaken/projects/opensource/teraslice/packages/xlucene-evaluator/dist/src/parser'
 @ ../xlucene-evaluator/dist/src/parser/engine.js
 @ ../xlucene-evaluator/dist/src/parser/parser.js
 @ ../xlucene-evaluator/dist/src/parser/index.js
 @ ../xlucene-evaluator/dist/src/document-matcher/index.js
 @ ../xlucene-evaluator/dist/src/index.js
 @ ../data-types/dist/src/types/versions/v1/boolean.js
 @ ../data-types/dist/src/types/versions/mapping.js
 @ ../data-types/dist/src/graphql-helper.js
 @ ../data-types/dist/src/index.js
 @ ./src/DataType/config.ts
 @ ./src/DataType/routes.ts
 @ ./src/index.ts
```

kstaken:
Playground just gives an error. 

```
{
  "error": "Not allowed to request resource"
}
```

kstaken:
Running a join on a large volume of data I received this error. 

```
UserInputError: Invalid join on field \"location\", you may only join on a field that is being returned in a query
```

This exact query worked at size 10,000 but when I pushed it to size 100,000 this error occurred. The location field was defined in the result of the outer query. I don't know if this is because there was a record in the result set that didn't have the field (very likely) or if it was something else related to the size. Obviously you can not assume every record is going to have the necessary data in it.

kstaken:
When you reload the playground it normally persists the tabs and queries you have open and automatically restores them. It stores this state in browser storage so even if you close the window and come back your state will be back. We seem to have lost this capability, now when you reload it resets as new which was very unexpected and lost the query I was working with.

kstaken:
Does this have support for geospatial sorting? I'm not seeing the parameters in the schema.

packages/data-access-plugin/src/query-point/dynamic-server.ts
```diff
@@ -33,18 +42,34 @@ export class DynamicApolloServer extends apollo.ApolloServer {
...
                 if (this.playgroundOptions && req.method === 'GET') {
-                    // perform more expensive content-type check only if necessary
-                    // XXX We could potentially move this logic into the GuiOptions lambda,
-                    // but I don't think it needs any overriding
                     const accept = accepts(req);
                     const types = accept.types() as string[];
                     const prefersHTML = types.find((x: string) => x === 'text/html' || x === 'application/json') === 'text/html';
-
                     if (prefersHTML) {
+                        const endpoint = `${req.protocol}://${req.headers.host}${req.baseUrl}${req.url.slice(1)}`;
```
peterdemartini: Manually formatting and piecing together the URL like this may cause problems at some point. We shouldn't block this PR because this change but eventually we need to to use something like the [nodejs URL lib](https://nodejs.org/dist/latest-v10.x/docs/api/url.html) to format it. There are a ton of edge cases are around URLs even in something as simple as this.

PR #1333, opened 2019-08-13
v0.55.0 Terafoundation improvements

### LINKED ISSUE #960: Connectors should support names with dashes
Currently, terafoundation will find connectors by using the convention `terafoundation_${name}_connector`, I suggest we add support for `terafoundation-${name}-connector`.

### LINKED ISSUE #962: Terafoundation apps crash when config contains endpoints for which modules are not installed
Terafoundation apps crash when the config contains endpoints for which modules are not installed.  Instead, terafoundation should just emit a WARN log entry saying that the module is not available.

peterdemartini:
- Support connectors with dashes and namespaces
- Warn to `stderr` instead of throw and error when a connector exists in the schema but cannot be found.
- Support es module default exports for connectors
- Better error handling in `terafoundation`
- [BREAKING] Remove `ops_directory` in the `terafoundation` config (this was broken and isn't used)
- [BREAKING] Remove `elasticsearch` as a `terafoundation` logging option
- Add a bunch more test coverage
- Refactor some of the files in `terafoundation`
- Switch millisecond duration options in schema to use [`duration`](https://github.com/mozilla/node-convict#validation) instead of `Number`
- Remove dependency of `moment` and `lodash` in `terafoundation`
- Add a couple utils to `@terascope/utils` 

One of the advantages of using `duration` is that you can human readable durations like `1 minute` or `30 seconds` or `2 hours` 

Resolves #960 
Resolves #962


peterdemartini:
Let hold merging this, I found a bug in teraslice that I should fix in another PR before this

peterdemartini:
I can't find any references outside of the old terafoundation repo. I already removed and updated the documentation I could find.

PR #1352, opened 2019-08-19
v0.55.1 - Add env_vars to job and teraslice config

peterdemartini:
Add `env_vars` to job and teraslice config

For example:

```js
{
   // ...
   "env_vars": {
        "UV_THREADPOOL_SIZE": "10"
   }
   // ...
}
```


packages/job-components/src/interfaces/jobs.ts
```diff
@@ -58,6 +58,7 @@ export type JobConfig = Partial<ValidatedJobConfig>;
 export interface ValidatedJobConfig {
     analytics: boolean;
     assets: string[];
+    /** This may wont exist until ran in an execution */
```
godber:
```suggestion
    /** This may not exist until ran in an execution */
```

peterdemartini:
I merging this since the tests pass

PR #1366, opened 2019-08-20
v0.55.4 - Fixes to messaging and logging

peterdemartini:
- Log error first and switch logger.error instead of logger.fatal in some places
- Fix unexpected shutdown logic in shutdown
- Retry send slice complete forever
- Fix issue where disconnect timeout could be much longer than expected
- Fixes to worker shutdown
- Improvements to logging to help narrow down socket connection issues
- Fix a minor issue with the test-reader slicer
- Change default `action_timeout` to `2 minutes`


packages/teraslice-messaging/src/messenger/interfaces.ts
```diff
@@ -75,7 +72,6 @@ export interface UpdateClientState {
 export enum ClientState {
     Offline = 'offline',
     Online = 'online',
-    Disconnected = 'disconnected',
```
kstaken: Why isn't this state required anymore? How will its removal affect the analytics provided by the execution controller?

peterdemartini: This shouldn’t introduce any change to the workers_disconnected stats and disconnect logic. Previously there were two disconnected states, Disconnected (when the socket connection disconnects) and Offline (disconnect time + the work disconnect timeout). The workers_disconnected was track by the same event that has now been to consolidated to now “Offline”. Given the confusion I should rename it to “Disconnected” because that would be more clear. 

We were tracking when the worker is goes “disconnected” and “offline” and again when it went “Offline” in the execution controller, which lead to duplication and incorrect time tracking.

peterdemartini: I changed it to disconnected to make it more clear.

peterdemartini:
I have manually tested a few different scenarios that these changes would affect, I think this PR is ready to be reviewed again @godber  @kstaken

PR #1406, opened 2019-09-05
ts-transforms cli stream and jexl expressions

jsnoble:
- should now be able to steam ldjson into the cli. `-f ldjson` to stream, 
- cli writes ldjson to console
- added jexl expressions to extractions, specified by `exp` on extraction rules
- source/target are the default now instead of source_field/target_field
- a post processor that is an extraction will not have to specify a `post_process: 'extraction'` parameter
- fully tested out cli
- added a few util functions


peterdemartini:
This needs a more descriptive title and labels added. So when we look back through the history it will easier to find things.

kstaken:
I'm not clear on what the purpose of the streaming batch size is. Is there an actual benefit to that? It seems like it should just work on a per record basis and setting `-s 1` does exactly what I would expect so I'm not seeing the usefulness of the option.

Beyond that this looks really good from a functionality perspective. 

Branch needs to be updated and it looks like jexl errors could be handled better. Currently it throws a stack so they seem unhandled.

```
/snapshot/teraslice/node_modules/jexl/lib/Expression.js:53
    if (res.error) throw res.error
                   ^

Error: Invalid expression token: #@
    at Lexer._createToken (/snapshot/teraslice/node_modules/jexl/lib/Lexer.js:158:13)
    at Lexer.getTokens (/snapshot/teraslice/node_modules/jexl/lib/Lexer.js:89:26)
    at Lexer.tokenize (/snapshot/teraslice/node_modules/jexl/lib/Lexer.js:129:17)
    at Expression.compile (/snapshot/teraslice/node_modules/jexl/lib/Expression.js:28:26)
    at Expression._getAst (/snapshot/teraslice/node_modules/jexl/lib/Expression.js:72:26)
    at _eval.promise.resolve.then (/snapshot/teraslice/node_modules/jexl/lib/Expression.js:59:24)
    at PromiseSync.then (/snapshot/teraslice/node_modules/jexl/lib/PromiseSync.js:25:23)
    at Expression._eval (/snapshot/teraslice/node_modules/jexl/lib/Expression.js:58:30)
    at Expression.evalSync (/snapshot/teraslice/node_modules/jexl/lib/Expression.js:52:22)
    at Jexl.evalSync (/snapshot/teraslice/node_modules/jexl/lib/Jexl.js:146:20)
```

packages/utils/src/utils.ts
```diff
@@ -84,6 +84,11 @@ export function ensureBuffer(input: string|Buffer, encoding: BufferEncoding = 'u
     throw new Error(`Invalid input given, expected string or buffer, got ${getTypeOf(input)}`);
 }
 
+export function isBoolean(input: any): boolean {
```
peterdemartini: It is important to use the is keyword so the type for the input will be updated outside of this function.

```suggestion
export function isBoolean(input: any): input is boolean {
```

packages/ts-transforms/src/phases/extraction-phase.ts
```diff
@@ -33,9 +30,10 @@ export default class ExtractionPhase extends PhaseBase {
         if (!this.hasProcessing) return dataArray;
         const resultsList: DataEntity[] = [];
 
-        for (let i = 0; i < dataArray.length; i += 1) {
-            const results = createTargetResults(dataArray[i]);
-            runExtractions(this.phase, dataArray[i], results);
+        for (const doc of dataArray) {
+            const results = createTargetResults(doc);
+            // @ts-ignore
```
peterdemartini: Is there a good to add this `// @ts-ignore`?

PR #1407, opened 2019-09-05
Remove data access packages

peterdemartini:
Removes the following packages:
   - ui-data-access
   - ui-core
   - ui-components
   - data-access
   - data-access-plugin


peterdemartini:
If case of emergency, I created a [branch](https://github.com/terascope/teraslice/tree/fixes-for-data-access) from master, in-case there needs to be any unexpected changes.

PR #1414, opened 2019-09-09
WIP: Add DataWindow

### LINKED ISSUE #1306: Add DataWindow extension of DataEntity
We need a framework level DataWindow class that represents an array of DataEntities and is it self a DataEntity. We experimented with this concept in standard-assets so this is just a generalization of those experiments.

peterdemartini:
We need a generalized approach of a `DataWindow`, which is a time or keyed window of DataEntities. Here is my proposal:

```ts
/**
 * Acts as an array of DataEntities associated to a particular key or time frame.
 * A `DataWindow` should be able to be used in-place of an array in most cases.
 */
export class DataWindow extends Array {
    /**
     * A utility for safely creating a `DataWindow`
     */
    static make(input: DataEntity[]|DataWindow, key?: string): DataWindow {
        if (DataWindow.isDataWindow(input)) {
            return input;
        }

        const window = new DataWindow(...input);
        if (key) {
            window.setKey(key);
        }
        return window;
    }

    static isDataWindow(input: any): input is DataWindow {
        return input instanceof DataWindow;
    }

    /**
     * Get the metadata for the `DataWindow`.
     *
     * If a field is specified, it will get that property of the metadata
    */
    getMetadata(_key?: string) {}

    /**
     * Given a field and value set the metadata on the record
    */
    setMetadata(_key: string, _val: any) {}

    /**
     * Get the unique `_key` for the window
     *
     * If no `_key` is found, an error will be thrown
    */
    getKey() {}

    /**
     * Set the unique `_key` for the window
     *
     * If no `_key` is found, an error will be thrown
    */
    setKey(_key: string|number) {}

    /**
     * Get the time at which this window was created.
    */
    getCreateTime(): Date {}

    /**
     * Get the window start time
     *
     * If none is found, `undefined` will be returned.
     * If an invalid date is found, `false` will be returned.
    */
    getStartTime(): Date|false|undefined {}

    /**
     * Set the window start time
     *
     * If the value is empty it will set the time to now.
     * If an invalid date is given, an error will be thrown.
     */
    setStartTime(_val?: string|number|Date): void {}

    /**
     * Get the window completion time
     *
     * If none is found, `undefined` will be returned.
     * If an invalid date is found, `false` will be returned.
    */
    getFinishTime(): Date|false|undefined {}

    /**
     * Set the window completion time
     *
     * If the value is empty it will set the time to now.
     * If an invalid date is given, an error will be thrown.
     */
    setFinishTime(_val?: string|number|Date): void {}
}
```

Resolves #1306


peterdemartini:
I have also debated having it extend a `Set` instead of an `Array` but that would involve more changes to processors since the `Set` API has important differences. However the advantage of using a `Set` would be:
    - a more straightforward API, `->get`, `-set`, `->has`, `->clear`
    - performance improvements when adding and removing large sets of data for a multi-slice duration.
    - we could more easily add an `BigSet` abstraction for dealing with more than 16 millions keys

peterdemartini:
Reasons why I think a `DataWindow` should extend an `Array` and not a `DataEntity`:

1. The metadata and methods can be specific towards a window of `DataEntities` without having to worry about being 100% compatible for the a `DataEntity`. For example:
     - The following metadata properties wouldn't be necessary:
     
       - `_processTime`
       - `_eventTime`
       - `_ingestTime`
     - The following static methods would be confusing or won't work:
       - `makeArray` - would it return an array of `DataWindow`s? Or would it convert a list of `DataEntities` into a `DataWindow`
       - `isDataEntity` - when given a `DataWindow` it technically should return `true` but I could see reasons to return `false`
       - `isDataEntityArray` - when given a `DataWindow` it would return `false` but given a `DataWindow[]` would return `true`. I think that is just confusing, because it is an array.
       - `fromBuffer`
     - The following methods would be confusing or won't work:
       - `getRawData`
       - `setRawData`
       - `toBuffer` 
1. It can be returned as the entire `Slice`
2. Similar to how a `DataEntity` is a simple API around an `object`, this would be a simple wrapper around an `array`.

kstaken:
Yeah, I think you're right.

peterdemartini:
We will want to switch to using DataFrames, I am going to close this, since our implementation in the future will be more different.

PR #1428, opened 2019-09-13
v0.57.1 - Fixes to DataEntity API changes

peterdemartini:
- A `DataEntity` should be able to be remove a `_key` via `entity.setMetadata('_key', undefined);`
- Fix issue with extending a `DataEntity` and getting the instance of the a real `DataEntity`

### DIFF
--- a/packages/utils/src/entities/data-entity.ts
+++ b/packages/utils/src/entities/data-entity.ts
@@ -206,10 +206,6 @@ export class DataEntity<
         if (field === '_createTime') {
             throw new Error(`Cannot set readonly metadata property ${field}`);
         }
-        if (field === '_key') {
-            return this.setKey(value as any);
-        }
-
         this[i.__ENTITY_METADATA_KEY].metadata[field] = value as any;
     }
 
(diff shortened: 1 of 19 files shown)


packages/utils/src/entities/data-entity.ts
```diff
@@ -155,7 +155,13 @@ export class DataEntity<
     }
 
     static [Symbol.hasInstance](instance: any): boolean {
-        return utils.isDataEntity(instance);
+        // prevent automatically thinking a base DataEntity is
+        // an instance of a class that extends from a DataEntity
+        if (this.name !== 'DataEntity' && this.name !== 'Object') {
+            return false;
+        }
+        if (!utils.isDataEntity(instance)) return false;
```
jsnoble: should this not be ?
```
return utils.isDataEntity(instance)
```
do you need the if here? What is utils.isDataEntity returning?

peterdemartini: Makes sense, fixing it now

PR #1440, opened 2019-09-19
Xlucene polygon support and fixes

jsnoble:
- fixed issues around single grouping and string support
- changed how geo is handled, but allow backwards compatibility
-now use geo function
```
location:geoDistance(point:"33.435518,-111.873616" distance:"5000m")
location: geoBox(top_left:"33.906320,-112.758421", bottom_right:"32.813646,-111.058902")
location:geoPolygon(points:["32.536967,-113.968710", "33.435967,-111.867710", "33.435967,-109.867710"])
```
- has geoPolygon support for new geoFunction style only
- updated docs


packages/xlucene-evaluator/src/utils.ts
```diff
@@ -72,7 +72,10 @@ export const GEO_DISTANCE_UNITS: { readonly [key: string]: GeoDistanceUnit } = {
 };
 
 export function parseGeoDistance(str: string): GeoDistanceObj {
-    const matches = trim(str).match(/(\d+)(.*)$/);
+    const matches = trim(str)
+        .replace(/['"]/g, '')
+        .match(/'?"?(\d+)(.*)'?"?$/);
```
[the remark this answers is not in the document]
jsnoble: It was removing the unneccesary quoted values, but I was able to make more changes to the underlying grammar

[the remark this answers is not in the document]
peterdemartini: If there has been made to the grammar, shouldn't we remove the code?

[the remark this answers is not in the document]
jsnoble: its removed

packages/xlucene-evaluator/src/parser/functions/geo/polygon.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    return points;
+}
+
+const geoPolygon: i.FunctionDefinition = {
+    name: 'geoPolygon',
+    version: '1',
+    create(field: string, params: any, { logger }) {
+        if (!field || field === '*') throw new Error('field for geoBox cannot be empty or "*"');
+        // eslint-disable-next-line @typescript-eslint/camelcase
```
peterdemartini:
```suggestion
```

packages/xlucene-evaluator/src/parser/functions/geo/polygon.ts
```diff
@@ -0,0 +1,63 @@
...
+import { geoMatcher } from './helpers';
+import { parseGeoPoint } from '../../../utils';
+import * as i from '../../interfaces';
+import { AnyQuery } from '../../../translator';
+
+function validate(params: i.Term[]) {
+    const geoPointsParam = params.find((node) => node.field === 'points');
+    if (geoPointsParam == null) throw new Error('geoPolygon query needs to specify a "points" parameter');
+    if (!Array.isArray(geoPointsParam.value)) throw new Error('points parameter must be an array');
+    // @ts-ignore we are ignoreing util we have a better story around list expressions
```
peterdemartini: Using`ts-ignore` is really bad practice event for these issues, use `any` in only the spots you need it. If you don't, the compiler will skip over any possible issue in the next line, instead just a particular value.

```suggestion
    // @ts-ignore we are ignoreing util we have a better story around list expressions
```

packages/xlucene-evaluator/src/parser/functions/geo/polygon.ts
```diff
@@ -0,0 +1,63 @@
+
+// @ts-ignore
+import { polygon, lineString } from '@turf/helpers';
+// @ts-ignore
```
peterdemartini:
```suggestion
```

packages/xlucene-evaluator/src/parser/functions/geo/polygon.ts
```diff
@@ -0,0 +1,63 @@
+
+// @ts-ignore
```
peterdemartini:
```suggestion
```

packages/xlucene-evaluator/src/parser/functions/geo/polygon.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (points.length < 3) throw new Error('geoPolygon points parameter must have at least three points');
+
+    return points;
+}
+
+const geoPolygon: i.FunctionDefinition = {
+    name: 'geoPolygon',
+    version: '1',
+    create(field: string, params: any, { logger }) {
+        if (!field || field === '*') throw new Error('field for geoBox cannot be empty or "*"');
```
peterdemartini: Make sure you to double check your copy and pasted code:

```suggestion
        if (!field || field === '*') throw new Error('field for geoPolygon cannot be empty or "*"');
```

packages/xlucene-evaluator/src/parser/interfaces.ts
```diff
@@ -165,3 +177,23 @@ export interface Wildcard extends StringDataType, TermLikeAST {
...
+    create: (
+        field: string,
+        params: any,
+        config: FunctionConfig
+        ) => FunctionMethods;
+}
+
+export interface FunctionMethods {
+    match(arg: any): boolean;
+    toElasticsearchQuery(options: UtilsTranslateQueryOptions): any;
```
peterdemartini: This should return `any` it should have a specific contract:
```suggestion
    toElasticsearchQuery(options: UtilsTranslateQueryOptions): { query: AnyQuery, sort?: AnySortQuery };
```

packages/xlucene-evaluator/src/parser/functions/geo/polygon.ts
jsnoble: removed

packages/xlucene-evaluator/src/parser/functions/geo/polygon.ts
jsnoble: removed

packages/xlucene-evaluator/src/parser/functions/geo/polygon.ts
jsnoble: removed

packages/xlucene-evaluator/src/parser/functions/geo/polygon.ts
jsnoble: I can remove this since it does not throw type errors

packages/xlucene-evaluator/src/parser/functions/geo/polygon.ts
jsnoble: cant remove this as no types are provided

packages/xlucene-evaluator/src/parser/functions/geo/polygon.ts
jsnoble: fixed

packages/xlucene-evaluator/src/parser/interfaces.ts
jsnoble: fixed

packages/xlucene-evaluator/src/parser/interfaces.ts
```diff
@@ -1,29 +1,18 @@
 
 import { Logger } from '@terascope/utils';
 import { UtilsTranslateQueryOptions, AnyQuery, AnyQuerySort } from '../translator';
```
peterdemartini: This should be more specific `../translator/interfaces`

jsnoble: Why? I am having all the interfaces exported through translator as it was to constricting what was happening before. Essentially for all other module we have, we export though the index.

peterdemartini: Requiring the whole directory, makes the parser a dependent and dependency off the translator. If you require just the translator interfaces, you avoid this. In these confusing scenarios, I usually choose to add comment at the top of the files to make sure that it’s important to avoid being dependent and a certain file.

jsnoble: I may be confused about how typescript interface files work, but I am not understanding this suggestion makes anything less circular dependent? Does somehow changing imports from `src/translator` to `src/translator/interfaces` somehow not not require the `translator`? You are still dependent on translator no matter what. You still have the same issue either way. If we really wanted this to not require types from translator then we should consider moving nearly all the types out of both to the top level interface file

peterdemartini: It does, but typescript and javascript don’t really care about folders, only files, (unless you forward everything the index file like we do). Currently the `translator/interfaces.ts` doesn’t have a dependency on `parser/interfaces.ts` any more but `translator/index` exports more than one file with a dependency on `parser/interfaces.ts`.

`translator/interfaces.ts` does not import another another file from the translator folder so by importing just that file you avoid getting all of the translator files.

PR #1451, opened 2019-09-30
v0.58.0 - Convert terafoundation to typescript

peterdemartini:
Converts `terafoundation` to `typescript`, in order to be able to export typescript interfaces, I had to make a breaking to what gets exported. I decided to take the opportunity to export classes.

**BREAKING CHANGES:**

- `terafoundation` no longer exports a main function, for the cluster context, it now exports `ClusterContext` that use can destructure:
```js
// previously
require('terafoudation')({
    name: 'testserver'
})
// is now
const { ClusterContext } = require('terafoudation');
new ClusterContext({
    name: 'testserver'
})
```
- Removed `master_plugin` and `bootstrap` from the `terafoundation` configuration because they were half-baked (with comments indicated this) and unused (even internally).

No other major features have been added or removed, I doubled checked compatibility and usage in our internal repos.


kstaken:
Are there corresponding PRs for dependent projects related to this breaking change?

peterdemartini:
No PRs for those yet, but I can add them, I will have to create a prerelease version and use that in the other projects

PR #1465, opened 2019-10-04
Quote fix

jsnoble:
quote fix for spaces usage

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,8 +83,7 @@
                     "from": "base",
                     "name": "connectors"
                 }
-            ],
-            "registry": "terascope/teraslice"
+            ]
         },
         "npm": {
             "registry": "https://registry.npmjs.org/"
--- a/package.json
+++ b/package.json
@@ -71,7 +71,9 @@
         "root": true,
         "type": "monorepo",
         "docker": {
-            "registries": ["terascope/teraslice"],
+            "registries": [
+                "terascope/teraslice"
+            ],
             "cache_layers": [
                 {
                     "from": "node:10.16.3-alpine",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -81,7 +83,8 @@
                     "from": "base",
                     "name": "connectors"
                 }
-            ]
+            ],
+            "registry": "terascope/teraslice"
```
peterdemartini: Can you remove this? This has been replaced in favor of `registries` which is already there.

packages/xlucene-evaluator/test/cases/parser/terms.ts
```diff
@@ -3,58 +3,82 @@ import { TestCase } from './interfaces';
...
-            field: null,
             quoted: true,
-            value: 'foo',
+            field: 'id',
+            value: 'some \\"thing\\" else',
         },
     ],
+
     [
-        "'foo'",
```
peterdemartini: Looks like removed the test for a single quoted string (without a field). Was there a good reason for that?

packages/xlucene-evaluator/test/cases/parser/terms.ts
```diff
@@ -3,58 +3,82 @@ import { TestCase } from './interfaces';
...
             field_type: FieldType.String,
             quoted: false,
-            field: null,
-            value: 'foo bar',
+            field: 'id',
+            value: "some'other'thing",
         },
     ],
     [
-        '"foo"',
```
peterdemartini: Looks like removed the test for a double quoted string (without a field). Was there a good reason for that?

PR #1515, opened 2019-10-31
add geopoint and geoshape to datatypes

godber:


godber:
So far this just implements geopoint.  I've added geoshape to the interfaces.

packages/data-types/src/types/versions/v1/geopoint.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export default class GeoPointType extends BaseType {
+    toESMapping(_version?: number) {
+        return { mapping: { [this.field]: { type: 'geo_point' as ElasticSearchTypes } } };
+    }
+
+    // TODO: need notion of injecting custom types, what about duplicates
+    toGraphQL() {
+        const customType = `
+            type GeoPointType {
```
peterdemartini: I think now is a good time to namespace the types (see comment above)

```suggestion
            type DTGeoPointV1 {
```

godber:
refs: #1524

packages/xlucene-evaluator/src/utils.ts
```diff
@@ -169,6 +178,124 @@ export type CreateJoinQueryOptions = {
...
+    const { lat, lon } = results;
+    return `${field}:geoContainsPoint(point:"${lat},${lon}")`;
+}
+
+function coordinateToXlucene(cord: CoordinateTuple) {
+    // tuple is [lon, lat], need to return "lat, lon"
+    return `"${cord[1]}, ${cord[0]}"`;
+}
+
+function wrap(arr: any[]) {
```
peterdemartini: This function name should be more specific. Also is the function properly stringifying the array?

```suggestion
function makeArrayString(arr: any[]) {
```

packages/xlucene-evaluator/src/interfaces.ts
```diff
@@ -31,7 +63,9 @@ export enum FieldType {
     Integer = 'integer',
     Float = 'float',
     Boolean = 'boolean',
-    Object = 'object'
+    Object = 'object',
+    GeoPoint = 'geoPoint',
```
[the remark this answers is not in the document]
jsnoble: in general we typically do not use underscores in value strings. This would equate to `{ location: 'geo_point', ipAddress: 'ip' } ` instead of `{ location: 'geoPoint', ipAddress: 'ip' }`. Counterpoint would be that  we use underscores for configuration field names. Though this is a value and not a field this object is a configuration. Do we have any other places that have underscored values?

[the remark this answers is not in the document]
peterdemartini: We should really write a convention for naming things, but you're right this is a little bit confusing and we don't really use snake case for values that often but we don't use camel case either, it seems kebab case is the most common (even with in [xlucene](https://github.com/terascope/teraslice/blob/master/packages/xlucene-evaluator/src/parser/interfaces.ts#L49-L63)).

IMO this what our convention has been/should be:

- **snake case** for configuration and object properties
- **kebab case** for string values
- **camel case** for variables and function names
- **pascal case** for class names and type definitions

packages/xlucene-evaluator/src/parser/functions/geo/contains-point.ts
```diff
@@ -0,0 +1,54 @@
...
+import { point } from '@turf/helpers';
+import { pointInGeoShape } from './helpers';
+import { parseGeoPoint } from '../../../utils';
+import * as i from '../../interfaces';
+import { AnyQuery, ESGeoShapeType } from '../../../translator/interfaces';
+import { GeoShapeRelation } from '../../../interfaces';
+
+function validate(params: i.Term[]) {
+    const geoPointParam = params.find((node) => node.field === 'point');
+    if (geoPointParam == null) throw new Error('geoDistance query needs to specify a "point" parameter');
```
peterdemartini:
```suggestion
    if (geoPointParam == null) throw new Error('geoContainsPoint query needs to specify a "point" parameter');
```

PR #1525, opened 2019-11-06
elasticsearch-store changes

peterdemartini:
**Features:** 

- Add soft delete support
- Add the ability to add extra constraints to Query Access (so multiple constraints can be added without have to worry about correctly joining the together or potentially mutating a user defined constrained).
- Add support for read/write hooks for a records (useful for validation and fixing records before/after storing them).

**Breaking Changes:**

- Changes to some of the base `IndexModel` fields, `id` => `_key`, `created` => `_created`, and `updated` => `_created`.
- Moves more methods from the `IndexModel` to the `IndexStore` since the `IndexStore` should be more generic and would be benefit from those methods.
- Changes the argument order of some the methods on `IndexModel` to be more consistent (for example, `createWithId(doc, id)` is now `createById(id, doc)`).
- Renames some of the methods on `IndexModel` to have a convention different than the ones on `IndexStore`, now there is `createRecord`, `updateRecord`, `fetchRecord`, and `deleteRecord`
- Removes support for the old read/write hook API

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ jobs:
       name: 'ES Test Suite (elasticsearch 5) (node 10)'
       node_js: '10.16'
       # run only on pull-requests
-      if: branch = master AND type IN (pull_request)
+      if: branch = master AND type IN (pull_request) AND commit_message !~ /^WIP/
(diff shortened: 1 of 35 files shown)


peterdemartini:
This PR should be good-to-go. @kstaken will you review it?

PR #1526, opened 2019-11-08
v0.60.0: teraslice client / CLI and CI fixes/changes

peterdemartini:
- Fixes to teraslice (for CI test failures)
- Fixes to `ts-scripts`
- Bumps some packages
- Make `teraslice-client-js` API more consistent and add the ability to recover an execution (not a job).
- When creating an execution context (job submit, job recover), the `ex_id` will also be returned in the response.
- Adds basic `teraslice-cli ex recover` command.

```js
const { TerasliceClient } = require('teraslice-client-js');

const client = new TerasliceClient({
    host: 'http://localhost:5678'
});

// client.ex no longer exists (it will throw an error if accessed)
// client.executions now provides a similar API to client.jobs

// - you can list executions;
const executions = await client.executions.list();
// - you can list errors;
const errors = await client.executions.errors();

// - If you submit a job via executions.submit, it will create a job but return an instance of executions
const ex = await client.executions.submit({
    name: 'example-job',
    operations: [
        {
            _op: 'test-reader'
        },
        {
            _op: 'noop'
        }
    ]
});

// The returned execution will no allow have the per-execution APIS.
await ex.status() // => running
await ex.waitForStatus('completed');
await ex.errors();
```

### DIFF
--- a/packages/teraslice/lib/workers/execution-controller/scheduler.js
+++ b/packages/teraslice/lib/workers/execution-controller/scheduler.js
@@ -37,17 +37,38 @@ class Scheduler {
         this._processSlicers();
     }
 
+    /**
+     * Initialize the recovery instance or the execution context,
+     * if recovery is initialized, the execution context will not be
+     * initialized until the execution if finished and the cleanup
+     * type is set.
+    */
+    async initialize() {
+        if (this.recoverExecution) {
+            await this._initializeRecovery();
+        } else {
+            await this._initializeExecution();
+        }
+    }
+
+    /**
+     * Run the execution, this will block until complete (or failed)
+    */
     async run() {
         if (this.recoverExecution) {
-            await this._recoverSlices();
+            this.logger.info(`execution: ${this.exId} is starting in recovery mode`);
+            this.ready = true;
+            this.start();
+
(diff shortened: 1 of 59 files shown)


peterdemartini:
We should temporarily hold off on merging this until #1534 is merged.

godber:
Note, this line results in teraslice returning the ex_id` along with the `job_id` when starting a job:

https://github.com/terascope/teraslice/pull/1526/files#diff-d602aa9138808d127feb73b665009113R276

packages/teraslice/lib/workers/execution-controller/scheduler.js
```diff
@@ -379,20 +400,17 @@ class Scheduler {
...
+    }
 
-        this.logger.info(`execution: ${this.exId} is starting in recovery mode`);
-        this.ready = true;
-        this.start();
-
-        await this._waitForRecovery();
-        await this._recoveryComplete();
+    async _initializeExecution() {
+        await this.executionContext.initialize(this.startingPoints);
```
godber: Could we benefit from more logging here?  Like could some of these `await`s take to long, block forever, or fail?  It might be good to know what the last thing that happened was.

Issue #1547, opened 2019-11-21
Ambiguous slice error messages


peterdemartini:
While debugging a issue with teraslice, I found there only vaque error messages when a slice failed.

```sh
 worker: 127.0.0.1__xxxxx has failure completing its slice {
       slice:
       {
         slice_id: '7384cbc9-0d4d-4b54-beef-b7b78e805dab',
         slicer_id: 1,
         slicer_order: 5000,
         request: { ... }
      },
      analytics: { ... },
      error: '{}'
}
```

The `error` shouldn't be set to `{}`.

PR #1557, opened 2019-11-25
v0.60.2 Better slice error messages

peterdemartini:
Ensures more detailed slice errors and that more error teraslice errors are logged, or at least a stack trace to find the problem.

- Fixes script processor exit code (it checked for `code > 1` not `code > 0` for child process errors)
- Adds `logError` helper to utils
- Other misc logging/error handling improvements.

Resolves https://github.com/terascope/teraslice/issues/1547

### DIFF
--- a/packages/teraslice/lib/cluster/storage/state.js
+++ b/packages/teraslice/lib/cluster/storage/state.js
@@ -79,7 +79,8 @@ module.exports = async function stateStorage(context) {
             state
         };
 
-        if (state === 'error') {
+        // it will usaully just be error
+        if (state === 'error' || error) {
             if (error) {
                 record.error = getFullErrorStack(error);
             } else {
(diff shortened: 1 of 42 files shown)


packages/teraslice/lib/cluster/storage/state.js
```diff
@@ -78,8 +79,12 @@ module.exports = async function stateStorage(context) {
             state
         };
 
-        if (error) {
-            record.error = toString(error);
+        if (state === 'error') {
+            if (error) {
```
godber: Are there cases where `error` evaluates `true` but the `state != 'error'` that need handling?

peterdemartini: Currently, no but I added an additional check to make we don’t lose the error if the status is not to `error`.

PR #1559, opened 2019-11-26
Test harness recovery fixes

jsnoble:
- better typing for recovery data in slicer and test-harness
- slicer-core and testHarness will throw if recoveryData is given but isRecoverable return false;
- added slicer recovery example in tests


packages/job-components/src/operations/core/slicer-core.ts
```diff
@@ -58,6 +58,9 @@ export default abstract class SlicerCore<T = OpConfig>
      * @param recoveryData is the data to recover from
      */
     async initialize(recoveryData: SlicerRecoveryData[]): Promise<void> {
+        if (recoveryData?.length > 0 && !this.isRecoverable()) {
+            throw new Error('cannot provide recovery data to a slicer that is not recoverable. Please create the isRecoverable method and have it return true if recovery is desired');
```
peterdemartini: Errors should start with a capital letter, (especially when the second sentence that starts with a capital letter). While this error message would make more sense within the test harness, it may be confusing when running the job since the person using the reader isn't given it "recovery data", it is just recovering the slicer.

```suggestion
            throw new Error('Slicer is not recoverable, please create the isRecoverable method and return true to enable recovery');
```

We should throw this error in the constructor too so it will fail before it actually starts recovering the previous slices. This can be achieved by using the `recovered_execution` boolean on the [execution config](https://github.com/terascope/teraslice/blob/d09841be3ac4d2dd33ce02a3c447e1a7a39f634d/packages/job-components/src/interfaces/jobs.ts#L105).

packages/teraslice-test-harness/src/slicer-test-harness.ts
```diff
@@ -47,10 +49,15 @@ export default class SlicerTestHarness extends BaseTestHarness<SlicerExecutionCo
...
      * @param recoveryData is an array of starting points to recover from
     */
-    async initialize(recoveryData?: object[]) {
+    async initialize(recoveryData: SlicerRecoveryData[] = []) {
         await super.initialize();
-        await this.executionContext.initialize(recoveryData);
+        // teraslice checks to see if slicer is recoverable
+        // should throw test recoveryData if slicer is not recoverable
+        if (recoveryData.length > 0 && !this.executionContext.slicer().isRecoverable()) {
+            throw new TSError('cannot provide test recovery data if slicer is not set to recoverable, please add the isRecoverable method to the slicer and return true');
```
peterdemartini: Shouldn't we just like the slicer core library throw this error?

peterdemartini: But maybe it would be helpful to validate the `recoveryData`, (like making sure that its an array of objects with `lastSlice`) instead?

packages/teraslice-test-harness/test/fixtures/asset/recoverable-reader.ts/fetcher.ts
```diff
@@ -0,0 +1,14 @@
+
+import { Fetcher, SliceRequest, AnyObject } from '@terascope/job-components';
+
+export default class TestFetcher extends Fetcher<AnyObject> {
+    // @ts-ignore
+    constructor(...args) {
```
peterdemartini: This is a useless constructor, you can just remove this this.

packages/teraslice-test-harness/test/fixtures/asset/recoverable-reader.ts/slicer.ts
```diff
@@ -0,0 +1,28 @@
+import { Slicer, AnyObject, SlicerRecoveryData } from '@terascope/job-components';
+
+export default class Counter extends Slicer<AnyObject> {
+    count = 0;
+    // @ts-ignore
+    constructor(...args) {
```
peterdemartini: This is a useless constructor, you can just remove this this.

packages/job-components/src/operations/core/slicer-core.ts
jsnoble: yeah, moving it to the constructor is a better idea

packages/teraslice-test-harness/src/slicer-test-harness.ts
jsnoble: I added recoveryData validation but I opted to keep the error check here. The reason why is because the recover_exid is normally added internally by the teraslice system. Its not something commonly known, nor should others know about the internals so much, especially the user creating the test for the slicer. So in the most likely case of the user not creating an executionConfig with the recovery_ex-id string, I am validating it here. Plus there is no good way to add it in by the time we are at initialize since the constructor already did a lot of setup.

peterdemartini: I think your solution works well.

PR #1571, opened 2019-12-06
ESlint Fixes / Improvements

peterdemartini:
- fix inconsistencies in with line spaces in typescript and javascript files.
- enable `@typescript-eslint/brace-style` for consistency between js and ts
- enable `@typescript-eslint/no-extra-parens` for consistency between js and ts
- enable `@typescript-eslint/func-call-spacing` for consistency between js and ts
- enforce more naming conventions for typescript
- add [eslint-plugin-jest](https://www.npmjs.com/package/eslint-plugin-jest) support to catch things like:
   - Invalid `expect` / `it` / `describe` 
   - Duplicate test titles (to avoid auto-grouping them)
   - Committed disabled or focused tests
   - Using `jasmine` globals

These linting changes have already caught a variety of problems (mainly in the tests)

**NOTE:** Ignore the file changed file count (it just fixes the multiple empty lines)

### DIFF
--- a/packages/data-types/src/types/versions/v1/object.ts
+++ b/packages/data-types/src/types/versions/v1/object.ts
@@ -1,4 +1,3 @@
-
 import { FieldType } from 'xlucene-evaluator';
 import BaseType from '../base-type';
 import { ElasticSearchTypes } from '../../../interfaces';
--- a/packages/job-components/test/job-validator-spec.ts
+++ b/packages/job-components/test/job-validator-spec.ts
@@ -38,25 +38,6 @@ describe('JobValidator', () => {
             expect(validJob.assets).toBeDefined();
         });
 
-        it('throws an error with faulty operation configuration', () => {
-            const jobSpec: JobConfig = {
-                name: 'test',
-                // @ts-ignore
-                operations: [
-                    {
-                        something: 'else',
-                    },
-                    {
-                        _op: 'noop',
-                    },
-                ],
-            };
-
-            expect(() => {
-                api.validateConfig(jobSpec);
-            }).toThrowError();
-        });
-
         it('will properly read an operation', () => {
             const jobSpec: JobConfig = {
                 name: 'test',
@@ -91,27 +72,6 @@ describe('JobValidator', () => {
                         _op: 'noop',
                     },
                 ],
-            };
-
-            expect(() => {
-                api.validateConfig(jobSpec);
-            }).toThrowError();
-        });
-
-        it('will throw based off crossValidation errors', () => {
-            const jobSpec: JobConfig = {
-                name: 'test',
-                lifecycle: 'persistent',
-                assets: ['fixtures'],
-                operations: [
-                    {
-                        _op: 'example-reader'
-                    },
-                    {
-                        _op: 'noop',
-                        failCrossValidation: true
-                    },
-                ]
             };
 
             expect(() => {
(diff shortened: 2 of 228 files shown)


packages/data-types/src/types/versions/v1/object.ts
```diff
@@ -1,4 +1,3 @@
-
```
jsnoble: I know that consistency is important and we either need to do things one way or the other. That being said I think its better that we have a space at the beginning of the file just like we do at the end of a file. It gives the code more room and it's not crammed at the top. I think it makes it easier to scan. What are your thoughts

peterdemartini: Empty lines at the top of files is not common practice, I am not aware of any project that does it and I personally think it makes things look worse than better (by appearance anyways).

Line endings at the end-of-file are significant because "Trailing newlines in non-empty files are a common UNIX idiom. Benefits of trailing newlines include the ability to concatenate or append to files as well as output files to the terminal without interfering with shell prompts." [see eol-last rule docs](https://eslint.org/docs/rules/eol-last). Adding the line at the top has been a recent thing you've been doing that I think broken the consistency.

Regarding scanning things: I never care what is at the top and when there is a starting empty line and scroll up to the top of time, I am always wondering if there is more above that any my editor is just not showing it.

**TLDR:** It is not really common practice, it breaks away from the convention in 90% of our files, and eof line endings are used for a separate [reason](https://eslint.org/docs/rules/eol-last).

packages/terafoundation/test/sysconfig-spec.ts
```diff
@@ -23,11 +23,6 @@ describe('sysconfig helpers', () => {
             const filePath = getTestFile('test-config.yaml');
             expect(() => parseConfigFile(filePath)).not.toThrow();
         });
-
-        it('should work with a yaml file', () => {
-            const filePath = getTestFile('test-config.yml');
```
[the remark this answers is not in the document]
peterdemartini: This test was removed because there is already an identical test for this (with the same title)

packages/job-components/test/job-validator-spec.ts
```diff
@@ -38,44 +38,6 @@ describe('JobValidator', () => {
             expect(validJob.assets).toBeDefined();
         });
 
-        it('throws an error with faulty operation configuration', () => {
-            const jobSpec: JobConfig = {
-                name: 'test',
```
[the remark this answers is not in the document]
peterdemartini: This test was removed because there is already an identical test for this (with the same title)

packages/data-types/src/types/versions/v1/object.ts
jsnoble: I guess we will have to agree to disagree on what looks worse on it then. I checked a few projects and while it is a bit more common to have no spaces at the top they do have files with spaces as well. As for the line ending comment, I already know there are important file reasons beyond aesthetics. As for the comment about this happening recently is not true. I have been added spaces to files for a couple of years now unless the vast majority of it already does not. Most of my older stuff has already been rewritten by you to this format. Even still, despite the the changes over time there being over 200 file changes to fix this clearly shows that a good part used to have spaces so its not breaking convention in 90% of our files. This is not a hill for me to stand on, its not really that big a deal. Consistency is the most important. Most of these linting changes make sense, it just want to advise you to be careful about the desire to change linting styles arbitrarily since Kimbro wants to stay out of it.

Issue #1546, opened 2019-11-21
K8s jobs should have a labels property that gets copied to k8s resource labels


godber:
Teraslice job authors should be able to specify a job option called `labels` that is a list of key value pairs that are simply copied to labels on k8s resources (or any other clustering service).

godber:
Note to the person implementing this, if we chose to do so, don't forget to sanitize the values so they are valid for k8s.

godber:
One possible use case for these labels would be to indicate teraslice jobs that have workers that maintain state and hence care needs to be taken when scaling or restarting them.

godber:
Current details on valid k8s labels and values for k8s v1.17:

https://v1-17.docs.kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set


Note: "Automated system components (e.g. kube-scheduler, kube-controller-manager, kube-apiserver, kubectl, or other third-party automation) which add labels to end-user objects must specify a prefix."

So right now, the other places where we add labels (e.g. `jobId`, `exId`, `jobName`) we are not compliant.  This is something we should fix.

> Labels are key/value pairs. Valid label keys have two segments: an optional prefix and name, separated by a slash (/). The name segment is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between. The prefix is optional. If specified, the prefix must be a DNS subdomain: a series of DNS labels separated by dots (.), not longer than 253 characters in total, followed by a slash (/).

> Valid label values must be 63 characters or less and must be empty or begin and end with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.

I won't fix the prefix elsewhere, but when adding these new labels I will include the prefix.  Or, for this use case, actually, perhaps the labels ARE user labels and don't need to be prefixed.

The prefix we would use would probably be `k8s.terascope.io/`

godber:
Actually we should use prefixes as defined here: https://github.com/terascope/teraslice/issues/1583

godber:
@kstaken do you have any concerns with me implementing this feature?  If not, do you have opinions on how you'd like these labels to appear on the jobs?

```json
{
    "labels1": [
        ["labelKey1", "value1"],
        ["labelKey2", "value2"]
    ],

    "labels2": {
        "labelKey1": "value1",
        "labelKey2": "value2"
    }
}
```

[...]

PR #1592, opened 2019-12-16
K8s winter 2019 b1

godber:
This MR addresses the following issues:

* #1546

### DIFF
--- a/docs/configuration/clustering-k8s.md
+++ b/docs/configuration/clustering-k8s.md
@@ -103,6 +103,21 @@ Note that the `assets_volume` should also be mounted to your Teraslice master po
 Support for Kubernetes based clustering adds additional properties to a
 Teraslice job definition.  These are outlined below.
 
+### Labels
+
+Key value pairs added into a job's `labels` array, as shown below, will result
+in labels being added to the k8s resources.  The k8s labels will be prefixed
+with `job.teraslice.terascope.io/`.
+
+```json
+    "labels": [
+        ["key1", "value1"]
+    ],
+```
+
+If `labels` is omitted, the k8s resources will just have the standard set of
+labels that Teraslice uses.
+
 ### Resources
 
 It is possible to set CPU and memory resource constraints for your Teraslice
--- a/docs/configuration/clustering-k8s.md
+++ b/docs/configuration/clustering-k8s.md
@@ -115,9 +115,9 @@ in labels being added to the k8s resources.  The k8s labels will be prefixed
 with `job.teraslice.terascope.io/`.
 
 ```json
-    "labels": [
-        ["key1", "value1"]
-    ],
+    "labels": {
+        "key1": "value1"
+    },
 ```
 
 If `labels` is omitted, the k8s resources will just have the standard set of
--- a/docs/configuration/clustering-k8s.md
+++ b/docs/configuration/clustering-k8s.md
@@ -91,12 +91,17 @@ to support k8s based Teraslice deployments.
 |        Configuration         |                                                                        Description                                                                         |  Type  |  Notes   |
 | :--------------------------: | :--------------------------------------------------------------------------------------------------------------------------------------------------------: | :----: | :------: |
 |        assets_volume         |                               Name of kubernetes volume to be shared across all pods, where Teraslice assets will be stored                                | String | optional |
+| execution_controller_targets |                                 array of `{"key": "rack", "value": "alpha"}` targets for execution controllers                                             | String | optional |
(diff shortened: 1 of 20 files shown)


packages/job-components/src/job-schemas.ts
```diff
@@ -45,6 +45,17 @@ export function jobSchema(context: Context): convict.Schema<any> {
                 }
             },
         },
+        labels: {
+            default: null,
+            doc: 'An array of arrays containing key value pairs used to label kubetnetes resources.',
+            format(arr: any) {
+                if (arr != null) {
+                    if (!Array.isArray(arr)) {
+                        throw new Error('labels is required to be an array');
```
peterdemartini: We should validate that the array of is an array of tuple strings.

packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/k8sResource.js
```diff
@@ -122,6 +125,23 @@ class K8sResource {
         }
     }
 
+    _setJobLabels() {
+        if (this.execution.labels != null) {
+            _.forEach(this.execution.labels, (label) => {
+                const keyName = label[0].replace(/[^a-zA-Z0-9\-._]/g, '-').substring(0, 63);
```
peterdemartini: If `label[0]` is not a string this will blow up, I'd imagine we want to gracefully handle that (by either logging or throwing a more readable error). Same with `label[1]`

docs/configuration/clustering-k8s.md
```diff
@@ -91,18 +91,38 @@ to support k8s based Teraslice deployments.
...
+
+Key value pairs added into a job's `labels` array, as shown below, will result
+in labels being added to the k8s resources.  The k8s labels will be prefixed
+with `job.teraslice.terascope.io/`.
+
+```json
+    "labels": [
+        ["key1", "value1"]
+    ],
+```
```
godber: I might change this to `[("key": "key1", "value": "value1"}]` since it's more explicit and that is what is used for targets.

packages/job-components/src/job-schemas.ts
godber: Yeah, I was deliberately cutting corners on the validations for now.

Will you take an IOU :)

packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/k8sResource.js
godber: Might make this one a FIXME too.  Depending on where you are with your PR.

peterdemartini: That is fine

docs/configuration/clustering-k8s.md
peterdemartini: Another simple syntax would be `["key1=value1", "key2=value2"]` (similar to [docker labels](https://docs.docker.com/config/labels-custom-metadata/)). A regex can be used to extract the key values `/^\s*([\w.-]+)\s*=\s*["']{0,1}([^"'\s]*)?["']{0,1}\s*$/`

packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/k8sResource.js
```diff
@@ -122,6 +145,22 @@ class K8sResource {
         }
     }
 
+    _setJobLabels() {
+        if (this.execution.labels != null) {
+            Object.entries(this.execution.labels).forEach(([k, v]) => {
+                const key = `${this.jobLabelPrefix}/${k.toString().replace(/[^a-zA-Z0-9\-._]/g, '-').substring(0, 63)}`;
+                const value = v.toString().replace(/[^a-zA-Z0-9\-._]/g, '-').substring(0, 63);
```
godber: @peterdemartini can you think of a `k` or `v` value here that I can put in tests to make those fail?  Asked another way, should I try to at least do a little error handling here, maybe like this:

```javascript
                try {
                    const key = `${this.jobLabelPrefix}/${k.toString().replace(/[^a-zA-Z0-9\-._]/g, '-').substring(0, 63)}`;
                } catch {
                    throw new Error(`Unable to generate k8s label key from: ${k}`);
                }
                try {
                    const value = v.toString().replace(/[^a-zA-Z0-9\-._]/g, '-').substring(0, 63);
                } catch {
                    throw new Error(`Unable to generate k8s label value from: ${v}`);
                }
```

Or should I just let the (future) validation save me.

PR #1593, opened 2019-12-20
Xlucene variables

jsnoble:
all tests pass both here and in spaces. The parsing time is much quicker but there are some more minor improvements that could be made since there seems to be a difference in parsing time between a polygon and a multipolygon even though they should be pretty similar. 

- [x]  add a few more tests to elasticsearch-store
- [x] need to bump packages


packages/elasticsearch-store/src/index-store.ts
```diff
@@ -373,8 +374,15 @@ export default class IndexStore<T extends Record<string, any>> {
...
     }
 
-    async countBy(fields: AnyInput<T>, joinBy?: JoinBy, arrayJoinBy?: JoinBy): Promise<number> {
-        return this.count(this.createJoinQuery(fields, joinBy, arrayJoinBy));
+    async countBy(
+        fields: AnyInput<T>,
+        joinBy?: JoinBy,
+        options?: RestrictOptions
+    ): Promise<number> {
+        console.log('incoming', fields, options?.variables)
```
peterdemartini:
```suggestion
```

packages/elasticsearch-store/src/index-store.ts
```diff
@@ -373,8 +374,15 @@ export default class IndexStore<T extends Record<string, any>> {
...
+        const { query, variables } = this.createJoinQuery(fields, joinBy, options?.variables);
+        console.log('what is query', query, variables)
```
peterdemartini:
```suggestion
```

packages/xlucene-evaluator/peg/lucene.pegjs
```diff
@@ -600,6 +635,9 @@ WildcardCharSet "wildcard"
 FieldChar "field"
   = [_a-zA-Z0-9-\.\?\*]
 
+VariableChar
+  = [_a-zA-Z0-9-\.\?\*]
```
peterdemartini: Variables cannot include some of these characters:

```suggestion
  = [_a-zA-Z0-9]
```

packages/xlucene-evaluator/test/cases/translator/geo.ts
```diff
@@ -115,22 +115,11 @@ export default [
...
-                                    },
-                                    {
-                                        lat: 85.3,
-                                        lon: 134.4
-                                    },
-                                    {
-                                        lat: 89.3,
-                                        lon: 154.4
-                                    }
+                                    [140.43, 70.43],
```
peterdemartini: Do we still support the old format?

packages/xlucene-evaluator/src/utils.ts
```diff
@@ -165,127 +170,130 @@ export function parseGeoPoint(point: GeoPointInput): GeoPoint;
...
 export function coordinateToXlucene(cord: CoordinateTuple) {
     // tuple is [lon, lat], need to return "lat, lon"
     return `"${cord[1]}, ${cord[0]}"`;
 }
 
-function makeArrayString(arr: any[]) {
-    return `[${arr}]`;
+function geoPolyQuery(field: string, polyVariableName: string, vParam?: string) {
+    if (vParam) {
+        return `${field}:geoPolygon(points: ${polyVariableName} relation: ${vParam})`;
```
peterdemartini:
```suggestion
        return `${field}:geoPolygon(points: ${polyVariableName}, relation: ${vParam})`;
```

packages/xlucene-evaluator/src/utils.ts
```diff
@@ -302,42 +310,76 @@ function isGeoJSONType(type: FieldType) {
...
+
+    constructor(variables?: AnyObject) {
+        this.variables = variables || {};
+    }
+
+    private _makeKey(field: string) {
+        let num = 1;
+        let name = `${field}_${num}`;
+
+        while (this.variables[name]) {
```
peterdemartini:
```suggestion
        while (this.variables[name] !== undefined) {
```

packages/xlucene-evaluator/src/utils.ts
```diff
@@ -302,42 +310,76 @@ function isGeoJSONType(type: FieldType) {
...
+            if (this.variables[vField] === undefined) throw new Error(`must provide variable "${vField}"`);
+            return value;
+        }
+        const key = this._makeKey(field);
+        this.variables[key] = value;
+        return `$${key}`;
+    }
+
+    getVaraibles() {
+        return this.variables;
```
peterdemartini:
```suggestion
        return { ...this.variables };
```

packages/xlucene-evaluator/test/cases/translator/geo.ts
jsnoble: yes, we have to for now

packages/xlucene-evaluator/src/utils.ts
jsnoble: comma are not needed, and we need to test to make sure that it works without them, but stylistically I can put it in

peterdemartini: Sorry, I should have been more clear. I was thinking for debugging purposes it would be easier to read.

packages/xlucene-evaluator/src/parser/context.ts
```diff
@@ -81,10 +89,30 @@ export default function makeContext(args: any) {
...
+                type: i.ASTType.Conjunction,
+                nodes: [node]
+            };
+        });
+    }
+    // cannot allow variables that are objects, errors, maps, sets, buffers
+    function validateRestrictedVariable(val: any) {
+        const type = typeof val;
+        if (!['string', 'number', 'boolean'].includes(type)) {
+            throw new Error(' non-function variables may only be numbers, strings, boolean or an array of these primitives');
```
peterdemartini: This error message isn't very human friendly. I think something like `Unsupported type of Buffer  received for variable $foo` instead of ` non-function variables may only be numbers, strings, boolean or an array of these primitives`

```suggestion
            throw new Error('Unsupported type of ${getTypeOf(val)} received for variable $${variable}');
```

Issue #954, opened 2019-01-16
New command `await` for job status


kstaken:
We should add a new command that works in both `tjm` context and general jobs context to await on a running job reaching a particular status or any of a set of statuses. Should probably have a configurable timeout that by default is `unlimited` since jobs can run for long periods.

Something like

```
earl tjm await job.json completed
earl tjm await job.json completed,stopped
earl jobs await JOB_ID completed -t 600
```

peterdemartini:
I like it. I recommend making the statuses configured via a flag instead of a positional argumentent so it’s easier to read and allows yargs to deal with the array parsing.

kstaken:
Yeah that makes sense

ciorg:
command would be:
earl jobs await CLUSTER JOB_ID --status completed,stopped --timeout 10min

earl tjm JOB.JSON --status completed,stopped
  
tjm and jobs commands _should_ share code because they do many of the same things.  Would be easier for tjm to use the jobs helper class for all job-related actions, stop, start, etc... than for jobs to use tjm helper class.

Plan is to add the await function to the jobs class and tjm will import that class.

godber:
I have confirmed this is functional:

```
earl tjm await --start example-job.json
Started example-data-generator-job on http://192.168.64.5:30678
> job: fd8af957-743a-41a6-88d0-8e0376dbef3a current status: initializing
> job: fd8af957-743a-41a6-88d0-8e0376dbef3a waiting for status completed or stopped
> job: fd8af957-743a-41a6-88d0-8e0376dbef3a reached status: completed
```

Thanks Charlie!

PR #1608, opened 2020-01-07
teraslice-cli await command for tjm and jobs

ciorg:
* await command blocks until status is reached for both tjm and jobs
* added stopped to terminal statuses in waitForStatus function in teraslice-client-js 
* jobs await parsing test
* added documentation
* other small fixes in docs


packages/teraslice-cli/src/cmds/jobs/await.ts
```diff
@@ -0,0 +1,62 @@
...
+        try {
+            newStatus = await util.pRace(desiredStatus.map(
+                (status) => jobs.awaitStatus(
+                    status,
+                    jobs.config.args.id,
+                    jobs.config.args.timeout
+                )
+            ));
+        } catch (e) {
+            // @ts-ignore
```
peterdemartini: Why is there a `@ts-ignore` here?

packages/teraslice-cli/src/cmds/tjm/await.ts
```diff
@@ -0,0 +1,71 @@
...
+        reply.green(`> job: ${jobFile.jobId} waiting for status ${desiredStatus.join(' or ')}`);
+
+        let newStatus;
+
+        try {
+            newStatus = await util.pRace(desiredStatus.map(
+                (status) => jobFuncs.waitForStatus(status, 5000, argv.timeout)
+            ));
+        } catch (e) {
+            // @ts-ignore
```
peterdemartini: Why is there a `@ts-ignore` here?

packages/teraslice-cli/src/cmds/jobs/await.ts
ciorg: I get 
```
[ts] Property 'includes' does not exist on type 'ExecutionStatus[]
```

not sure why it doesn't think includes is an array property.

ciorg:
addresses issue #954

Issue #1633, opened 2020-01-18
teraslice-cli assets deploy does not pull prerelease if explicitly requested


briend:
```
teraslice-cli assets deploy myTS terascope/file-assets@0.11.3
Unable to download terascope/file-assets@0.11.3 asset: Error: Error downloading terascope/file-assets@0.11.3: Error: could not find a release for terascope/file-assets (darwin x64)
    at GithubAsset.download (/usr/local/lib/node_modules/teraslice-cli/dist/src/helpers/github-asset.js:49:19)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

I believe it's intentionally skipping drafts and prereleases:

https://github.com/terascope/teraslice/blob/master/packages/teraslice-cli/src/helpers/github-asset.ts#L38-L39

Maybe a force option to override this makes sense?

godber:
I think the spot you found is a good place to start.  I don't think we want a force, I think what I described internally is the desired behavior:

```
Here's how it should work:

earl assets deploy ts-test1 terascope/file-assets

does not specify a version, so the question is, what version should it provide you?  Github has a "release" mechanism that allows you to answer that question, by indicating the "Latest Release" here:

https://github.com/terascope/file-assets/releases

That is the "default" or latest stable release.

If a user explicitly specifies a version of an asset, the user should get that version, regardless of its status on the releases page.
```

I suspect we just didn't think through this case when implementing that helper.  Though perhaps I'm wrong about the desired behavior.

godber:
We should just remove that the draft and prerelease constraints in the case that version is specified.

godber:
The latest version in NPM:

```
npm install -g -u teraslice-cli
...
+ teraslice-cli@0.14.1
updated 3 packages in 22.651s
```

 now has this fix:

```
# specifying an unreleased version works
earl assets deploy ts-minikube-dev1 terascope/elasticsearch-assets@v1.7.1
Downloading terascope/elasticsearch-assets@v1.7.1...
elasticsearch-v1.7.1-... Asset posted to ts-minikube-dev1: 599c89138094fec15d53459abbc0d52c3f804784

# default still gives the latest release
earl assets deploy ts-minikube-dev1 terascope/elasticsearch-assets
Downloading terascope/elasticsearch-assets@v1.6.2...

Asset posted to ts-minikube-dev1: 0813c035327bcb35d505bd03ff032fb75c80d279
```

Thanks Jared.

PR #1639, opened 2020-01-21
Cli fix

jsnoble:
- fixes bad logging when changing workers on job
- fixes ex/ to better list all when no status is given


godber:
@jsnoble anyc chance of you looking at #1633 for this PR too?

packages/teraslice-client-js/src/executions.ts
```diff
@@ -73,7 +73,7 @@ export default class Executions extends Client {
 
 function _parseListOptions(options: ListOptions): SearchQuery {
     // support legacy
-    if (!options) return { status: '*' };
+    if (!options) return {};
```
peterdemartini: We should also handle the case when options is `*` too, which should return an object empty object.

Issue #1527, opened 2019-11-08
tjm/ earl `add workers` to job prints useless message


ciorg:
using earl or tjm to add workers to a job prints out:

`[object Object] for JOB_NAME on CLUSTER_URL`

where JOBNAME / CLUSTER_URL are the real job name and cluster url

the command should be `sucessfully added n workers to job or job_id on cluster url` or something like that.

obviously replace job/ job_id and cluster url with real info

PR #1653, opened 2020-01-23
remove --start from await command, addresses issue #1527 when adding workers

ciorg:
* removed --start flag from await command
* fixed bug #1527 
* removed duplicate await and workes code for jobs and tjm commands
* expanded tests slightly
* bumped version


packages/teraslice-cli/src/cmds/jobs/await.ts
```diff
@@ -15,47 +14,30 @@ const cmd: CMD = {
...
+        reply.info(`> job: ${jobs.config.args.id} waiting for status ${desiredStatus.join(' or ')}`);
+
+        const newStatus = await jobs.awaitManyStatuses(
+            desiredStatus,
+            jobs.config.args.id,
+            jobs.config.args.timeout
+        );
+
+        reply.info(`> job: ${jobs.config.args.id} reached status: ${newStatus}`);
+        process.exit(0);
```
jsnoble: you probably should not process.exit() here. We don't do this for any command, and this command might be piped to another command, and we don't want to exit

packages/teraslice-cli/src/cmds/tjm/await.ts
```diff
@@ -16,55 +15,30 @@ const cmd: CMD = {
...
-                newStatus = e.context.lastStatus;
-            } else {
-                reply.fatal(e.message);
-            }
-        }
+        const status = await jobs.awaitManyStatuses(desiredStatus, jobFile.id, argv.timeout);
 
-        reply.green(`> job: ${jobFile.jobId} reached status: ${newStatus}`);
+        reply.green(`> job: ${jobFile.id} reached status: ${status}`);
+        process.exit(0);
```
jsnoble: this should be removed

packages/teraslice-cli/src/cmds/jobs/await.ts
ciorg: with out the process.exit(0) the process hangs, can still use && to chain commands together.  I think it's an artifact of the pRace

peterdemartini: Instead of using pRace this await for multiple status logic could exist in the teraslice-client-js `waitForStatus`, so there interface would be `waitForStatus(target: ExecutionStatus[]| ExecutionStatus)` that way you don't have to two parallel requests and deal with a left over promise when one finishes.

ciorg: that seems like a fairly simple change in the client and would simplify the await commands.

I'll make that change

ciorg:
this is ready for review again

PR #1654, opened 2020-01-24
Data-Library

jsnoble:


packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -0,0 +1,476 @@
...
+    ReplaceRegexConfig
+} from './interfaces';
+import {
+    isString,
+    isValidDate,
+    isNumber
+} from '../validations/field-validator';
+import { Repository } from '../interfaces';
+
+export const respoitory: Repository = {
```
peterdemartini:
```suggestion
export const repository: Repository = {
```

packages/types/src/xlucene-interfaces.ts
```diff
@@ -0,0 +1,21 @@
+export enum XluceneFieldType {
```
peterdemartini: It is xLucene not Xlucene

```suggestion
export enum xLuceneFieldType {
```

packages/types/src/xlucene-interfaces.ts
```diff
@@ -0,0 +1,21 @@
...
+    Integer = 'integer',
+    Float = 'float',
+    Boolean = 'boolean',
+    Object = 'object',
+    GeoPoint = 'geo-point',
+    GeoJSON = 'geo-json',
+    Number = 'number'
+}
+
+export interface XluceneTypeConfig {
```
peterdemartini: It is xLucene not Xlucene

```suggestion
export interface xLuceneTypeConfig {
```

packages/types/src/xlucene-interfaces.ts
```diff
@@ -0,0 +1,21 @@
...
+    [field: string]: XluceneFieldType;
+}
+
+export interface XluceneVariables {
```
peterdemartini:
```suggestion
export interface xLuceneVariables {
```

packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -0,0 +1,476 @@
...
+        result = ts.toNumber(toBoolean(result));
+    }
+
+    result = ts.toNumber(result);
+
+    if (isNaN(result)) throw new Error('could not convert to a number');
+    return result;
+}
+
+export function decodeBase64(input: string) {
```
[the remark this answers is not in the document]
jsnoble: the browser does not have the Buffer class. I believe we decided that we wanted utils to be able to work for the browser.

packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -0,0 +1,476 @@
...
+    if (!isString(input)) throw new Error('Input must be a string');
+    return Buffer.from(input, 'base64').toString('utf8');
+}
+
+export function encodeBase64(input: string) {
```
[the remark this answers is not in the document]
jsnoble: buffer does not work in browser

packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -0,0 +1,476 @@
...
+export function encodeMD5(input: string) {
+    if (!isString(input)) throw new Error('Input must be a string');
+    return crypto.createHash('md5').update(input).digest('hex');
+}
+
+// TODO: better types for this
+export function encodeSHA(input: any, { hash = 'sha256', digest = 'hex' }) {
+    if (!isString(input)) throw new Error('Input must be a string');
+    // TODO: guard for hash ??
+    if (!['latin1', 'hex', 'base64'].includes(digest)) throw new Error('Parameter digest is misconfigured');
```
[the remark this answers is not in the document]
jsnoble: this is what should up in node types, however after checking deeper there are others that work

packages/data-mate/src/transforms/helpers.ts
```diff
@@ -0,0 +1,212 @@
...
+    xLuceneVariables,
+    xLuceneFieldType,
+    GeoShapeRelation
+} from '@terascope/types';
+import { isGeoShapePoint } from '../validations/field-validator';
+
+// TODO: move these
+export type JoinBy = 'AND'|'OR';
+
+export interface XluceneQueryResult {
```
peterdemartini:
```suggestion
export interface xLuceneQueryResult {
```

packages/data-types/src/types/v1/number.ts
```diff
@@ -0,0 +1,17 @@
...
+import BaseType from '../base-type';
+import { ElasticSearchTypes } from '../../interfaces';
+
+export default class NumberClass extends BaseType {
+    toESMapping(_version?: number) {
+        return { mapping: { [this.field]: { type: 'double' as ElasticSearchTypes } } };
+    }
+
+    toGraphQL() {
+        return this._formatGql('Int');
```
peterdemartini: This needs to be float in order to avoid have a 32 bit integer size and to support floats.

```suggestion
        return this._formatGql('Float');
```

packages/data-mate/src/interfaces.ts
```diff
@@ -0,0 +1,14 @@
+import { FieldTypeConfig } from '@terascope/data-types';
+
+export type ArgSchema = Config & { describe?: string };
```
peterdemartini: Should we be consistent with the descriptions?

```suggestion
export type ArgSchema = Config & { description?: string };
```

packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -0,0 +1,448 @@
...
+export function toNumber(input: any, args?: { booleanLike?: boolean }) {
+    let result = input;
+
+    if (args?.booleanLike && ts.isBooleanLike(input)) {
+        result = ts.toNumber(toBoolean(result));
+    }
+
+    result = ts.toNumber(result);
+
+    if (isNaN(result)) throw new Error('could not convert to a number');
```
peterdemartini: We should `Number.isNaN`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN

```suggestion
    if (Number.isNaN(result)) throw new Error('could not convert to a number');
```

packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -0,0 +1,448 @@
...
+    return input.split(delimiter);
+}
+
+// option to specify, seconds, millisecond, microseconds?
+export function toUnixTime(input: any): number {
+    if (!isValidDate(input)) {
+        throw new Error('Not a valid date, cannot transform to unix time');
+    }
+
+    const parsed = isNaN(input) ? Date.parse(input) : Number(input);
```
peterdemartini: We should Number.isNaN, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN

```suggestion
    const parsed = Number.isNaN(input) ? Date.parse(input) : Number(input);
```

Issue #1649, opened 2020-01-22
Ensure that execution controllers are started up before starting workers


godber:
We have been seeing cases where Teraslice workers appear to take two minutes to start processing records while other workers have already started doing their work.  This is apparent if you can look at the number of records processed per worker, over time.

If you enable DEBUG logging on the cluster and look at worker logs, you will see the following in the workers that don't start processing immediately:

```
Error: xhr poll error when connecting to ExecutionController at ...
```

We think the cause of this is that some of the workers start up too quickly, before the execution controller is up and accessible ... thus causing their initial requests to timeout.  Workers that come up slightly later, manage to contact the execution controller and begin work right away.

You can see the potential for this here:

https://github.com/terascope/teraslice/blob/b33b821cb9ab6c575972a02699d301b933e1d95b/packages/teraslice/lib/cluster/services/execution.js#L321-L329

In the kubernetes case, the call to `allocateSlicer()` ([ref](https://github.com/terascope/teraslice/blob/master/packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/index.js#L85-L111)) will return as soon as the request to the kubernetes API is completed.  That does not imply that those pods have started and are running however.

godber:
@peterdemartini suggests that this is probably also true in the native clustering case.

godber:
A kubernetes only solution would be to use an init container that blocks startup of the workers until the execution controller is responding:

https://kubernetes.io/docs/concepts/workloads/pods/init-containers/

godber:
There are at least two classes of solutions for this:

* master based polling or event driven solution (implemented in PR above)
* worker init container based solution

While testing failure scenarios I realized one of the disadvantages of the master based solutions is that they will be inherently slower than the worker init container solution.  This is because the worker deployments are not created until after the controller is up, which on minikube has proven to be at least 6s.  After this is up, THEN the worker deployment gets created and its containers will get pulled.  Whereas the init container based solution will result in all of the containers getting pulled at about the same time, and just the starting of worker containers will be blocked until they can reach the execution controller.

The other problem with the master polling the execution controller is that it doesn't test for the ACTUAL problem, which is whether or not the worker can reach the execution controller.  This is still a pretty unlikely/edge case so it's not that big of a deal.  I just keep feeling like it's going to bite us in some unforeseen way.

godber:
In my testing I noticed the following log entry:

```
[...]

PR #1673, opened 2020-02-11
check execution controller before worker startup

godber:
To address #1649 we will try to check that the execution
controller has started up before allocating the workers.


godber:
@peterdemartini, when you get a chance, look for the comments with `FIXME`.

godber:
The e2e test failed.  It looks like it failed in checking the worker allocations, this looks like my change broke native clustering, unless there are known issues there.

packages/teraslice/lib/cluster/services/execution.js
```diff
@@ -15,6 +15,8 @@ const {
     cloneDeep,
 } = require('@terascope/utils');
 const { makeLogger } = require('../../workers/helpers/terafoundation');
+// FIXME: Where should this module go?
+const { waitForTcpPortOpen } = require('./cluster/backends/kubernetes/net');
```
peterdemartini: This should go in [lib/utils/port_utils.js](https://github.com/terascope/teraslice/blob/master/packages/teraslice/lib/utils/port_utils.js)

packages/teraslice/lib/cluster/services/execution.js
```diff
@@ -326,6 +328,25 @@ module.exports = function executionService(context, { clusterMasterServer }) {
...
+                        port: execution.slicer_port,
+                        logger,
+                        // FIXME: Should we create a new timeout setting for this?
+                        // This is how long the master will retry connecting to
+                        // the execution controller before it errors out.
+                        retryTimeout: 60000
+                    });
+                } catch (err) {
+                    // FIXME: We should error out the job here too, right?
+                    throw new TSError(err, {
```
peterdemartini: Yes, we should probably set the execution status to `terminated` and maybe attempt to clean up the k8s resources?

packages/teraslice/test/lib/cluster/services/cluster/backends/kubernetes/net-spec.js
```diff
@@ -0,0 +1,78 @@
...
+
+// Run tests with DEBUG=True to see output
+function l(msg) {
+    if (process.env.DEBUG) {
+        // eslint-disable-next-line no-console
+        console.log(msg);
+    }
+}
+
+const logger = {
```
peterdemartini: Use the `debugLogger` from `@terascope/utils`

```suggestion
const logger = debugLogger('port_utils');
```

packages/teraslice/test/lib/cluster/services/cluster/backends/kubernetes/net-spec.js
```diff
@@ -0,0 +1,78 @@
...
+let serverPort;
+
+async function setServerPort() {
+    serverPort = await Porty.find({ min: 14500, max: 14549 });
+    logger.debug(`serverPort: ${serverPort}`);
+}
+
+async function startServer() {
+    server = await net.createServer();
+    await server.listen(serverPort, () => { });
```
peterdemartini:
```suggestion
    return new Promise((resolve) => server.listen(serverPort, () => { resolve(); });
```

packages/teraslice/test/lib/cluster/services/cluster/backends/kubernetes/net-spec.js
```diff
@@ -0,0 +1,78 @@
...
+}
+
+async function stopServer() {
+    await server.close();
```
peterdemartini:
```suggestion
    return new Promise((resolve) => server.close(() => { resolve(); });
```

packages/teraslice/test/lib/cluster/services/cluster/backends/kubernetes/net-spec.js
```diff
@@ -0,0 +1,78 @@
...
+
+        test('should not raise an error', async () => {
+            const options = {
+                host: 'localhost',
+                logger,
+                port: serverPort,
+                retryFrequency: 100,
+                retryTimeout: 2000,
+            };
+            await expect(waitForTcpPortOpen(options)).resolves.toBe();
```
peterdemartini: I prefer this matcher since it is more clear.

```suggestion
            await expect(waitForTcpPortOpen(options)).toResolve();
```

packages/teraslice/test/lib/cluster/services/cluster/backends/kubernetes/net-spec.js
```diff
@@ -0,0 +1,78 @@
...
+        });
+    });
+
+    // eslint-disable-next-line jest/no-disabled-tests
```
peterdemartini:
```suggestion
```

packages/teraslice/lib/cluster/services/execution.js
```diff
@@ -326,6 +328,25 @@ module.exports = function executionService(context, { clusterMasterServer }) {
...
+                // the execution controller is up when the workers are started.
+                try {
+                    await waitForTcpPortOpen({
+                        host: execution.slicer_hostname,
+                        port: execution.slicer_port,
+                        logger,
+                        // FIXME: Should we create a new timeout setting for this?
+                        // This is how long the master will retry connecting to
+                        // the execution controller before it errors out.
+                        retryTimeout: 60000
```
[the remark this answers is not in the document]
godber: Technically that timeout has a different meaning, right?  By doing so, I'd be overloading the purpose of that setting.  It may be a good place to start.  My question would then be, "Should I change the wording of that doc and the copy of that in the docs to reflect the new overloaded meaning?"

packages/teraslice/test/lib/cluster/services/cluster/backends/kubernetes/net-spec.js
godber: Oh, yeah, thanks!  That was all new stuff to me, I was just pleased to figure out how to do it at all.

godber:
@peterdemartini you were right about a missing config setting, but that hasn't fixe the failing test.

godber:
So, we've managed to get the tests passing here, mostly by making `submitAndStart` wait for the expected number of workers to be running.

This MR also includes changes that need to get into master even if this branch doesn't make it:

* bumps the ES version used in e2e tests to 6.8.6.
* fixes an attempt to push a test image outside of CI

**CONCERN:** I have a slight concern about the cases where executions fail or are stopped but the execution controller has never been detected by my `waitForTcpPortOpen` function.  In this case the Teraslice master will continue to test for that port every `1s` until the `retryTimeout` timeout is met (currently set to the `slicer_timeout`, so `120s` by default).

I am going to think about this concern and try some things before I mark this PR as ready.

packages/teraslice/lib/cluster/services/execution.js
```diff
@@ -302,6 +303,47 @@ module.exports = function executionService(context, { clusterMasterServer }) {
...
+    async function _handleAllocationFailure(err, execution) {
+        const error = new TSError(err, {
+            reason: `Failured to provision execution ${execution.ex_id}`
+        });
+        try {
+            await exStore.setStatus(
+                execution.ex_id,
+                'failed',
+                exStore.executionMetaData(null, getFullErrorStack(error))
+            );
```
godber: Somewhere near here, I need to delete the kubernetes service and job created in the call to `clusterService.allocateWorkers`

packages/teraslice/lib/cluster/services/execution.js
```diff
@@ -310,45 +352,22 @@ module.exports = function executionService(context, { clusterMasterServer }) {
...
-                try {
-                    await clusterService.allocateWorkers(execution, execution.workers);
-                } catch (err) {
-                    throw new TSError(err, {
-                        reason: `Failure to allocateWorkers ${execution.ex_id}`
-                    });
-                }
+                // run this in the background
+                _waitForExecutionAllocation(execution)
+                    .catch((err) => _handleAllocationFailure(err, execution));
```
godber: Doesn't this catch imply that `_waitForExecutionAllocation` needs to return a Promise?

godber:
As fun as this PR is, I am going to close it.

I am going to put this in the workers as suggested here: https://github.com/terascope/teraslice/issues/1649#issuecomment-589474388

I will pull the `waitForTcpPortOpen` and its tests out, and re-implement in TypeScript.

Peter has cherry picked my other small fixes that need to make it into master.

New work will be in a `k8s-feb-2020-r2` branch.

PR #1688, opened 2020-02-22
convert waitForTcpPortOpen to ts and make workers wait for execution controller startup

godber:
A second attempt at resolving #1649 but this time doing the check from the workers.

First commit is just the `waitForTcpPortOpen` added in the util module.


packages/teraslice-messaging/src/utils/index.ts
```diff
@@ -24,3 +25,88 @@ export function formatURL(hostname = os.hostname(), port: number): string {
...
+ * @param {number} options.retryFrequency - time in ms between retry attempts
+ * @param {number} options.retryTimeout - time in ms until it times out
+ *  connecting, name resolution timeouts are separate and not configurable
+ * @param {Object} options.logger - Teraslice logger object
+ */
+export async function waitForTcpPortOpen(options: {
+    retryFrequency?: number;
+    retryTimeout?: number;
+    port: any;
+    host: any;
```
godber: I tried to get rid of these `any`s on `port` and `host` but ran into trouble, so I've just left them.

Issue #1701, opened 2020-02-26
Stopped persistent jobs missing `_slicer_stats` in execution context


macgyver603:
This appears to be a regression between `v0.63.3` and `v0.63.4`. When stopping a persistent job, the execution context is missing `_slicer_stats` like so:

```
zburnham$ curl -Ss my-test-ts.test.lan/jobs/7edc8a6e-38fb-472c-b5a4-9e0038c3b6ce/ex
{
    "name": "kf-to-es-es-1-6-1-kf-2-5-2",
    "lifecycle": "persistent",
    "workers": 3,
    "recycle_worker": 1000,
    "operations": [
        {
            "_op": "kafka_reader",
            ...
        },
        {
            "_op": "elasticsearch_index_selector",
            ...
        },
        {
            "_op": "elasticsearch_bulk",
            ...
        }
    ],
    "assets": [
        ...
    ],
    "analytics": true,
    "performance_metrics": false,
    "autorecover": false,
    "max_retries": 3,
    "apis": [],
    "probation_window": 300000,
    "slicers": 1,
    "labels": null,
    "env_vars": {},
    "targets": [],
    "volumes": [],
    "job_id": "7edc8a6e-38fb-472c-b5a4-9e0038c3b6ce",
    "ex_id": "a8e01ee1-c35e-49b4-8028-baa8df4b83c6",
    "metadata": {},
    "_status": "stopped",
    "_context": "ex",
    "_created": "2020-02-25T23:13:16.464Z",
    "_updated": "2020-02-25T23:15:10.989Z",
    "_has_errors": false,
    "_slicer_stats": {},                      <------------------- Missing Stats
    "_failureReason": "",
    "slicer_port": 45680,
    "slicer_hostname": "ts-exc-kf-to-es-es-1-6-1-kf-2-5-2-7edc8a6e-38fb"
}
```

macgyver603:
This is working again with `v0.63.6`.

PR #1702, opened 2020-02-28
v0.63.6 - Fix/slicer stats on stopped jobs

peterdemartini:
- Fix `_slicer_stats.slicers`, which was previously returning a function
- Fix updating the `executionMetaData` function to not set `_slicer_stats` to `null` and effectively remove the `_slicer_stats` when updating the execution.
- Improves the code documentation around `executionMetaData` functionality

Resolves https://github.com/terascope/teraslice/issues/1701

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "teraslice-workspace",
     "displayName": "Teraslice",
-    "version": "0.63.5",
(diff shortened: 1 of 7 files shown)


godber:
This looks good, I tested 15 executions without the missing _slicer_stats occurring.

PR #1758, opened 2020-03-16
Memory leak fixes

peterdemartini:
Memory leak fixes:

- Avoid creating child loggers too often (this causes memory leaks)
- Make sure to properly dereference the function in the once util

Other fixes:

- fix s3 connector in `terafoundation`
- fix xlucene reference in `README`
- improve setup script to tsc --build --force
- Slight improvement to matches check in pRetry
- Fix reference to file in `k8Resource.js`


peterdemartini:
I verified this does fix memory leaks in our internal project

PR #1761, opened 2020-03-19
Directive support in data-mate

jsnoble:
- added "Any" data-type
- initial changes for directive support in spaces

### DIFF
--- a/packages/data-types/src/types/v1/any.ts
+++ b/packages/data-types/src/types/v1/any.ts
@@ -1,0 +1,17 @@
+import BaseType from '../base-type';
+
+export default class AnyType extends BaseType {
+    toESMapping(_version?: number) {
+        return {
+            mapping: {}
+        };
+    }
+
+    toGraphQL() {
+        return this._formatGql('JSON');
+    }
+
+    toXlucene() {
+        return {};
+    }
+}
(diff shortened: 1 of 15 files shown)


packages/data-types/src/types/v1/any.ts
```diff
@@ -0,0 +1,17 @@
+import BaseType from '../base-type';
+
+export default class AnyType extends BaseType {
+    toESMapping(_version?: number) {
+        return {
+            mapping: {}
```
peterdemartini: I am thinking for this, we should probably use an [unindexed field](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/enabled.html) so that it doesn't get auto indexed by elasticsearch causing weird issues.

```suggestion
            mapping: { [this.field]: { enabled: false }}
```

packages/data-types/test/types/v1/any-spec.ts
```diff
@@ -0,0 +1,43 @@
+import AnyType from '../../../src/types/v1/any';
+import { FieldTypeConfig } from '../../../src/interfaces';
+
+describe('Boolean V1', () => {
```
peterdemartini: This file references Boolean in a few places

```suggestion
describe('Any V1', () => {
```

packages/data-types/test/types/v1/any-spec.ts
```diff
@@ -0,0 +1,43 @@
...
+    const field = 'someField';
+    const typeConfig: FieldTypeConfig = { type: 'Boolean' };
```
peterdemartini:
```suggestion
    const typeConfig: FieldTypeConfig = { type: 'Any' };
```

PR #1779, opened 2020-03-24
K8s state filter change

godber:
The PR provides the following:

* The teraslice master will now filter for pods in its own cluster when getting the pod list rather than getting ALL pods in kubernetes and filtering later.  This is the only functional change to Teraslice.
* Updated development workflow to support both minikube and hybrid development modes.  The hybrid mode runs the teraslice master outside of kubernetes for faster development iterations


godber:
In reviewing this I recommend you just avert your eyes from the mess in examples/k8s.  The `.js` changes are fairly straight forward, and most lines changed are deleted comments and a change in indentation due to the removal of a conditional.

godber:
Ugh, forgot to update tests ... fixing.

PR #1783, opened 2020-03-27
ts-transforms, document-matcher and directive changes

jsnoble:
- ts-transforms can now work with variables, fixed an issue where a selector node was attached straight away to a post-process node
- document-matcher can now use variables
- after talking with kimbro, we are momentariy using setField with FieldTransform. Do not use it this way in other parts of the library, its very unstable, use the setField in RecordTransform
- fixed a bug in extract in data-mate


packages/ts-transforms/src/interfaces.ts
```diff
@@ -57,6 +57,11 @@ export interface SelectorConfig {
     selector: string;
 }
 
+export interface MatcherConfig {
+    typeConfig?: xLuceneTypeConfig;
```
peterdemartini: Should we call this `type_config` to be consistent with the other places where the xlucene type config is used?

jsnoble: it would be a breaking change. There are is code out there that uses that. I agree with you fully, but should we try to take care of that now? it would break the assets that use this

PR #1798, opened 2020-03-31
data-mate error fixes and data-type object support for arrays

jsnoble:
- added support in data-type `Object` for array support
- fixed extract bug so it does not throw on no results
- added `map` and `setDefault` to field transforms
- added `exists`, `guard`,  `isArray`,`every`, `guard` to field validations

### DIFF
--- a/packages/data-mate/src/transforms/field-transform.ts
+++ b/packages/data-mate/src/transforms/field-transform.ts
@@ -12,7 +12,8 @@ import {
 import {
     isString,
     isValidDate,
-    isNumber
+    isNumber,
+    isArray
 } from '../validations/field-validator';
 import { Repository } from '../interfaces';
 
@@ -154,7 +155,45 @@ export const repository: Repository = {
         },
         output_type: 'Any' as AvailableType
     },
+    setDefault: {
+        fn: setDefault,
+        config: {
+            value: {
+                type: 'Any'
+            }
+        },
+        output_type: 'Any' as AvailableType
+    },
+    map: {
+        fn: map,
+        config: {
+            fn: {
+                type: 'String'
+            },
+            args: {
+                type: 'Object'
+            }
+        },
+        output_type: 'Any' as AvailableType
+    },
 };
+
+export function setDefault(input: any, args: { value: any }) {
+    if (input === undefined) {
+        if (args.value === undefined) throw new Error('Parameter value cannot be set to undefined');
+        return args.value;
+    }
+    return input;
+}
+
+export function map(input: any[], args: { fn: string; options?: any }) {
+    if (!isArray(input)) throw new Error('Input must be an array');
+    const { fn, options } = args;
+    const repoConfig = repository[fn];
+    if (!repoConfig) throw new Error(`No function ${fn} was found in the field transform respository`);
+
+    return input.map((data) => repoConfig.fn(data, options));
+}
 
 // TODO: this is currently a hack for directives, this will evolve, do not use it for other purposes
 export function setField(_input: any, args: { field: string; value: any }) {
@@ -372,7 +411,7 @@ export function extract(
     }
 
     const results = extractAndTransferFields();
-    if (results == null) throw new Error('Nothing to extract');
+    if (results == null) return null;
 
     return results;
 }
--- a/packages/data-mate/src/transforms/field-transform.ts
+++ b/packages/data-mate/src/transforms/field-transform.ts
@@ -170,7 +170,7 @@ export const repository: Repository = {
             fn: {
                 type: 'String'
             },
-            args: {
+            options: {
                 type: 'Object'
             }
         },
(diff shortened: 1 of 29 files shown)


packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -154,8 +155,46 @@ export const repository: Repository = {
...
+            args: {
+                type: 'Object'
+            }
+        },
+        output_type: 'Any' as AvailableType
+    },
 };
 
+export function setDefault(input: any, args: { value: any }) {
+    if (input === undefined) {
```
peterdemartini, line 182: This should also work against `null` values.

packages/data-mate/src/validations/field-validator.ts
```diff
@@ -366,3 +371,36 @@ export function isPostalCode(input: any, { locale }: { locale: 'any' | PostalCod
 export function isValidDate(input: any): boolean {
     return !isBoolean(input) && ts.isValidDate(input);
 }
+
+// NOTE: this function will throw compared to all other validations
+export function guard(input: any) {
+    if (input === undefined) throw new Error('input is empty');
```
peterdemartini, line 377: This should guard against `null` too. Also the Error message is not very helpful since it is not input, it’s a value (from the users perspective), `Expected value not to be empty, got ${getTypeOf(input)}`

packages/data-mate/src/validations/field-validator.ts
```diff
@@ -366,3 +371,36 @@ export function isPostalCode(input: any, { locale }: { locale: 'any' | PostalCod
...
+export function exists(input: any): boolean {
+    return input !== undefined;
+}
+
+export function isArray(input: any): boolean {
+    if (Array.isArray(input)) return true;
+    return false;
+}
+
+export function some(input: any, { fn }: { fn: string }): boolean {
```
peterdemartini, line 390: This needs the ability to specify the options to the validator function. Currently this won’t work because the arguments for the javascript some function are used (which works for the input arg but the second argument will always be an number)

packages/data-mate/src/validations/field-validator.ts
```diff
@@ -366,3 +371,36 @@ export function isPostalCode(input: any, { locale }: { locale: 'any' | PostalCod
...
+    return input.some(repoConfig.fn);
+}
+
+export function every(input: any, { fn }: { fn: string }): boolean {
+    if (!isArray(input)) return false;
+
+    const repoConfig = repository[fn];
+    if (!repoConfig) throw new Error(`No function ${fn} was found in the field validator respository`);
+
+    return input.every(repoConfig.fn);
```
peterdemartini, line 405: Same issue as the some function

packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -154,8 +155,46 @@ export const repository: Repository = {
...
+        if (args.value === undefined) throw new Error('Parameter value cannot be set to undefined');
+        return args.value;
+    }
+    return input;
+}
+
+export function map(input: any[], args: { fn: string; options?: any }) {
+    if (!isArray(input)) throw new Error('Input must be an array');
```
peterdemartini, line 190: The error messages in this library should match the users perspective so using Input is odd since that they are inputting this directly, to them it’s just a value for field. It’s also super useful to tell them what received in addition to what was expected so they don’t have to reprocess potentially large queries to debug it)

packages/data-mate/src/validations/field-validator.ts
```diff
@@ -366,3 +383,36 @@ export function isPostalCode(input: any, { locale }: { locale: 'any' | PostalCod
...
 }
+
+// NOTE: this function will throw compared to all other validations
+export function guard(input: any) {
+    if (input === undefined) throw new Error('input is empty');
+    return true;
+}
+
+export function exists(input: any): boolean {
+    return input !== undefined;
```
peterdemartini, line 394:
```suggestion
    return input != null;
```

packages/data-mate/src/validations/field-validator.ts
peterdemartini: This still needs to be fixed

packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -154,15 +155,54 @@ export const repository: Repository = {
...
+export function setDefault(input: any, args: { value: any }) {
+    if (ts.isNil(input)) {
+        if (ts.isNil(args.value)) throw new Error('Parameter value cannot be set to undefined or null');
+        return args.value;
+    }
+    return input;
+}
+
+export function map(input: any[], args: { fn: string; options?: any }) {
+    if (!isArray(input)) throw new Error(`Input must be an array, received ${ts.getTypeOf(input)}`);
```
peterdemartini, line 190: The `input` is `null` or `undefined` I don't think it would be an error condition, it would just return `null`

packages/data-mate/src/validations/field-validator.ts
```diff
@@ -386,12 +386,12 @@ export function isValidDate(input: any): boolean {
 
 // NOTE: this function will throw compared to all other validations
 export function guard(input: any) {
-    if (input === undefined) throw new Error('input is empty');
+    if (ts.isNil(input)) throw new Error('input is empty');
```
peterdemartini, line 389: This error doesn't tell me anything

```suggestion
    if (ts.isNil(input)) throw new Error('Expected value not to be empty');
```

Issue #1009, opened 2019-02-06
Reconsider the use of k8s service for the execution controller


godber:
I think my line of reasoning on using a service for the execution controller was to provide the workers a well defined single point to access the execution controller.  I've forgotten some of the implementation details at the moment, but it's worth reconsidering whether it's worth it.  And perhaps if it has some tie in with the restartability of execution controllers.

godber:
I've spent some time refreshing my understanding of Kubernetes networking and digging into exact behaviors.  At this point and as far as I can tell, pods don't get DNS names other than the ones like this `<IP with dashes>.<namespace>.pod.cluster.local`

```
host 10-31-11-52.namespace1.pod
10-31-11-52.namespace1.pod.cluster.local has address 10.31.11.52
```

It would still be possible and perhaps beneficial to remove the service and use the above name or just IP address.  We'd probably end up having to add the service back in if we wanted to support high availability execution controllers.

godber:
I think the appropriate way to wait for the pods to be created is to use the `.spec.selector.matchLabels.controller-uid` returned from the creation of the execution controller job, use that as a selector on pods and wait for that pod to reach the Running state.  I am implementing this now.

Tracking job events was a dead end since they didn't provide any pod information.

PR #1809, opened 2020-04-02
K8s remove exc svc

godber:
This PR does the following

* removes the Kubernetes Service that sat between the workers and execution controller pods.  The kubernetes deployment for the workers won't even get created until after the execution controller pod is in the `Running` state and the `podIp` is available. refs: #1009, #1649 
* removes the call to `waitForTcpOpen` in the messaging library. refs: #1649


godber:
@peterdemartini ... I should probably remove `waitForTcpOpen` huh?  I've removed the call from the messaging module (there not much point now).

I should remove the util function and tests too right?
https://github.com/terascope/teraslice/blob/master/packages/teraslice-messaging/src/utils/index.ts#L41

peterdemartini:
> @peterdemartini ... I should probably remove `waitForTcpOpen` huh?  I've removed the call from the messaging module (there not much point now).
> 
> 
> 
> I should remove the util function and tests too right?
> 
> https://github.com/terascope/teraslice/blob/master/packages/teraslice-messaging/src/utils/index.ts#L41

Sure, since it didn’t help, we can remove it, if we ever need it back we can look it up in git history.

packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/k8s.js
```diff
@@ -57,6 +59,38 @@ class K8s {
...
+     *
+     * NOTE: If your selector will return multiple pods, this method probably
+     * won't work for you.
+     * @param {String} selector kubernetes selector, like 'controller-uid=XXX'
+     * @param {String} ns       namespace to search, this will override the default
+     * @param {Number} timeout  time, in ms, to wait for pod to start
+     */
+    async waitForSelectedPod(selector, ns, timeout = 10000) {
+        const namespace = ns || this.defaultNamespace;
+        let now = (new Date()).getTime();
```
peterdemartini, line 74: I prefer `Date.now()` for simplicity

godber:
Assuming these test pass (they should) ... this is ready to go.

packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/k8s.js
```diff
@@ -57,6 +59,38 @@ class K8s {
...
+        // eslint-disable-next-line no-constant-condition
+        while (true) {
+            const result = await this.client.api.v1.namespaces(namespace)
+                .pods().get({ qs: { labelSelector: selector } });
+
+            // NOTE: This assumes the first pod returned.
+            const pod = result.body.items[0];
+
+            if (pod.status.phase === 'Running') return pod;
+            if (now > end) throw new Error('timeout');
```
peterdemartini, line 86: This error message should be more descriptive

packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/k8s.js
```diff
@@ -57,6 +59,38 @@ class K8s {
...
+            this.logger.debug(`waiting for pod matching: ${selector}`);
+
+            await pDelay(500);
+            now = (new Date()).getTime();
```
peterdemartini, line 90: Use `Date.now()`

godber: Geez there was a second one of those?

godber:
Hold on ... I accidentally committed some garbage to example jobs.  Don't merge.

PR #1823, opened 2020-04-06
Data mate array handling and documentation

jsnoble:
- added array handling to all field validations/transforms where appropriate
- added jsdocs for all methods
- added additional tests to verify array handling and added missing tests
- various bug fixes


packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -453,34 +991,77 @@ export function replaceRegex(input: string, {
...
     if (global) options += 'g';
 
-    try {
-        const re = new RegExp(regex, options);
-        return input.replace(re, replace);
-    } catch (e) {
-        throw new Error(e.message);
+    if (isArray(input)) {
+        return input.filter(ts.isNotNil).map((data: any) => {
+            const re = new RegExp(regex, options);
```
peterdemartini, line 996: This is missing the same string validation as below.

packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -492,37 +1073,135 @@ export function toUnixTime(input: any, { ms = false } = {}) {
...
+    if (isArray(input)) {
+        return input.filter(ts.isNotNil).map((data: any) => makeUnitTime(data, { ms }));
     }
 
+    if (!isValidDate(input)) throw new Error('Not a valid date, cannot transform to unix time');
+
+    return makeUnitTime(input, { ms });
+}
+
+function makeIso(input: any, args?: { resolution?: 'seconds' | 'milliseconds' }) {
```
peterdemartini, line 1107: We should be consistent with the naming here:

```suggestion
function _makeISO(input: any, args?: { resolution?: 'seconds' | 'milliseconds' }) {
```

packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -492,37 +1073,135 @@ export function toUnixTime(input: any, { ms = false } = {}) {
...
+
+    if (isArray(input)) {
+        return input
+            .filter(ts.isNotNil)
+            .map((data: any) => makeIso(data, args));
+    }
+
     if (!isValidDate(input)) {
-        throw new Error('Not a valid date');
+        throw new Error('Input is not valid date');
```
peterdemartini, line 1140:
```suggestion
        throw new Error('Input is not valid date, received ${input}');
```

packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -492,37 +1073,135 @@ export function toUnixTime(input: any, { ms = false } = {}) {
...
     if (ts.isNil(input)) return null;
-    const { format, resolution } = args;
-    if (!isString(format)) throw new Error(`Invalid parameter format, must be a string, recieved ${ts.getTypeOf(input)}`);
-    // convert string to date
-    // validate input as datelike
+
+    if (isArray(input)) {
+        return input
+            .filter(ts.isNotNil)
+            .map((data: any) => makeIso(data, args));
```
peterdemartini, line 1136: This missing the `isValidDate` check

packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -492,37 +1073,135 @@ export function toUnixTime(input: any, { ms = false } = {}) {
...
+export function toUnixTime(input: any, { ms = false } = {}) {
     if (ts.isNil(input)) return null;
-    if (!isValidDate(input)) {
-        throw new Error('Not a valid date');
+
+    if (isArray(input)) {
+        return input.filter(ts.isNotNil).map((data: any) => makeUnitTime(data, { ms }));
     }
 
+    if (!isValidDate(input)) throw new Error('Not a valid date, cannot transform to unix time');
```
peterdemartini, line 1102:
```suggestion
    if (!isValidDate(input)) throw new Error(`Not a valid date, cannot transform ${input} to unix time`);
```

packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -453,34 +991,77 @@ export function replaceRegex(input: string, {
...
+        return input.split(delimiter);
+    }
+
+    throw new Error('Input must be a string or an array');
 }
 
-// option to specify, seconds, millisecond, microseconds?
-export function toUnixTime(input: any, { ms = false } = {}) {
-    if (!isValidDate(input)) throw new Error('Not a valid date, cannot transform to unix time');
+function makeUnitTime(input: any, { ms = false } = {}) {
```
peterdemartini, line 1064:
```suggestion
function makeUnixTime(input: any, { ms = false } = {}) {
```

packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -453,34 +991,77 @@ export function replaceRegex(input: string, {
...
-    const delimiter = args ? args.delimiter : '';
+    if (isArray(input)) return input;
 
-    return input.split(delimiter);
+    if (isString(input)) {
+        const delimiter = args ? args.delimiter : '';
+        return input.split(delimiter);
+    }
+
+    throw new Error('Input must be a string or an array');
```
peterdemartini, line 1061:
```suggestion
    throw new Error(`Input must be a string, or an array, got ${getTypeOf(input)}`);
```

packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -537,32 +1216,138 @@ export function parseDate(input: any, args: { format: string }) {
     return parsed;
 }
 
+/**
+ * Will use date-fns parse against the input and return a date object
+ *
+ * @example
+ *
+ * const resutls1 = fieldTransform.parseDate('2020-01-10-00:00', { format: 'yyyy-MM-ddxxx' })
```
peterdemartini, line 1224: You misspelled `results` in this a few times. Also I really dislike using generic variable names that increment. Since this is pseudo code just call it `result`.

packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -537,32 +1216,138 @@ export function parseDate(input: any, args: { format: string }) {
     return parsed;
 }
 
+/**
+ * Will use date-fns parse against the input and return a date object
```
peterdemartini, line 1220: This doesn't return a date object, it returns a date formatted string (make sure to update the examples)

packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -453,34 +991,77 @@ export function replaceRegex(input: string, {
...
+ * @param {ReplaceLiteralConfig} { search, replace }
+ * search is the word that is to be changed to the value specified with the paramter replace
+ * @returns { string | string[] | null } returns null if input is null/undefined
+ */
+
+export function replaceLiteral(input: StringInput, { search, replace }: ReplaceLiteralConfig) {
     if (ts.isNil(input)) return null;
+
+    if (isArray(input)) {
+        return input.filter(ts.isNotNil).map((data: any) => data.replace(search, replace));
```
peterdemartini, line 1027: This is missing the `isString` validation

packages/data-mate/src/transforms/field-transform.ts
jsnoble: it does not return a string, but a date. The docs are correct.  https://date-fns.org/v2.11.1/docs/parse

packages/data-mate/src/validations/field-validator.ts
```diff
@@ -416,3 +1255,11 @@ export function every(input: any, { fn, options }: { fn: string; options?: any }
 
     return input.every((data: any) => repoConfig.fn(data, options));
 }
+
+export function isTuple(input: any) {
```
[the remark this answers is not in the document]
jsnoble: this is geopoint specific. I renamed it to isNumberTuple for further clarification

packages/data-mate/test/field_transform-spec.ts
```diff
@@ -257,159 +341,432 @@ describe('field transforms', () => {
...
+
+        it('should encode values', () => {
+            expect(fieldTransform.encodeSHA1(source)).toBeDefined();
+            expect(fieldTransform.encodeSHA1(null)).toBe(null);
+            expect(fieldTransform.encodeSHA1(undefined)).toBe(null);
+        });
+
+        it('should encode values in an array', () => {
+            // @ts-ignore
+            expect(fieldTransform.encodeSHA1([source, null])).toBeArrayOfSize(1);
```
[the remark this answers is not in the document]
jsnoble: 'as any' hides a bunch of things as well, it essentially loses all types for that input. though it is marginally better if there is a second parameter as that can still be typed.  In the end it's still dangerous. Granted the intent of placing those ignores there was to test ways the function can be misused

jsnoble:
As for the documentation, im not sure fully what `This is still missing updated, user documentation like, in the docs/packages/data-mate/overview.md that explains the functions for an end user, not a developer.` means. Can you point to any of our documentation or outside documentation for this? How would they differ. A user would have to know what the method name is, what it does, and how to use it. How is this any different? It already does so. The overview is generated letting others know what exists, a description of what each thing does, and an example of how to use it. As for formating of the docs, what is off compared to all our other documentation like in utils or other packages that is generated. Is this supposed to be special cased compared to the others?

jsnoble:
as a side note I thought that the build of this generates docs. Am I supposed to save what yarn docs outputs or let it handle that by itself?

packages/data-mate/src/transforms/field-transform.ts
peterdemartini: The doc comment as a `@returns` that says it returns a string or array of strings. But regardless maybe we want it to return a date formatted string rather than return just parsing the date? I think that would be more useful since the you the format you want when you call `date.toString()` which will happen when it it is json stringified.

packages/data-mate/test/field_transform-spec.ts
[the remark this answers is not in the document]
peterdemartini: If you use `@ts-ignore` that means that the compiler will ignore the statement which means if `fieldTransform` is renamed or `encodeSHA1` takes an extra argument or `toBeArrayOfSize` is available, the compiler won't throw, it will just continue on. Using `as` basically just isolates it to the part you need to be loosely typed on.

peterdemartini:
> as a side note I thought that the build of this generates docs. Am I supposed to save what yarn docs outputs or let it handle that by itself?

`yarn docs` get called before publishing the documentation website so it is not a huge deal if you don't commit the `docs/packages/*/api/*`

packages/data-mate/src/transforms/field-transform.ts
ciorg: If it returns a string wouldn't that be very similar to the formatDate?  Where you are already getting a stringified date?  If I just want the year or hour or whatever from a timestamp or a date I would run parse then could grab what I needed.  That's what I think the use is in returning a date object.

packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -271,91 +455,421 @@ export function toISDN(input: any) {
...
+ * const encoded = 'HELLO%20AND%20GOODBYE';
+ *
+ * expect(fieldTransform.decodeUrl(encoded)).toEqual(source);
+ *
+ * @export
+ * @param {StringInput} input
+ * @returns { string | null } returns null if input is null/undefined
+ */
+
+export function decodeUrl(input: StringInput) {
```
peterdemartini, line 590: We should be consistent with naming:

```suggestion
export function decodeURL(input: StringInput) {
```

packages/data-mate/src/transforms/field-transform.ts
```diff
@@ -271,91 +455,421 @@ export function toISDN(input: any) {
...
+ * const encoded = 'HELLO%20AND%20GOODBYE';
+ *
+ * expect(fieldTransform.encodeUrl(source)).toEqual(encoded);
+ *
+ * @export
+ * @param {StringInput} input
+ * @returns { string | null } returns null if input is null/undefined
+ */
+
+export function encodeUrl(input: StringInput) {
```
peterdemartini, line 614: We should be consistent with naming:

```suggestion
export function encodeURL(input: StringInput) {
```

packages/data-mate/src/validations/field-validator.ts
```diff
@@ -252,69 +511,347 @@ export function inIPRange(input: any, args: { min?: string; max?: string; cidr?:
...
+ * expect(FieldValidator.isUrl('https://someurl.cc.ru.ch')).toBe(true);
+ * expect(FieldValidator.isUrl('ftp://someurl.bom:8080?some=bar&hi=bob')).toBe(true);
+ * expect(FieldValidator.isUrl('http://xn--fsqu00a.xn--3lr804guic')).toBe(true)
+ *
+ * @export
+ * @param {*} input
+ * @returns {boolean} boolean
+ */
+
 export function isUrl(input: any): boolean {
```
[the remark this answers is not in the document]
jsnoble: Im not sure this is right, for sure it would not be consistent to make names like that, we don't do that for most of out code base, we almost always use PascalCase.  Are we going to change everything to toBOOLEAN and toNUMBER? it's a little too much. What about isGEOSHAPEMULTIPOLYGON?  If anything the few cases where we have uppercase like isJSON should maybe to correctly pascalcased. Currently about 7 out of  43 have uppercase words to it. Most of those are special identifiers or are acronyms like JSON, or ISSN. If we are to be consistent than anything they should be lowercased to isJson or isIssn, which might be weird for the ISSN

packages/data-mate/src/validations/field-validator.ts
```diff
@@ -204,27 +392,98 @@ export function isIP(input: any) {
...
+ *
+ * @example
+ * expect(FieldValidator.isIPCidr('8.8.0.0/12')).toBe(true);
+ *
+ * @export
+ * @param {*} input
+ * @returns {boolean} boolean
+ */
+
+export function isCidr(input: any) {
```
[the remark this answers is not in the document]
jsnoble: please refer to top comment

packages/data-mate/src/validations/field-validator.ts
```diff
@@ -252,69 +511,347 @@ export function inIPRange(input: any, args: { min?: string; max?: string; cidr?:
...
 
+/**
+ * Validates that the input is ascii chars or a list of ascii chars
+ *
+ * @export
+ * @param {*} input
+ * @returns {boolean} boolean
+ */
+
 export function isAscii(input: any): boolean {
```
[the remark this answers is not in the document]
jsnoble: top comment

packages/data-mate/src/validations/field-validator.ts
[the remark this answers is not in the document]
jsnoble: what would be the factor in determining on when to abandon  camelcased function names?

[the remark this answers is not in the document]
jsnoble: should we have only words that are acronyms be uppercased?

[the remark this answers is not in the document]
peterdemartini: This consistency I am suggesting only applies to acronyms, so `toBoolean`/`toNumber`,etc wouldn't have to change. Camel casing has varied support for the formatting of an acronyms, which can either be treated like the first letter of each of word is already in the correct capitalization or the just treat the acronyms as one word. IMO it's more correct/readable to use `toJSON` than `toJson` since that is way javascript does it and it makes it easier to deal acronyms like `PhD`. However in this library we are not using consistently.

The reason I suggest making this change now is because this will soon be used widely and changing it will be really difficult. We just have to pick a format and stick to it. Another alternative would be to use snake case since that may be easier to read.

kstaken:
@peterdemartini should PRs include the generated docs?

peterdemartini:
Anything under `docs/packages/*/api/**/*.md` doesn't need to be committed since it is autogenerated. I thought I added a .gitignore for that but I guess not.

docs/packages/data-mate/overview.md
```diff
@@ -6,405 +6,1403 @@ sidebar_label: Data-Mate
...
-```bash
-isBoolean(false); # true
-isBoolean('astring'); # false
-isBoolean(0); # false
+```javascript
+FieldValidator.isBoolean(false); # true
+FieldValidator.isBoolean('astring'); # false
+FieldValidator.isBoolean(0); # false
+FieldValidator.isBoolean([true, undefined]); # true
+FieldValidator.isBoolean(['true', undefined]; # false
```
peterdemartini, line 175: Javascript doesn't have a `#`, use `//`. This making copying a little easier and reading it more clear because syntax highlighting will pick it up.

(FYI - this happens a lot in this file)

docs/packages/data-mate/overview.md
```diff
@@ -6,405 +6,1403 @@ sidebar_label: Data-Mate
...
+
+`toJSON(input) - Converts input to JSON
+if given an array it will convert everything in the array excluding null/undefined values`
+
+`returns null if input is null/undefined`
+
+```javascript
+ const obj = { hello: 'world' };
+ const results = FieldTransform.toJSON(json);
+ * results === '{"hello": "world"}'
```
peterdemartini, line 1046:
```suggestion
 // results === '{"hello": "world"}'
```

packages/data-mate/src/validations/field-validator.ts
```diff
@@ -252,69 +504,349 @@ export function inIPRange(input: any, args: { min?: string; max?: string; cidr?:
...
+ * @example
+ * expect(FieldValidator.equals('12345', { value: '12345' })).toBe(true);
+ * expect(FieldValidator.equals('hello', { value: 'llo' })).toBe(false);
+ *
+ * @param {*} input
+ * @param {{ value: string }} { value }
+ * @returns {boolean} boolean
+ */
+
+export function equals(input: any, args: { value: string }): boolean {
```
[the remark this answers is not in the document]
peterdemartini: The functionality was updated here but not the docs

Issue #1827, opened 2020-04-07
Undefined error in k8s teraslice v0.64.2


godber:
Some teraslice jobs will fail to start up with the following error message in the master

```
"_failureReason": "TSError: Failured to provision execution 45a5f585-3a60-4189-b3c0-661cbd8c6690, caused by TypeError: Cannot read property 'status' of undefined
    at Timeout.allocator [as _onTimeout] (/app/source/packages/teraslice/lib/cluster/services/execution.js:337:31)
    at process._tickCallback (internal/process/next_tick.js:68:7)
Caused by: TypeError: Cannot read property 'status' of undefined
    at K8s.waitForSelectedPod (/app/source/packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/k8s.js:85:21)
    at process._tickCallback (internal/process/next_tick.js:68:7)"
```

godber:
This is fixed as of v0.64.3, but that release still has issues, see #1831

PR #1828, opened 2020-04-08
test for undefined responses

godber:
This should address the undefined response from the API ... refs: #1827


godber:
I need to return the `pod` object as well, once status is running.  The linter was yelling at me about destructuring  arrays.  I'll push a small change to improve readability.

PR #1832, opened 2020-04-09
Update typedoc and improve doc scripts

peterdemartini:
- Updates typedoc to include styling improvements
- Slightly improve doc scripts
- Removes HDFS asset bundle from docs
- Removes the `docs/packages/*/api` files from git
- Ignore destructuring rule for 'camelcase' rule in eslint


packages/scripts/src/helpers/docs/typedoc.ts
```diff
@@ -14,12 +17,14 @@ async function writeDocFile(filePath: string, { title, sidebarLabel }: { title:
...
-        .slice(isOverview(filePath) ? 3 : 2)
+        .slice(isOverview(filePath) ? 5 : 4)
         .join('\n')
         .trim();
 
     // fix paths
-    contents = contents.replace(/README\.md/g, 'overview.md');
+    contents = contents
+        .replace(/(\]\(.*)index\.md/g, '$1overview.md')
+        .replace(/(\]\(.*)index\.md/g, '$1overview.md');
```
jsnoble, line 27: why is this line duplicated? Should this be a single replace call or should the regex be different?

peterdemartini: It was duplicated because some reason it didn't deal with multiple `index.md` in one line and I wasn't able to get the regex to work right and this worked just fine. Do you think I should I add a comment for that to make it clear or try and fix it right?

jsnoble: yeah, because thats really strange. usually a `/g` will will get everything, I would be confused why it doesn't do it

PR #1860, opened 2020-04-20
Fix process.env usage in @terascope/utils for webpack

peterdemartini:
The [`webpack.DefinePlugin`](https://webpack.js.org/plugins/define-plugin/) only allows getting the `NODE_ENV` environment variable using `process.env.NODE_ENV` and doesn't work with destructuring.

### DIFF
--- a/packages/utils/src/env.ts
+++ b/packages/utils/src/env.ts
@@ -1,6 +1,4 @@
-const { NODE_ENV = 'production' } = process.env;
-
-export const isProd = NODE_ENV === 'production';
-export const isTest = NODE_ENV === 'test';
-export const isDev = NODE_ENV === 'development';
+export const isProd = !process.env.NODE_ENV || process.env.NODE_ENV === 'production';
+export const isTest = process.env.NODE_ENV === 'test';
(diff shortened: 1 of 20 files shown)


packages/utils/src/env.ts
```diff
@@ -1,6 +1,4 @@
-const { NODE_ENV = 'production' } = process.env;
-
-export const isProd = NODE_ENV === 'production';
-export const isTest = NODE_ENV === 'test';
-export const isDev = NODE_ENV === 'development';
+export const isProd = !process.env.NODE_ENV || process.env.NODE_ENV === 'production';
```
jsnoble, line 1: what behaviour changes should we expect to see if everything defaults to production? Is this going to cause wonky errors unless everyone has a environment variable?

peterdemartini: It already did default to `production`

peterdemartini: It was defaulted to `production` because we don't use `NODE_ENV` everywhere (like running a teraslice job locally) and it better to assume it is production than to assume it is development.

PR #1864, opened 2020-04-22
Data mate transforms integration

jsnoble:
- add initial support of data-mate to ts-transforms


jsnoble:
basic examples of using extract function

```javascript
// when using jexlExp, the second parameter is the important one
const results = FieldTransform.extract('bar', { foo: 'bar' }, { jexlExp: 'foo|toUpperCase' });
results === 'BAR'
```

examples of using jexl by itself:
```javascript
import { jexl } from '@terascope/data-mate';

jexl.evalSync('foo|toUpperCase', { foo: 'bar' }) === "BAR";

jexl.evalSync('foo|isString', { foo: 'bar' }) === true;
jexl.evalSync('foo|isNumber', { foo: 'bar' }) === false;

// example of running an extract
// needs the parentContext (second parameter) have to hard code for now
const results = jexl.evalSync('foo|extract({ foo: "bar" }, { jexlExp: "foo|toUpperCase" })', { foo: 'bar' })
results === 'BAR';


// Record level fn's, these only really work well on sub objects
// currently don't have a good way to inject parentContext (the second parameter)
const results = jexl.evalSync('field|required({}, { fields: ["foo"]})', { field: { foo: 'bar' } })).toEqual({ foo: 'bar' }
results === { foo: 'bar' }

const results = jexl.evalSync('field|dropFields({}, { fields: ["foo"]})', { field: { foo: 'bar', hello: 'there' } })
results === { hello: 'there' }
;

```

jsnoble:
Problematic Validator FN's (these should not be used if possible)
- alpha
- alphanumeric
- ascii
- divisibleby
- fqdn
- isbn

package.json
```diff
@@ -35,7 +35,9 @@
         "ms": "^2.1.2",
         "typescript": "^3.8.3"
     },
-    "dependencies": {},
+    "dependencies": {
+        "sort-package-json": "1.40.0"
```
peterdemartini, line 39: This shouldn't be added here, the dependency version should be locked down in the `packages/scripts/package.json` to `~1.40.0` - that those installing `@terascope/utils` they don't get the same problem.

packages/data-mate/src/jexl/index.ts
```diff
@@ -0,0 +1,206 @@
...
+    if (ts.isNil(args)) throw new Error('Argument parameters must be provided');
+    if (!FieldValidator.isString(args.jexlExp) || !FieldValidator.isLength(args.jexlExp, parentContext, { min: 1 })) throw new Error('Argument parameter jexlExp must must be provided and be a string value');
+    if (!FieldValidator.isString(args.field) || !FieldValidator.isLength(args.field, parentContext, { min: 1 })) throw new Error('Argument parameter field must must be provided and be a string value');
+
+    if (FieldValidator.isArray(input)) {
+        return input
+            .map((data: any) => {
+                if (!ts.isObjectLike(data)) return null;
+                const value = jexl.evalSync(args.jexlExp, data);
+                if (ts.isNotNil(value)) data[args.field] = value;
```
peterdemartini, line 195: What happens if you want to set the `value` to `null`? Or delete the value for a field?

Also this mutates the incoming data

packages/data-mate/src/jexl/index.ts
```diff
@@ -0,0 +1,206 @@
...
+                return data;
+            })
+            .filter(ts.isNotNil);
+    }
+
+    if (!ts.isObjectLike(input)) return null;
+
+    const value = jexl.evalSync(args.jexlExp, input);
+    if (ts.isNotNil(value)) input[args.field] = value;
+    return input;
```
peterdemartini, line 205: Same as above

packages/data-mate/src/jexl/index.ts
```diff
@@ -0,0 +1,206 @@
...
+import { FieldTransform, RecordTransform } from '../transforms';
+import { FieldValidator, RecordValidator } from '../validations';
+import {
+    Repository, InputType, RepoConfig, ExtractFieldConfig, RecordInput
+} from '../interfaces';
+
+class Jexl extends jexlCore.Jexl {
+    _context = {};
+
+    evalSync(expression: string, _context: ts.AnyObject | undefined) {
```
peterdemartini, line 13: We should probably have support for `eval` that uses `Promise` (so we don't block the event loop) that would mean making a change to the places that use `data-mate` and make them handle returning `Promise` and `sync` results.

While this is required for this PR, I think it should definitely be considered because we will run into performance issues that can only be solved this library.

packages/data-mate/src/transforms/record-transform.ts
```diff
@@ -100,15 +108,19 @@ function _migrate(doc: ts.AnyObject, from: string, to: string) {
  *
  * const obj = { hello: 'world' };
  * const config = { field: 'other', value: 'stuff' };
- * const results = RecordTransform.setField(cloneDeep(obj), config);
+ * const results = RecordTransform.setField(cloneDeep(obj), cloneDeep(obj), config);
```
peterdemartini, line 111: We shouldn't have to cloneDeep before passing this, the `setField` shouldn't mutate.

packages/data-mate/src/transforms/record-transform.ts
```diff
@@ -138,15 +150,19 @@ export function setField(input: RecordInput, args: { field: string; value: any }
  *
  * const obj = { hello: 'world', other: 'stuff', last: 'thing' };
  * const config = { fields: ['other', 'last']} ;
- * const results = RecordTransform.dropFields(cloneDeep(obj), config);
- * expect(results).toEqual({ hello: 'world' });
+ * const results = RecordTransform.dropFields(cloneDeep(obj), cloneDeep(obj), config);
```
peterdemartini, line 153: Clone deeping shouldn't be necessary, it shouldn't mutate.

packages/ts-transforms/test/operations/transforms/base64encode-spec.ts
```diff
@@ -60,33 +60,30 @@ describe('Base64Encode operator', () => {
...
-        const results4 = test.run(data4);
-        const results5 = test.run(data5);
-        const results6 = test.run(data6);
-        const results7 = test.run(data7);
-        const results8 = test.run(data8);
-        const results9 = test.run(data9);
-        const results10 = test.run(data10);
-        const results11 = test.run(data11);
-        const results12 = test.run(data12);
+        const results1 = test.run(cloneDeep(data1));
```
peterdemartini, line 63: This is a bad smell that the incoming data shouldn't mutated. if you forget to do it, you get unexpected results, and you're more likely to into mutation problems when doing complex operations (even within a single test run)

packages/utils/src/objects.ts
```diff
@@ -22,6 +23,10 @@ export function getFirstKey<T>(input: T): (keyof T) | undefined {
     return Object.keys(input)[0] as keyof T;
 }
 
+export function isObjectLike(input: any) {
```
peterdemartini, line 26: This differs from the behavior of [`_.isObjectLike`](https://lodash.com/docs/4.17.15#isObjectLike), either we should name it something else or add it to `isSimpleObject`

packages/data-mate/src/jexl/index.ts
jsnoble: If you want to set a value to null, you can use setField, if you want to remove a field use dropFields. All of our record level transformations are mutative. In ts-transforms null is a noop and is rejected and removed. That is by design. We also pretty much do the same in directives.

packages/data-mate/src/jexl/index.ts
jsnoble: same as above

packages/data-mate/src/jexl/index.ts
jsnoble: This is not required for the PR, we currently have no use case, or any planned use case for any async transformation. This would require a ton of changes for something not needed or planned. As for not blocking the event loop that can easily be solved awaiting  between each call. Calls have to be sync  for now. This hack that I did will only really work becuase these calls are all sync in all our codebases.

packages/data-mate/src/transforms/record-transform.ts
jsnoble: why should it not mutate? its a mutative operation by design. there are complications in trying to clone this. If its a dataentity it has to be cloned in its entirety including its metadata. Cloning every single object multiple times per each operation would bring ts-transforms and directives to its knees

packages/utils/src/objects.ts
jsnoble: hmm, thats a good point. The only real difference though is that isObjectLike from lodash accepts arrays. From a deep understanding of javascript you may now that an array is an object, but I would say that that it seems to do what the function says it does if you did not have deep knowledge, it would seem to be a weird bug for others using this though its technically true. javascript is a little wierd. isSimpleObject might be a better place but would you say a DataEntity would be a simple object? Im not fully sure either

packages/data-mate/src/jexl/index.ts
peterdemartini: I am talking about the design of data-mate, having a dependency on mutations make it use less predictable and safe. We are going to building a lot stuff that does multiple data-mate "pipelines" on the same list of records or fields.

Also an important rule of thumb in programming, mutating and return a value is considered bad practice.

> Every method should either be a command that performs an action, or a query that returns data to the caller, but not both. In other words, Asking a question should not change the answer. More formally, methods should return a value only if they are referentially transparent and hence possess no side effects.
- Source to [Command Query Separation (CQS)](https://en.wikipedia.org/wiki/Command%E2%80%93query_separation) by Bertrand Meyer in [Object Oriented Software Construction](http://amzn.to/2hZ4P9C), also https://www.rahulpnath.com/blog/avoid-state-mutation/

packages/data-mate/src/transforms/record-transform.ts
peterdemartini: Once I again, I am questioning the design of being dependent on mutation. Correctness is always before performance. We won't need to clone operation just when needed (sometimes a shallow clone is all we need).

packages/data-mate/src/jexl/index.ts
```diff
@@ -0,0 +1,205 @@
...
+        if (isMultiValue) return results;
+        return results ? results[0] : results;
+    }
+
+    function callExpression() {
+        try {
+            return jexl.evalSync(jexlExp as string, parentContext);
+        } catch (err) {
+            const errMessage = `Invalid jexl expression: ${jexlExp}, error: ${err.message}`;
+            throw new ts.TSError(errMessage);
```
peterdemartini, line 143:
```suggestion
            throw new ts.TSError(err, {
                message: `Invalid jexl expression: ${jexlExp}`
            });
```

PR #1949, opened 2020-05-26
Elasticsearch api track total hits

ciorg:
* added track_total_hits to _buildRangeQuery
* bumped version
* updated tests

### DIFF
--- a/packages/elasticsearch-api/index.js
+++ b/packages/elasticsearch-api/index.js
@@ -60,6 +60,7 @@ module.exports = function elasticsearchApi(client = {}, logger, _opConfig) {
                 delete query._sourceInclude;
             }
         }
+
         return _searchES(query).then((data) => {
             if (config.full_response) {
                 return data;
@@ -522,7 +523,6 @@ module.exports = function elasticsearchApi(client = {}, logger, _opConfig) {
         const query = {
             index: opConfig.index,
             size: msg.count,
-            track_total_hits: true,
             body: _buildRangeQuery(opConfig, msg),
         };
 
@@ -539,6 +539,8 @@ module.exports = function elasticsearchApi(client = {}, logger, _opConfig) {
             const retry = _retryFn(_performSearch, query, reject);
 
             function _performSearch(queryParam) {
+                _esV7adjustments(queryParam);
+
                 client
                     .search(queryParam)
                     .then((data) => {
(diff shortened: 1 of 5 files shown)


packages/elasticsearch-api/index.js
```diff
@@ -522,6 +522,7 @@ module.exports = function elasticsearchApi(client = {}, logger, _opConfig) {
         const query = {
             index: opConfig.index,
             size: msg.count,
+            track_total_hits: true,
```
peterdemartini, line 525: I believe you have to use `trackTotalHits: true` (see the [docs](https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/16.x/api-reference.html#api-search)). Also this should only be set if the es version is `>=7` to avoid creating warnings in the server logs.

peterdemartini: You get the elasticsearch version via `getESVersion` https://github.com/terascope/teraslice/blob/2a2d3e0677dfd69556655cabad4365fab273e9d3/packages/elasticsearch-api/index.js#L52

ciorg: I'm was looking at these [docs]( https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html#request-body-search-track-total-hits) when adding it to the query.

Now I understand what you were saying yesterday about putting it on the request, I hadn't seen the api doc options for trackTotalHits.

Also the getESVersion doesn't work for v7 indices, it just returns 6 no matter what.  
I was looking at a good way to improve it, could use client.info() but that is an async function and would change other functions to be async as well, which would break some stuff.

peterdemartini: The `apiVersion` has to be to `7.0` for the `getESVersion` to return `7`. We can't rely on `client.info()` because the elasticsearch client will throw an error or drop the field if the `apiVersion` is not set to the right version

packages/elasticsearch-api/test/api-spec.js
```diff
@@ -2,6 +2,7 @@
 
 const Promise = require('bluebird');
 const { debugLogger } = require('@terascope/utils');
```
peterdemartini, line 4: Use cloneDeep from `@terascope/utils` because `lodash` isn't a dependency/devDependency of this package

PR #1968, opened 2020-06-02
Harness and Parallel-Slicer fixes

jsnoble:
- fixed parallel-slicer issues around Promise.race, it would forever be stuck in pending if no slicer was available. Also, some slicer would not correctly call.

- Updated test harness to work better with parallel-slicers, added methods to get all slices, as well as process all slices

- Added getOperationAPI to WorkerTestHarness and JobTestHarness

- fixed a case in utils to check if its an object or dataentity

### DIFF
--- a/packages/job-components/src/operations/parallel-slicer.ts
+++ b/packages/job-components/src/operations/parallel-slicer.ts
@@ -84,8 +84,6 @@ export default abstract class ParallelSlicer<T = OpConfig> extends SlicerCore<T>
     }
 
     private async processSlicer(slicer: SlicerObj) {
-        this.logger.info('slicer', slicer);
-
         slicer.processing = true;
         let result: SlicerResult;
 
@@ -114,7 +112,6 @@ export default abstract class ParallelSlicer<T = OpConfig> extends SlicerCore<T>
                 }
             }
         } finally {
-            this.logger.info(`slicer ${slicer.id} has finished processing`);
             slicer.processing = false;
         }
     }
--- a/packages/job-components/src/operations/parallel-slicer.ts
+++ b/packages/job-components/src/operations/parallel-slicer.ts
@@ -1,4 +1,4 @@
-import { times, isFunction } from '@terascope/utils';
+import { times, isFunction, pDelay } from '@terascope/utils';
 import {
     SlicerFn, SlicerResult, OpConfig, SlicerRecoveryData
 } from '../interfaces';
@@ -66,11 +66,15 @@ export default abstract class ParallelSlicer<T = OpConfig> extends SlicerCore<T>
         // which will prevent other from calling
 
         const promises = this._slicers
-            .filter((slicer) => !slicer.processing && !slicer.done)
-            .map((slicer) => this.processSlicer(slicer));
+            .filter((slicer) => !slicer.processing && !slicer.done);
 
         // calling race on an empty array will be forever pending
-        if (promises.length > 0) await Promise.race(promises);
+        if (promises.length > 0) {
+            await Promise.race(promises.map((slicer) => this.processSlicer(slicer)));
+        } else {
+            // promises are a microtask, if no action then we need to delay
+            await pDelay(10);
+        }
 
         return this.isFinished;
     }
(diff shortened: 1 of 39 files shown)


packages/job-components/src/operations/parallel-slicer.ts
```diff
@@ -54,18 +54,28 @@ export default abstract class ParallelSlicer<T = OpConfig> extends SlicerCore<T>
...
     }
 
     async handle(): Promise<boolean> {
         if (this.isFinished) return true;
 
+        // must filter out done slicers, if they are done
+        // they will return undefined which will always win race
+        // which will prevent other from calling
+
         const promises = this._slicers
```
peterdemartini, line 68: Let's call this variable: `slicers` since they aren't Promises yet

packages/teraslice-test-harness/src/slicer-test-harness.ts
```diff
@@ -120,24 +126,44 @@ export default class SlicerTestHarness extends BaseTestHarness<SlicerExecutionCo
...
     }
 
-    setWorkers(count: number) {
+    async getAllSlices(): Promise<SliceResults>;
+    async getAllSlices(options: { fullResponse: false }): Promise<SliceResults>;
+    async getAllSlices(options: { fullResponse: true }): Promise<SliceResults>;
+    async getAllSlices({ fullResponse = false } = {}): Promise<SliceResults> {
+        const results = [];
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
peterdemartini, line 135: Use `@ts-expect-error`, this will remove the need for `// eslint-disable-next-line @typescript-eslint/ban-ts-comment` and if it doesn't actually skip a compiler error, it will need to be removed. This is will help us remember to remove them they aren't need and prevent more problems than needed

packages/teraslice-test-harness/src/slicer-test-harness.ts
```diff
@@ -120,24 +126,44 @@ export default class SlicerTestHarness extends BaseTestHarness<SlicerExecutionCo
         return sliceRequests;
     }
 
-    setWorkers(count: number) {
+    async getAllSlices(): Promise<SliceResults>;
+    async getAllSlices(options: { fullResponse: false }): Promise<SliceResults>;
+    async getAllSlices(options: { fullResponse: true }): Promise<SliceResults>;
+    async getAllSlices({ fullResponse = false } = {}): Promise<SliceResults> {
```
peterdemartini, line 132: This function show throw if not in once mode

packages/teraslice-test-harness/src/slicer-test-harness.ts
```diff
@@ -120,24 +126,44 @@ export default class SlicerTestHarness extends BaseTestHarness<SlicerExecutionCo
         return sliceRequests;
     }
 
-    setWorkers(count: number) {
+    async getAllSlices(): Promise<SliceResults>;
+    async getAllSlices(options: { fullResponse: false }): Promise<SliceResults>;
+    async getAllSlices(options: { fullResponse: true }): Promise<SliceResults>;
```
peterdemartini, line 131: This overloaded methods are suppose to make easier to use, so if you pass in `{ fullResponse: true }` it knows that the result be a `SliceRequest[]` and not a `Slice[]`

packages/teraslice-test-harness/src/slicer-test-harness.ts
```diff
@@ -19,6 +20,9 @@ import { JobHarnessOptions } from './interfaces';
  * the Execution Controller, mainly the Slicer.
  * This is useful for testing Slicers.
 */
+
+type SliceResults = (AnyObject|SliceRequest|Slice|null)[];
```
peterdemartini, line 24: Why would this include `AnyObject`?

packages/teraslice-test-harness/src/slicer-test-harness.ts
```diff
@@ -79,15 +82,16 @@ export default class SlicerTestHarness extends BaseTestHarness<SlicerExecutionCo
...
     */
-    async createSlices(): Promise<SliceRequest[]>;
-    async createSlices(options: { fullResponse: false }): Promise<SliceRequest[]>;
-    async createSlices(options: { fullResponse: true }): Promise<Slice[]>;
-    async createSlices({ fullResponse = false } = {}): Promise<SliceRequest[]|Slice[]> {
+    async createSlices(): Promise<SliceResults>;
+    async createSlices(): Promise<SliceResults>;
+    async createSlices(options: { fullResponse: false }): Promise<SliceResults>;
+    async createSlices(options: { fullResponse: true }): Promise<SliceResults>;
+    async createSlices({ fullResponse = false } = {}): Promise<SliceResults> {
```
peterdemartini, line 89: Changing the overloading method signatures like is a breaking change. Since the places that were getting the correct return type if `fullResponse` is used or not used, will not get a union type of `(AnyObject|SliceRequest|Slice|null)`. These should be just modified to include `null` for each of the overloaded method signatures.

PR #1986, opened 2020-06-11
Elasticsearch api fixes

jsnoble:
- fix types in elasticsearch-api
- get/mget/search now return data-entities with metadata
- mget => if full-response is false then it returns an array of data-entities of all found data, it filters out not-found. If you use full response, it returns as normal

### DIFF
--- a/packages/elasticsearch-api/index.js
+++ b/packages/elasticsearch-api/index.js
@@ -17,7 +17,8 @@ const {
     toBoolean,
     uniq,
     random,
-    cloneDeep
+    cloneDeep,
+    DataEntity
 } = require('@terascope/utils');
 
 const DOCUMENT_EXISTS = 409;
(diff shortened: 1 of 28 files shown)


packages/types/src/misc.ts
```diff
@@ -3,3 +3,8 @@ export type MACDelimiter = 'space' | 'colon' | 'dash' | 'dot' | 'none' | 'any';
 export interface MACAddress {
     delimiter: MACDelimiter | MACDelimiter[];
 }
+
+export interface SenderApi {
```
peterdemartini, line 7: This belongs in `@terascope/job-components` since can only be used places that use `job-components`, the naming is also pretty generic and doesn't follow pascal casing (it should be `SenderAPI`). I suggest naming it something like `MultiRouteSenderAPI` is more descriptive.

Also does it really take `...params: any[]` for the methods? Can it be more specific?

packages/elasticsearch-api/index.js
```diff
@@ -48,6 +49,25 @@ module.exports = function elasticsearchApi(client = {}, logger, _opConfig) {
         return _searchES(query).then((data) => get(data, 'hits.total.value', get(data, 'hits.total')));
     }
 
+    function parseElasticsearchData(doc) {
```
peterdemartini, line 52: *minor comment:* This is kind of odd naming, it doesn't really describe it well since `Data` could mean an array of records or `parse` could you are serializing the raw JSON or a buffer. Something like `convertDocToDataEntity` makes a lot more sense.

packages/types/src/misc.ts
jsnoble: I renamed it to RouteSenderAPI, naming it MultiRouteSenderAPI would be incorrect since this is just the interface for a single route, not for multiple different senders, that sits at a layer above this. As each sender has different inputs, queries and clients its really hard to be specific. This is meant for all senders of all types

Issue #1931, opened 2020-05-20
Socket timeouts when uploading large assets


godber:
We have started encountering socket timeouts when uploading large assets or assets with lots of small files that take a long time to unzip.

godber:
I thought I could fix it by doing this:

https://github.com/terascope/teraslice/pull/1914/commits/6a781f618f8e4e90dfbfd9f41df9ff5ea907b60b

But that didn't resolve it.  Reading the asset code more closely, I see there's another express server here:

https://github.com/terascope/teraslice/blob/master/packages/teraslice/lib/cluster/services/assets.js#L16

I suppose that could be it as well.

godber:
With the change above, we get a 502 instead.

godber:
We still have an issue here.  We still get 502 gateway timeouts, even with `api_response_time` set in the `v0.65.2` release.  This most recent time, the behavior was:

* Upload a 36MB, 22k file asset
* 502 response after 3.5 min
* After some time, the asset was successfully unzipped and loaded to ES and visible in `/txt/assets` despite the error


Maybe there's another spot where timeouts need to be adjusted, maybe this change doesn't have the effect I desired ...
https://github.com/terascope/teraslice/blob/master/packages/teraslice/lib/cluster/services/assets.js#L172

This whole process should probably be asynchronous after the asset upload completes.  The user should receive an `asset_upload_id` that can be used to track down the asset after it's done completely uploading.  Though ... even so, not blocking until complete does create workflow issues.

peterdemartini:
The elasticsearch request might be timing out when posting the record. I think the default timeout is 2 minutes. We might need to adjust the time for those requests.

godber:
The solution to this problem is to be aware of your assets size and ensure that it doesn't contain too many small files.  Ultimately we implemented the "bundled" asset build to sidestep this problem.  The `teraslice-cli` now builds bundled assets by default and will download bundled assets by default.  One should avoid using our unbundled assets even though we still build and provide them.

Without any further evidence of this problem I think this is resolved.

Issue #1008, opened 2019-02-06
Asset should support node version, teraslice version and OS restrictions


peterdemartini:
Our `asset.json` should be able to optionally restrict the version of node, teraslice, and os/arch it supports. I imagine this should be similar to how the [package.json](https://docs.npmjs.com/files/package.json#engines) works. The versions restrictions need to handle semver [ranges](https://www.npmjs.com/package/semver#ranges). Teraslice should verify these restrictions every time an asset is deployed, or a job is submitted/started.

This metadata should also exist in the assets elasticsearch index as well since we could use this information to find the "latest" asset and prevent potential avoid attempting to load incompatible assets.

jsnoble:
We have this currently, its used to create names and loading of assets

godber:
> We have this currently, its used to create names and loading of assets

This is technically true but teraslice doesn't use these at asset upload time in any way.  I'm not sure this matters at this point thought.

PR #2001, opened 2020-06-18
v0.66.0: Asset improvements

peterdemartini:
- [x] Uploading assets will now happen in the background by default (a `blocking=true` can be used in the query string to block until the asset uploads)
- [x] Add `--blocking` to the `teraslice-cli assets deploy` command
- [x] Specify the timeout on the asset `indexWithId` request (on blocking requests it is the time remaining on the `api_response_timeout`, for non-blocking it is just the full `api_response_timeout`)
- [x] Add the `platform` / `arch` / `node_version` support in the `asset.json` and store it in the elasticsearch (which will be used for finding the asset to use)
- [x] Bump packages

Ref: https://github.com/terascope/teraslice/issues/1008
Ref: https://github.com/terascope/teraslice/issues/1931


packages/teraslice/lib/utils/file_utils.js
```diff
@@ -56,14 +86,15 @@ async function saveAsset(logger, assetsPath, id, binaryData, metaCheck) {
         await decompress(binaryData, newPath);
         logger.trace(`decompressed ${id} to ${newPath}`);
 
-        const metaData = await verifyAssetJSON(id, newPath);
-        logger.trace(`asset ${id} saved to file ${newPath}`);
+        const metadata = await verifyAssetJSON(id, newPath);
+
+        logger.trace(`asset ${id} saved to file ${newPath}`, metadata);
```
godber, line 91: Could you change all these logger.traces to logger.info

Issue #1953, opened 2020-05-27
job start up errors in kubernetes


godber:
We have an internal process that runs approximately 50 teraslice jobs over the course of 7 minutes periodically.  In release [`v0.64.2`](https://github.com/terascope/teraslice/releases/tag/v0.64.2) I changed how jobs start up.  This release and later results in a higher rate of failed job start ups than prior to this release so the internal cluster that runs the periodic workload cannot be upgraded.  I don't have any details beyond that at the moment.

I either need to revert the k8s job start up changes in `v0.64.2`, which successfully eliminated the 2 minute worker startup problem or address the new unknown job startup failures.  I think there are options for the later.

godber:
This problem does not appear to be present in `v0.66.0`.  I will leave this open a little bit, I still think there's something here.  The one problem that I have seen in the `v0.66.0` (as well as other releases is)

```
TSError: slicer for ex 85d2e4... had an error, shutting down execution, caused by Error: No workers have connected to slicer in the allotted time: 10m
    at ExecutionController._terminalError (/app/source/packages/teraslice/lib/workers/execution-controller/index.js:322:23)
    at Timeout.workerConnectTimeoutId.setTimeout (/app/source/packages/teraslice/lib/workers/execution-controller/index.js:965:18)
    at ontimeout (timers.js:436:11)
    at tryOnTimeout (timers.js:300:5)
    at listOnTimeout (timers.js:263:5)
    at Timer.processTimers (timers.js:223:10)
Caused by: Error: No workers have connected to slicer in the allotted time: 10m
    at ExecutionController._startWorkConnectWatchDog (/app/source/packages/teraslice/lib/workers/execution-controller/index.js:949:21)
    at ExecutionController.run (/app/source/packages/teraslice/lib/workers/execution-controller/index.js:251:14)
    at Service.run (/app/source/packages/teraslice/worker-service.js:48:30)
    at Promise.resolve.then.then (/app/source/packages/teraslice/worker-service.js:82:21)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

And from investigating master logs in one of these cases, the master didn't appear to even try to create the Worker deployment.

godber:
This is such ancient news and hasn't repeated to my knowledge, closing.

PR #2019, opened 2020-06-24
Add k8s api retry

godber:
This PR adds retries on all of the calls to the k8s APIs.  Hopefully improving the following things:

* proper job startup when starting many jobs at once
* proper job startup when frequently starting jobs

refs #1953


godber:
@peterdemartini I may want to do some refactoring, but can you take a quick look when you get a chance to sanity check my use of `pRetry`?

packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/utils.js
```diff
@@ -38,4 +39,20 @@ function setMaxOldSpaceViaEnv(envArr, jobEnv, memory) {
...
 
-module.exports = { setMaxOldSpaceViaEnv, makeTemplate, getMaxOldSpace };
+const MAX_RETRIES = isTest ? 2 : 100;
+const RETRY_DELAY = isTest ? 50 : 500;
+
+function getRetryConfig() {
+    return {
+        retries: MAX_RETRIES,
+        delay: RETRY_DELAY,
+        matches: ['es_rejected_execution_exception', 'No Living connections'],
```
peterdemartini, line 49: Unless you know of the specific errors you want to retry on, you might want to remove the `matches` otherwise it will only retry on errors that contain those strings.

godber: Oh, yeah, I completely forgot to go back and look at that ... thanks for calling it out.

godber:
@mkhpalm ... I am going to delete this PR, but I am carrying over my `pRetry` changes (to a new PR) for every API call excluding the `post` and `patch` calls.  I suspect retrying `patch` is OK ... and retrying some errors on `post` is OK.  Other than that the rest of these calls to the k8s API should be idempotent, right?

godber:
I've included the retries here: https://github.com/terascope/teraslice/pull/2074

Closing this one.

PR #2054, opened 2020-07-13
Fix getOperationAPI in teraslice-test-harness

peterdemartini:
We added a `getOperationAPI` method to `teraslice-test-harness` but was return the created API instance which isn't ideal, there is already an API for that, so I changed to correctly return the API class instance.


jsnoble:
Will approve, there was an api that still needed to be exposed, we worked on it and it will be posted. Approving now

Issue #942, opened 2019-01-04
Honor the timeout value specified when stop is called via API - k8s


godber:
When POSTING to `_stop` a job, one can provide a `timeout` as shown here:
https://github.com/terascope/teraslice/blob/master/docs/api.md#post-v1jobsjobid_stop

This is not applied in the case of kubernetes clustering because I had forgotten about this feature of the API.  I suspect it is possible to pass this value through and override the default which gets set based on cluster configuration here:

https://github.com/terascope/teraslice/blob/v0.46.0/packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/deployments/worker.hbs#L65

I suspect the value in the deployment is overridden by calls to the API because `kubectl delete` has the following option:

```
--grace-period=-1: Period of time in seconds given to the resource to terminate gracefully. Ignored if negative.
```

So I suspect it should be possible to pass a graceperiod option of some sort to the delete calls, like this one:

https://github.com/terascope/teraslice/blob/v0.46.0/packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/k8s.js#L200-L201

peterdemartini:
This won't have the indended affect in all cases, if the worker shutdown is set to a timeout greater than the existing worker shutdown timeout specified in the teraslice configuration, the worker will still shut down within the timeout since it does not know of a different timeout.  If the timeout is less than the configurated timeout, the could potentially shutdown non-gracefully, but that risk likely acceptable.

godber:
The behavior you state is fine.  The primary use case for the `timeout` supplied through the API is to shorten the timeout period ... e.g. the user doesn't want to wait the default timeout period for some reason.  So it's OK if the worker doesn't know, at least for the intended use case.  A non-graceful shutdown should be considered a possible consequence of using `timeout`, yes.  For that matter, any shutdown has the potential to be non-graceful if any of the timeouts are shorter than the slice processing time.

godber:
I am going to hold on this request for now.  It seems that setting `gracePeriodSeconds` on a deployment doesn't propagate to the pods by design:

https://github.com/kubernetes/kubernetes/issues/25055
https://github.com/kubernetes/kubernetes/pull/24964

This doesn't mean it's impossible, just that I have to delete pods now.  The more code I write to wrangle k8s objects the more I think I should be using custom controllers/resources.

godber:
[...]

PR #2074, opened 2020-07-23
Add wait for num pods

godber:
This changes the teraslice job shutdown process to be more structured.  Previously we just "simultaneously" deleted the Worker deloyment and Execution Controller Job, which might leave us with the scenario described here:

https://github.com/terascope/teraslice/issues/942#issuecomment-591152677

To mitigate this we now do:

* Delete the worker k8s deployment first
* Wait for all worker pods to be gone
* Delete execution controller k8s job

Note that this is in parallel with the other messaging based shutdown activities, those processes would have already been messaged to shut down, this change is a matter of managing the k8s resources (and by virtue of that, the Teraslice processes).


packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/k8s.js
```diff
@@ -261,18 +299,48 @@ class K8s {
...
+            const err = new Error(`Error deleting workers: ${e}`);
+            this.logger.error(err);
+            return Promise.reject(err);
+        }
+
+        await this.waitForNumPods(
+            0,
+            `app.kubernetes.io/component=worker,teraslice.terascope.io/exId=${exId}`,
+            null,
+            600000 // 10 minutes, I wanted this to be longer than the pod shutdown timeout
```
godber, line 330: The value of `600000` is arbitrary but designed to be longer than the pod shutdown timeout.  Maybe it should use the pod shutdown timeout and add like 30s to it.  Also, I guess this should probably be in a try/catch ... with the catch doing ... `Promise.reject`?

godber:
I have some new changes I intend to make after talking things through with Zach

* add try catch around the wait for pod number
* make the catches around the worker delete and wait for pod number both NOT error (log only) so the execution controller job still gets deleted
* add pRetry on the api call to delete

packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/k8s.js
peterdemartini: Using a try catch around `waitForNumPods` will ignore all errors in that function, it would be safer to add a try catch around just the [k8s request](https://github.com/terascope/teraslice/blob/43a7b182827c7038dd3a5168641ff5750f223702/packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/k8s.js#L118-L119).

godber: I've rethought this behavior ... the critical thing that I want to happen here is that in the event of an error, I want the execution controller to be deleted.  Or at the very least, I want the code to try and do it.  Even if the worker delete fails.

The k8s worker deployment now has an `ownerReference` to the execution controller job such that if the job is deleted, so is the deployment (all the pods already link to the job/deployment).  So I rely on k8s garbage collection when the code encounters an unretryable error.

godber:
This is ready for review again.

packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/utils.js
```diff
@@ -4,6 +4,8 @@ const fs = require('fs');
 const path = require('path');
 const barbe = require('barbe');
 
+const isTest = require('@terascope/utils');
```
peterdemartini, line 7: This should be `const { isTest } = require('@terascope/utils')`

godber: Thanks, fixed.

PR #2092, opened 2020-07-30
v0.69.0 - Asset paths, isEmpty and TS-Transforms bug

jsnoble:
- added ability to set context.sysconfig.teraslice.assets_directory to an array of paths
- fixed isEmpty bug and ignoreWhiteSpace
- fixed bug in ts-transforms for output:false values

### DIFF
--- a/packages/job-components/src/interfaces/context.ts
+++ b/packages/job-components/src/interfaces/context.ts
@@ -20,7 +20,7 @@ export interface TerasliceConfig {
     action_timeout: number|300000;
     analytics_rate: number|60000;
     api_response_timeout?: number|300000;
-    assets_directory?: string;
+    assets_directory?: string[] | string;
     assets_volume?: string;
     cluster_manager_type: ClusterManagerType;
     /** This will only be available in the context of k8s */
--- a/packages/job-components/src/operation-loader.ts
+++ b/packages/job-components/src/operation-loader.ts
@@ -37,10 +37,19 @@ interface ValidLoaderOptions {
 export class OperationLoader {
     private readonly options: ValidLoaderOptions;
     private readonly availableExtensions: string[];
+    private readonly invalidPaths: string[];
+    allowedFile: (name: string) => boolean;
 
     constructor(options: LoaderOptions = {}) {
         this.options = this.validateOptions(options);
         this.availableExtensions = availableExtensions();
+        this.invalidPaths = ['node_modules', ...ignoreDirectories()];
+
+        this.allowedFile = (fileName: string) => {
+            const char = fileName.charAt(0);
+            const isPrivate = char === '.' || char === '_';
+            return !isPrivate && !this.invalidPaths.includes(fileName);
+        };
     }
 
     find(name: string, assetIds?: string[]): string | null {
@@ -326,6 +335,8 @@ export class OperationLoader {
         if (!isString(name)) {
             throw new TypeError('Please verify that the "_op" name exists for each operation');
         }
+
+        if (!this.allowedFile(name)) throw new Error(`Invalid name: ${name}, it is private or otherwise restricted`);
     }
 
     private findCode(name: string) {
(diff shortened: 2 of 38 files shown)


packages/job-components/src/operation-loader.ts
```diff
@@ -362,6 +372,21 @@ export class OperationLoader {
         }
         return path.join(__dirname, '..', '..', 'dist', 'src', 'builtin');
     }
+
+    private validateOptions(options: LoaderOptions): ValidLoaderOptions {
+        const clone = cloneDeep(options);
```
peterdemartini, line 377: This can be simplified a bit using `castArray` 

```ts
        const clone = cloneDeep(options);
        clone.assetPath = castArray(clone.assetPath);
        return assetPath;
```

packages/job-components/test/job-validator-spec.ts
```diff
@@ -4,7 +4,7 @@ import { JobValidator, TestContext, JobConfig } from '../src';
 
 describe('JobValidator', () => {
     const context = new TestContext('teraslice-operations');
-    context.sysconfig.teraslice.assets_directory = __dirname;
+    context.sysconfig.teraslice.assets_directory = [__dirname];
```
peterdemartini, line 7: We shouldn't change existing tests to use the array version, we should add additional ones to work with an array of asset directories.

packages/job-components/src/test-helpers.ts
```diff
@@ -188,7 +188,7 @@ export class TestContext implements i.Context {
             teraslice: {
                 action_timeout: 10000,
                 analytics_rate: 10000,
-                assets_directory: path.join(process.cwd(), 'assets'),
+                assets_directory: [path.join(process.cwd(), 'assets')],
```
peterdemartini, line 191: We should keep this as just a string so we don't break places that expect this to be a string.

packages/job-components/src/register-apis.ts
```diff
@@ -105,10 +107,12 @@ export function registerApis(
         },
     });
 
-    const assetDir = context.sysconfig.teraslice.assets_directory;
+    const configAssetsDir = context.sysconfig.teraslice.assets_directory || [];
+    const assetDir = Array.isArray(configAssetsDir) ? configAssetsDir : [configAssetsDir];
```
peterdemartini, line 111:
```suggestion
    const assetDir = castArray(context.sysconfig.teraslice.assets_directory);
```

packages/job-components/src/register-apis.ts
```diff
@@ -105,10 +107,12 @@ export function registerApis(
...
+
     context.apis.registerAPI('assets', {
         getPath(name: string): Promise<string> {
-            return getAssetPath(assetDir || '', assetIds || job.assets, name);
+            return getAssetPath(assetDir || [], assetIds || job.assets, name);
```
peterdemartini, line 115: We now this will be an array
```suggestion
            return getAssetPath(assetDir, assetIds || job.assets, name);
```

packages/job-components/src/interfaces/context.ts
```diff
@@ -20,7 +20,7 @@ export interface TerasliceConfig {
     action_timeout: number|300000;
     analytics_rate: number|60000;
     api_response_timeout?: number|300000;
-    assets_directory?: string;
+    assets_directory?: string[] | string;
```
peterdemartini, line 23: This might need a schema change to be able to work right

packages/job-components/src/operation-loader.ts
```diff
@@ -50,7 +57,10 @@ export class OperationLoader {
             });
         };
 
-        findCodeByConvention(this.options.assetPath, assetIds);
+        for (const assetPath of this.options.assetPath || []) {
```
peterdemartini, line 60:
```suggestion
        for (const assetPath of this.options.assetPath) {
```

Issue #2141, opened 2020-08-19
Teraslice Execution Controller Job should be cleaned up on scheduling failure


godber:
I was looking at https://github.com/terascope/teraslice/issues/2087 when I realized that if you start the `example-job-targets.json` without adding a label to the minikube node the following will happen:

* The teraslice job will ultimately become `failed` with `._failureReason` (within about 3 min)

```
TSError: Failured to provision execution 59e28951-b3e6-40ff-9516-9cc473b84bd3, caused by Error: Timeout waiting for pod matching: controller-uid=4d4be541-daee-44b0-b3ce-87301d07100b
    at Timeout.allocator [as _onTimeout] (/app/source/packages/teraslice/lib/cluster/services/execution.js:337:31)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
Caused by: Error: Timeout waiting for pod matching: controller-uid=4d4be541-daee-44b0-b3ce-87301d07100b
    at K8s.waitForSelectedPod (/app/source/packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/k8s.js:103:34)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Object.allocateSlicer (/app/source/packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/index.js:106:21)
    at async Timeout.allocator [as _onTimeout] (/app/source/packages/teraslice/lib/cluster/services/execution.js:322:29)
```

However, the k8s Job for the execution controller will hang around indefinitely:

```
kubectl get --namespace ts-dev1 pods,deployments,jobs -o wide                                                                                                                           badwater.local: Wed Aug 19 16:45:53 2020

NAME                                                                READY   STATUS    RESTARTS   AGE   IP           NODE       NOMINATED NODE   READINESS GATES
pod/elasticsearch-d57c598fc-dqks2                                   1/1     Running   0          17m   172.17.0.3   minikube   <none>           <none>
pod/teraslice-master-57b6b9b44d-jszdk                               1/1     Running   0          17m   172.17.0.4   minikube   <none>           <none>
pod/ts-exc-example-data-generator-job-targets-75b19c32-98d4-k7lbf   0/1     Pending   0          16m   <none>       <none>     <none>           <none>

[...]

Issue #856, opened 2018-10-16
`active` attribute on jobs


kstaken:
We should add a `monitor` boolean attribute to jobs. The primary usage of this would be to indicate to external monitoring systems what jobs should be monitored but it would also be useful for certain internal uses too.

In particular there should be a way to hit an API endpoint and get a list of all jobs listed as `monitor`able and their actual status on the cluster. 

Details of this will need a little thought but the driving use case for this is our inability to currently see what's 'supposed' to be running. As we've added more and more jobs, temporary and test jobs, jobs moving between clusters and jobs being replaced with new versions we need a better way to indicate a job is `important` to watch.

godber:
Our discussion last night about how `tjm` stores cluster metadata on the job json files and how that feels like data duplication to me made me think of this ticket.  Though I hadn't remembered that you specifically chose the attribute name to be `monitor`.  I remembered it as `active`.

Rather than storing anything in the json file, assuming `monitor` is properly managed, `tjm` could just query all clusters, find which cluster has the job with the `monitor` property set.  Then perform it's action on that job on that cluster.  This would mean there is no state duplication that can become out of sync. `tjm` would be using the source of truth, the cluster state of all of the teraslice clusters.

Maybe it's not necessarily a given the cluster states of all of the teraslice clusters is the source of truth.  Perhaps that's what's up for debate.  But the distributed teraslice cluster states is definitely A source of truth.  It's definitely what's happening now.

godber:
Then there's the question as to whether the `monitor` property as described by Kimbro in this ticket and `active` property as I've mentioned in the comment above are the same thing.  This made me wonder if instead of properties, we should provide a `tags` property that allows us to define multiple things along this line: `tag: [monitor, active]` ... teraslice could reserve certain tags for its use but allow for other tags (maybe/probably on another property) to be defined by users.  Maybe too messy.

godber:
I think if we were to have one boolean property it would be `inactive`.  I think this matches the existing workflow the best.  You create a job, it gets registered, used for a while, then some day, that pipeline is replaced and that job is no longer used.  It's inactive, if we could delete jobs we might delete that one since we know it won't get used again.  @ciorg might even mimic this behavior in the git repo he stores jobs in ... it looks like he has a `legacy/` subdirectory that he moves disused jobs to.

Above I mentioned tags ... we now have labels:

https://github.com/terascope/teraslice/issues/1546#issuecomment-566587947

godber:
[...]

PR #2142, opened 2020-08-20
Teraslice v0.70.0 - k8s resource cleanup, add active property

godber:
This PR does the following:

* cleans up k8s resource from jobs that fail to schedule - refs #2141
* adds `active` property to teraslice jobs - refs #856


godber:
Assuming these tests pass, this is ready for review.

PR #2174, opened 2020-09-04
Add MinIO and linkToMain to ts-scripts

peterdemartini:
- Adds MinIO service support to ts-scripts
- Adds `linkToMain` package.json config to keep version in sync with main package


jsnoble (approved):
LGTM

PR #2182, opened 2020-09-08
Add DataFrame to data-mate

peterdemartini:
### Goal

Add a columnar `DataFrame` that uses the `DataType` config. `DataFrame` should be immutable where possible, or at least the API should be constrained for mutating them, because the columns indices will need to be kept in-sync.

- [x] Add a basic Data Frame / Data Table
- [x] Add iterator support for a column/row
- [x] Add transform support for a column
- [x] Add reduce support for a column
- [x] Add rename column support
- [x] Add distinct value support for a column
- [x] Add orderBy to DataFrame
- [x] Add filterBy support to DataFrame
- [x] Add Aggregation support
- [x] Add a few more transforms and validators
- [x] Figure out how to store dates correctly
- [x] Add/update some basic docs
- [x] Bump packages

Deferred:

- Join support (this can be done externally)
- Higher level APIs on the DataFrame
- xLucene searching on a DataFrame (it is complicated and we don't need it yet)
- Integration into ts-transforms
- Vector data structure memory optimization
- Performance tuning / benchmarks
- Add more transformations and validators
- Add locale support


peterdemartini:
@kstaken I added some rough [DataFrame API docs](https://github.com/terascope/teraslice/blob/feature/add-data-frame/docs/packages/data-mate/data-frame.md) but it is still missing examples.

peterdemartini:
I am marking this ready for review but I still need to bump the packages

peterdemartini:
I am going to go ahead and merge this. We can make more changes in the future.

PR #2203, opened 2020-09-18
Fix time series index support in elasticsearch-store

peterdemartini:
- Fix time series support in elasticsearch-store
- Add `writeIndex` to  `IndexStore`
- Add `aggregate` function to `IndexStore`
- Add the ability to pass in more configuration options from `IndexModelConfig`
- Add `createBulkRecord` to `IndexModel`
- BREAKING: `indexQuery` is now `searchIndex` to  `IndexStore`
- BREAKING: The first argument to `getDefaultParams` in `IndexStore` is now the index to use
- Fix spelling issues

### DIFF
--- a/packages/elasticsearch-store/src/index-manager.ts
+++ b/packages/elasticsearch-store/src/index-manager.ts
@@ -32,6 +32,11 @@ export default class IndexManager {
         });
     }
 
+    /**
+     * Format the current index name.
+     *
+     * @param useWildcard if true a wildcard is added to the end of the end the index name
+    */
     formatIndexName<T = any>(config: IndexConfig<T>, useWildcard = true): string {
         utils.validateIndexConfig(config);
 
@@ -53,6 +58,10 @@ export default class IndexManager {
         return indexName;
     }
 
+    /**
+     * Format the template name, similar to formatIndexName except it excludes
+     * template wildcards and the time series parts of the index name.
+    */
     formatTemplateName<T = any>(config: IndexConfig<T>): string {
         utils.validateIndexConfig(config);
 
--- a/packages/elasticsearch-store/src/index-manager.ts
+++ b/packages/elasticsearch-store/src/index-manager.ts
@@ -6,7 +6,7 @@ import { IndexConfig, MigrateIndexOptions } from './interfaces';
 const _loggers = new WeakMap<IndexConfig<any>, ts.Logger>();
 
 /**
- * Manage Elasticsearch Indicies
+ * Manage Elasticsearch Indices
  */
 export default class IndexManager {
     readonly client: es.Client;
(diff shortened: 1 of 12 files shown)


packages/elasticsearch-store/src/index-model.ts
```diff
@@ -169,6 +182,24 @@ export default abstract class IndexModel<T extends i.IndexModelRecord> extends I
         return count === ids.length;
     }
 
+    /**
+     * Create a bulk record
+     */
+    async createBulkRecord(record: i.CreateRecordInput<T>): Promise<void> {
```
jsnoble, line 188: is `createBulkRecord` the best name for this function? Beyond creating a bulk formatted record it actually makes a bulk request as well. The name more implies the formatting and not the sending aspect of its functionality

packages/elasticsearch-store/src/index-manager.ts
```diff
@@ -58,11 +58,18 @@ export default class IndexManager {
 
         const { name, namespace } = config;
 
-        return utils.formatIndexName([namespace, name, utils.getDataVersionStr(config)]);
+        const indexName = utils.formatIndexName([
+            namespace, name, utils.getDataVersionStr(config)
+        ]);
+
+        if (utils.isTemplatedIndex(config.index_schema)) {
```
jsnoble, line 65: does this line need the `useWildcard` check ? Is there a difference in this line vs line 49 of the same file? It would be a little confusing to look at this and not know why it does different things when it looks so similar

jsnoble (changes requested):

packages/elasticsearch-store/src/index-model.ts
peterdemartini: I think that is a valid point but I can't think of a better name, unless we choose `createRecordWithBulkRequest` (which still is completely accurate and is kind of long).  Also `elasticsearch-store` has its own high level abstraction for `bulk` requests, developers using this library won't ever being worrying about how it formats the elasticsearch bulk request.

Does that make sense? Do you have a better recommendation for a name?

packages/elasticsearch-store/src/index-manager.ts
peterdemartini: It is a little confusing, there is slightly different behavior but there might be a better solution

packages/elasticsearch-store/src/index-model.ts
jsnoble: It seems like the formatting of the record is not the main point, it's more of an internal helper so should we put more focus on its main action which is sending the record. It also seems to only be working on a individual record so what if we called it `sendRecord` or something along the lines of `indexRecord`?

packages/elasticsearch-store/src/index-manager.ts
jsnoble: It's probably fine that it is different but there is no context, could we have some comments to explain? From a top level view it looks like its doing the same as line 49 almost word for word but there are no notes to explain why there is a slight difference or why this is special. I would have to dig deep into the code to understand the nuance about why it is slightly different

peterdemartini: I can add a comment to it. I was able to remove those lines entirely.

packages/elasticsearch-store/src/index-model.ts
peterdemartini: I don't think those improve the naming and make it more clear. The intention is to have an API like createRecord that will run in the request in the bulk queue. You're right that the name, maybe enqueueCreateRecord?

jsnoble (approved):

jsnoble:
LGTM assuming tests pass

PR #2258, opened 2020-10-19
elasticsearch-api remove type from es7 queries and bulk requests

ciorg:
* removes `type` from queries and bulk request
* expanded tests
* bumped version to 0.21.3


peterdemartini (approved):
LGTM

PR #2259, opened 2020-10-19
Fix xLucene field group issue

peterdemartini:
Fixes an issue where a field group wasn't properly handled in xLucene. For example, `foo: (bar baz)` thought the value was `bar baz` and not a `bar` or `baz`.


jsnoble (approved):
LGTM

PR #2272, opened 2020-10-21
Change the xLucene AST to be more declarative

peterdemartini:
Remove automagical variable validation and coercion from the xLucene parser/ast and push it down to the execution phase. This will make it easier to build on top of xlucene.


jsnoble (approved):
LGTM assuming tests pass

PR #2288, opened 2020-10-27
teraslice-test-harness downloads and includes external assets

ciorg:
* downloads github stored assets to ./test/.cache/assets
* incorporates dowloaded assets into the test assets directory
* added tests for downloadExternalAsset
* updated documentation
* bumped to v0.27.0


packages/teraslice-test-harness/src/utils.ts
```diff
@@ -33,3 +33,12 @@ export function isAssetDirRoot(assetDir: string): boolean {
     const assetJson = path.join(assetDir, 'asset', 'asset.json');
     return fs.existsSync(assetJson);
 }
+
+export function externalAssets(): string {
+    if (__dirname.includes('node_modules')) {
+        return path.join(__dirname.slice(0, __dirname.indexOf('node_modules')), 'test', '.cache');
```
peterdemartini, line 39: What does this do?

packages/teraslice-test-harness/src/download-external-asset.ts
```diff
@@ -0,0 +1,137 @@
...
+        try {
+            await decompress(zippedAssetPath, unzippedAsset);
+        } catch (err) {
+            throw new Error(`Error unzipping asset: ${assetInfo.asset_string}: ${err}`);
+        }
+    }
+
+    private _filterReleaseFunc(version: string | undefined) {
+        if (version) {
+            return (release: any) => release.tag_name.includes(version);
```
peterdemartini, line 72: If the tag name was 'v1.2.10' and the version is 'v1.2.1', it would match when it shouldn't have.

packages/teraslice-test-harness/src/download-external-asset.ts
```diff
@@ -0,0 +1,137 @@
...
+        this._ensureDirExists(this.zipped_path);
+
+        const [zippedAsset] = await this._downloadAssetZip(assetInfo);
+
+        return zippedAsset;
+    }
+
+    private async _downloadAssetZip(assetInfo: I.AssetInfo): Promise<string[]> {
+        try {
+            const result = downloadRelease(
```
peterdemartini, line 42:
```suggestion
            const result = await downloadRelease(
```

peterdemartini (changes requested):

packages/teraslice-test-harness/src/utils.ts
ciorg: Function to get the path to the .cache  directory.

The test dir is in the same dir as the node_modules, so it gets the path to that level, then adds the 'test/.cache'.  Except when it is itself being tested then that method can't be used to find the test directory.

packages/teraslice-test-harness/src/download-external-asset.ts
ciorg: good point

packages/teraslice-test-harness/src/download-external-asset.ts
ciorg: yeah

packages/teraslice-test-harness/src/utils.ts
peterdemartini: That won't work with some module setups (like in a monorepo), I think using `path.resolve('./test/.cache')` would be sufficient?

peterdemartini (changes requested):

packages/teraslice-test-harness/src/utils.ts
ciorg: except path.resolve won't work if it's the first time it's ran, because .cache doesn't exist yet.

ciorg: maybe it's just better to use the .teraslice dir

peterdemartini:
LGTM, don't forget to bumper

peterdemartini (approved):

PR #2309, opened 2020-11-12
Misc DataFrame Improvements

peterdemartini:
- [x] Fix issue with Object builder and null values
- [x] Add IPValue so IPs can be sorted
- [x] Add toArray to Vector
- [x] Move static valueFrom and valueToJSON to methods to make it easier
- [x] Improve Object data type performance
- [x] Add a Tuple type
- [x] Add unique to column that behaves like: `[1,2,3,2,3]` => `[1,2,3, null, null]`
- [x] Add `DataFrame.require(fields: string[])`
- [x] Add more transform/validators
- [x] Improvements to `@terascope/utils` and add more functional programming functions that make this more composable


types/ip-bigint/index.d.ts
```diff
@@ -0,0 +1,10 @@
+export interface ParsedIP {
+    number: bigint;
+    version: 4|6;
+    ipv4mapped?: any;
```
jsnoble, line 4: Is there a reason why the key values here are not camel-cased?

jsnoble (changes requested):
Besides the interface question, I see that in some places you replaced TSError with just regular Error while in others you preserved it. What was the reason? Is there some issue with returning a TSError in those spots? How can we consistly decide when to do one or the other, or why can't we just always use TSError?

types/ip-bigint/index.d.ts
peterdemartini: These are typescript definitions for the `ip-bigint` library, I am not in-control of the casing.

peterdemartini:
> Besides the interface question, I see that in some places you replaced TSError with just regular Error while in others you preserved it. What was the reason? Is there some issue with returning a TSError in those spots? How can we consistly decide when to do one or the other, or why can't we just always use TSError?

Using a custom error can cause *slightly off* results in the stack traces, particularly when using the `--enable-source-maps` flag (I am not sure why I can't figure it out). So I replaced it where it wasn't adding any value and didn't use any of the `TSError` options or error parsing. We should only be using `TSError` where adds value, otherwise it's just extra overhead and makes dealing with it a little more difficult.

jsnoble (approved):

PR #2334, opened 2020-11-30
Fix DataFrame->limit and other fixes

peterdemartini:
- Fix issue with DataFrame.limit
- Add DataFrame->toArray
- Fix unnecessary call to Object.entries() in getRow


jsnoble (approved):
LGTM

PR #2335, opened 2020-12-02
Fix dataframe count with null values

peterdemartini:
DataFrames `count` aggregations will counting null values, this is incorrect as it breaks unique value counts


jsnoble (approved):
LGTM

PR #2336, opened 2020-12-03
Fix valueToJSON serialization in ObjectVector

peterdemartini:
- Fixes `valueToJSON` serialization in `ObjectVector`
- Add `and` and `or` fp util
- Improve `not` fp util to accept multiple fns (like `and` and `or`)


jsnoble (approved):
LGTM

PR #2337, opened 2020-12-04
Improve xLucene geo shape validation

peterdemartini:
- Improves error message geo shape input
- Fixes error when given an invalid geo function
- Adds `forEachTermNode` to xlucene Parser


jsnoble (approved):
LGTM

PR #2410, opened 2021-01-08
Change teraslice-cli asset zip to follow symlinks

peterdemartini:
Changes teraslice-cli zip command to follow symlinks which will improve compatibility with monorepos


jsnoble (approved):
LGTM, need to resolve confilicts before being brought in

PR #2421, opened 2021-01-11
Add xpressions package

peterdemartini:
Adds a variable expressions package called `xpressions`, which can used to parse strings with templated variables and evaluate the contents of the templated variables.

- [x] Add the boilerplate package
- [x] Build out the `parse` function for validating the query and variables before executing
- [x] Build out the `evaluate` function for replacing the string with a variable, eventually this will support more complex expressions
- [x] Build out the `transform` function which will be used to validate and replace the templated variables with result of the evaluated expressions
- [x] Support for scoped variables


jsnoble (approved):
LGTM

PR #2422, opened 2021-01-13
Better error in DataFrame.concat

### LINKED ISSUE #2388: DataFrame.concat() will not add Columns
- DataFrame is having issues concatenating Columns, it can take DataFrame.toJSON(), but it incorrectly add columns resulting in wrong size. (An empty DataFrame was first created and we concatenated into it.
-  I am not sure if this other part is a problem, but trying to concat DataFrame.toArray() caused it to break on IP fields.

peterdemartini:
- Fixes #2388 since it was only issue because there were multiple copies of data-mate being used.
- Adds additional tests for DataFrame.concat
- Improves some of the debug info

### DIFF
--- a/packages/data-mate/package.json
+++ b/packages/data-mate/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@terascope/data-mate",
     "displayName": "Data-Mate",
-    "version": "0.19.4",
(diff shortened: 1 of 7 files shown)


jsnoble (approved):
LGTM

PR #2423, opened 2021-01-15
Allow nil values in xLucene ORs and xpression improvements

peterdemartini:
- Allow nil values in xLucene variables when used in the context of an OR variable
- Change xpressions to support explicit variables using $
- Improve xpression API to be more composable


jsnoble (approved):
LGTM

PR #2439, opened 2021-01-21
DataFrame nested object improvements

peterdemartini:
- Fix `IPValue`s to gracefully handle multiple version of the data-mate
- Add support for `skipNullFields` when calling `toJSON` to cleanup nil values
- Adds `DataFrame.deepSelect` for selecting nested properties or fields matching a wildcard
- Fixes an issue with `wildCardToRegex` query
- Fixes object deduplication in data frames
- Better serialization options


jsnoble (approved):
LGTM

PR #2454, opened 2021-01-25
v0.72.1 - Serialize asset unzipping and update e2e tests

peterdemartini:
- Support the latest elasticsearch bundle and use the data_generator from standard assets
- Improve autoloading to not overload the elasticsearch node in CI
- Only allow one zip file to be unzipped at at time
- Fixes to the teraslice-cli asset build command related to symlinks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 dist: bionic
 os: linux
 language: node_js
-node_js: '12.19'
+node_js: '12.20'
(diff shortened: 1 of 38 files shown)


peterdemartini:
Assigning this to @jsnoble to fix, see https://github.com/terascope/elasticsearch-assets/issues/613

jsnoble (approved):
LGTM

Issue #2438, opened 2021-01-21
Check returned job list for empty list


godber:
The k8s API could return an empty list here:

https://github.com/terascope/teraslice/blob/d11b5fce61f33198d531e776ec4269dcb1932341/packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/index.js#L129-L132

We should retry a few times if that is the case, then raise an error if it is retried too many times.  I think it's possible for this to return an empty list even when the execution controller is running, but the state in k8s is not updated so the list is not yet populated.

peterdemartini:
@godber  I think we can close, this right?

godber:
Yeah

PR #2455, opened 2021-01-26
fixing issue when getting job list from k8s

godber:
This change will retry in the event that it gets a job list from k8s that contains anything other than a single entry.  This is meant to address:

https://github.com/terascope/teraslice/issues/2438


packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/k8s.js
```diff
@@ -197,6 +197,19 @@ class K8s {
...
     }
 
+    async nonEmptyList(selector, objType) {
+        const jobs = await this.list(selector, objType);
+        if (jobs.items.length === 1) {
+            return jobs;
+        } if (jobs.items.length === 0) {
+            const msg = `Teraslice ${objType} matching the following selector was not found: ${selector} (retriable)`;
+            this.logger.warn(msg);
+            throw Error(msg);
```
peterdemartini, line 207: We should probably explicitly set `retryable: true` on error since `pRetry` is kind of specific about the errors it will retry.

```suggestion
            throw TSError(msg, { retryable: true });
```

packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/k8s.js
```diff
@@ -197,6 +197,19 @@ class K8s {
...
+        } else {
+            throw Error(`Unexpected number of Teraslice ${objType}s matching the following selector: ${selector}`);
```
peterdemartini, line 209: We should probably explicitly set `retryable: true` on error since `pRetry` is kind of specific about the errors it will retry.

```suggestion
            throw TSError(`Unexpected number of Teraslice ${objType}s matching the following selector: ${selector}`, {
                retryable: true
            });
```

packages/teraslice/test/lib/cluster/services/cluster/backends/kubernetes/k8s-spec.js
```diff
@@ -104,6 +104,33 @@ describe('k8s', () => {
...
+                .get('/apis/batch/v1/namespaces/default/jobs/')
+                .query({ labelSelector: 'app=teraslice' })
+                .reply(200, { items: [] });
+
+            try {
+                await k8s.nonEmptyList('app=teraslice', 'jobs');
+            } catch (error) {
+                expect(error).toEqual(
+                    Error('Teraslice jobs matching the following selector was not found: app=teraslice (retriable)')
+                );
```
peterdemartini, line 129: If this resolves instead of rejects, the test will pass, jest has an API for handling this more gracefully too.

```suggestion
            await expect(k8s.nonEmptyList('app=teraslice', 'jobs')).rejects.toThrowError('Teraslice jobs matching the following selector was not found: app=teraslice (retriable)');
```

peterdemartini (changes requested):

packages/teraslice/test/lib/cluster/services/cluster/backends/kubernetes/k8s-spec.js
godber: Oh, good point, it hadn't occurred to me.

I tried something like that to begin with but couldn't get it working right.

I will try your suggestion and see if it works.

godber: Ok, I think I have it now.

peterdemartini (approved):

PR #2456, opened 2021-01-27
Ip routable fix

ciorg:
date-mate field validator was not checking ips in the loopback range

since both isRoutableIp and IsNonRoutableIP shared basically the same logic I extracted the publicIP check to it's own function.  Not sure if that is the best place for it.  I checked in @terascope/utils but didn't see an obvious place there.

If it should go somewhere else let me know

Added tests to include all the private ip ranges and loopback range

bumped data-mate to 0.20.2


peterdemartini (approved):
LGTM

PR #2471, opened 2021-02-03
AggregationFrame/DataFrame performance improvements

peterdemartini:
- [x] Improves `date` aggregation performance by avoiding call to date-fns format
- [x] Removes unnecessary calls during aggregation
- [x] Adds `EventLoop` to `utils` to make easier to avoid blocking the event loop with long running synchronous tasks.
- [x] Adds `BigSet` to `utils` to allow creating sets with more than 16 millions keys.
- [x] Add `DataFrame->appendAll` method for creating a single data frame from many


jsnoble (approved):
LGTM

PR #2472, opened 2021-02-03
Add support for DataFrame->search(query)

peterdemartini:
- [x] Refactor the xLucene types to not due unions which are better for typescript
- [x] Add support for `DataSearch->search(query)` that supports xLucene queries 
- [x] Verify that search function works with IP, Date, and object values.


peterdemartini:
I ended up just re-using a lot of document matcher logic since we still have to have move around the data in order to do a filter. I know there is probably a way to make it operate like more efficiently but now is the time to rethink this implementation.

jsnoble (approved):
LGTM

PR #2486, opened 2021-02-10
v0.72.0 - Asset schema change

jsnoble:
- change elasticsearch schema of the asset mapping, changed blob keyword => binary


peterdemartini (approved):
LGTM

Issue #2487, opened 2021-02-10
Add support for bundled assets


peterdemartini:
Some of our assets are growing fairly large (ranging from 10-30MB), this puts strain on a elasticsearch and the file system since many times there can be a 100K of files in a given repo (which 90% aren't being used). We need a way to utilize tree shaking to get the asset bundles down in size.

In order to get this to work, we each asset bundle will need to have an index file that exports something like:

```js
module.exports = {
    /** A list of the operations provided by this asset bundle */
    ASSETS: {
        /** The key here would be the current file name */
        my_processor: {
            Processor: MyProcessor,
            Schema: MyProcessorSchema,
        },
        my_reader: {
            Fetcher: MyFetcher,
            Slicer: MySlicer,
            Schema: MyReaderSchema,
        },
        my_api: {
            API: MyAPI,
            Schema: MyAPISchema,
        },
    },
}
```

After some experimenting I think using [esbuild](https://esbuild.github.io/getting-started/) will be our best bet for a bundler, we can bundle using this command: `esbuild asset/src/index.ts --bundle --platform=node --target=node12.20 --outdir=asset`. Doing this I was able to reduce the size of an asset bundle from 23MB to 1.9MB.

We will need to consider how to deal with backwards compatibility, for now we should consider naming the zip file differently and keep building the old asset bundles so that way we more easily migrate support both.

Most of the work here should be handled in teraslice-cli and job-components should know how read both a bundled and unbundled asset directory.

PR #2488, opened 2021-02-11
WIP: op-loader to load new style of assets

jsnoble:
- allow loading of assets in the format listed in https://github.com/terascope/teraslice/issues/2487


peterdemartini (changes requested):

packages/job-components/src/operation-loader.ts
```diff
@@ -34,6 +38,105 @@ interface ValidLoaderOptions {
...
+    asset = 'asset',
+    /** is located in node_modules */
+    module = 'module',
+    /** is located in builtin dir */
+    builtin = 'builtin',
+    /** is located natively in teraslice */
+    'teraslice' = 'teraslice'
+}
+
+export const V3TypeDict = {
```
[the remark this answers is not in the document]
jsnoble: This is used to get the correct spelling of the key for the given type for the bundled asset repository
api => API
processor => Processor

```
export default {
    /** A list of the operations provided by this asset bundle */
    ASSETS: {
        /** The key here would be the current file name */
        v3_processor: {
            Processor: AssetProcessorExampleBatch,
            Schema: AssetProcessorSchema,
        },
        v3_reader: {
            Fetcher: AssetExampleFetcher,
            Slicer: AssetExampleSlicer,
            Schema: AssetReaderSchema,
        },
        v3_api: {
            API: AssetExampleAPI,
            Schema: AssetSchemaAPI,
        },
    },
};
```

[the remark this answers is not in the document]
peterdemartini: Is there are reason why we can't just use the Pascal case version instead of creating a lowercased version?

packages/job-components/src/operation-loader.ts
```diff
@@ -34,6 +36,101 @@ interface ValidLoaderOptions {
...
+    processor: 'Processor',
+    schema: 'Schema',
+    fetcher: 'Fetcher',
+    slicer: 'Slicer',
+    api: 'API',
+    observer: 'Observer'
+};
+
+interface BaseMetadata {
```
peterdemartini, line 99: It doesn't look like this interface, or ones that extend it, are being used?

peterdemartini (changes requested):
Beyond the unresolved comment and this new comment, I believe this might need to bumped again. Also CI doesn't seem to be passing.

packages/job-components/src/operation-loader/interfaces.ts
```diff
@@ -0,0 +1,150 @@
...
+    [AssetRepositoryKey.processor]: ProcessorConstructor,
+    [AssetRepositoryKey.schema]: SchemaConstructor,
+    [AssetRepositoryKey.api]?: APIConstructor,
+}
+
+export interface BundledReaderOperation {
+    [AssetRepositoryKey.fetcher]: FetcherConstructor,
+    [AssetRepositoryKey.slicer]: SlicerConstructor,
+    [AssetRepositoryKey.schema]: SchemaConstructor,
+    [OperationTypeName.api]?: APIConstructor
```
peterdemartini, line 104:
```suggestion
    [AssetRepositoryKey.api]?: APIConstructor
```

packages/job-components/src/operation-loader/interfaces.ts
```diff
@@ -0,0 +1,150 @@
...
+}
+
+export interface BundledAPIOperation {
+    [OperationTypeName.api]: APIConstructor
```
peterdemartini, line 108:
```suggestion
    [AssetRepositoryKey.api]: APIConstructor
```

packages/job-components/src/operation-loader/loader.ts
```diff
@@ -52,83 +50,140 @@ export class OperationLoader {
...
     }
 
-    find(name: string, assetIds?: string[]): string | null {
+    private _getBundledRepository(dirPath: string): Record<string, any>|undefined {
+        const asset = this.require(dirPath);
+        return get(asset, ASSET_KEYWORD) ?? get(asset, `default.${ASSET_KEYWORD}`);
+    }
+
+    private _getBundledOperation<T>(
+        dirPath: string, opName: string, type: string, shouldThrow = true
```
peterdemartini, line 59:
```suggestion
        dirPath: string, opName: string, type: OperationTypeName, shouldThrow = true
```

peterdemartini (changes requested):
Getting close, before we merge it though we should manually test it on one our asset bundles

peterdemartini (approved):
LGTM, assuming we manually test this on our asset bundles

PR #2504, opened 2021-02-17
Add hook support to scripts

peterdemartini:
- Add hook to support to ts-scripts to extend the release process
- Ensure using `.mjs` files works with our eslint-config


jsnoble (approved):
LGTM

PR #2506, opened 2021-02-18
Add limit to DataFrame.appendAll

peterdemartini:
- Add support to limit in appendAll so we can avoid extra overhead
- Fix getColumnOrThrow error message


jsnoble (approved):
LGTM

PR #2518, opened 2021-02-22
v0.73.1 - Make sure stopping jobs get set to stopped

peterdemartini:
Sometimes in development an execution gets stuck in stopping status since the process gets force killed in before it can be updated to stopped. Now the execution will be marked as stopped when starting up the execution service. I didn't add tests because it is not really testable. 

I also fixed an error message in the s3 connector

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "teraslice-workspace",
     "displayName": "Teraslice",
-    "version": "0.73.0",
+    "version": "0.73.1",
     "private": true,
     "homepage": "https://github.com/terascope/teraslice",
     "bugs": {
--- a/packages/terafoundation/package.json
+++ b/packages/terafoundation/package.json
@@ -1,7 +1,7 @@
 {
     "name": "terafoundation",
     "displayName": "Terafoundation",
-    "version": "0.30.2",
+    "version": "0.30.3",
     "description": "A Clustering and Foundation tool for Terascope Tools",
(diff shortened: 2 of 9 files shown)


godber:
Peter and I discussed this and concluded that handling jobs stuck in stopping is a good idea, but in the initial PR that was only done on Teraslice Master start up.  Realistically, this should be handled periodically by the master.  Note that this only changes the ES state for the job and doesn't do anything on the Native/k8s clustering side, it assumes that those components have done there job correctly (which is usually the case).

Peter is going to spend a few more minutes to see if he can move this to a periodic reaper or something.

In the meantime, people who need to stop jobs stuck in stopping can use curl to force a job into stopped so it can be restarted:

```
curl -Ss localhost:5678/jobs/<JOBID>/_stop -XPOST
```

Teraslice CLI tries to be "helpful" and I think it prohibits you from stopping a job already in stopping.  Either that, it doesn't have access to the right state to make the right decision.

godber:
We should probably get feedback from @jeffmontagna and @ciorg about the tjm change.  Make sure it won't impact any of their use cases?  Maybe instead it needs an option that forces the `stopping` -> `stopped` transition.

jeffmontagna:
tjm change looks good to me.

ciorg:
LGTM - not sure if @godber wants to approve the rest

godber (approved):
Looks good!

PR #2522, opened 2021-02-24
Fixes to ts-scripts syncing and hooks

peterdemartini:
- [x] Fix execute hook to respect `--quiet`
- [x] Fix syncing to correctly update the tsconfig
- [x] Fix syncing to update and use versions from the root package


jsnoble (approved):
LGTM

PR #2536, opened 2021-03-05
data-mate and utils update

ciorg:
* fixed false positive bug in isBase64 field validation
* fixed inNumberRange bug where if 0 is min or max sets limit to infinity
* bumped utils version to  0.34.2
* bumped data-mate version to 0.25.1


peterdemartini (approved):
LGTM, ideally we'd get the base64 code change in validator-js BUT I am not sure if this solution would be accepted since Buffer isn't available natively on the web.

PR #2588, opened 2021-03-25
Change vectors to have multiple data buckets

### LINKED ISSUE #2587: DataFrame columns should be able to store multiple Vectors
In order to improve incremental building and allow storing more data in a single column. A Column should be able to reference one or more Vectors, instead of just one. This will make `DataFrame.appendAll(frames)` and `DataFrame.concat(columns)` avoid rebuilding the whole Vector just to append.

peterdemartini:
- Change vectors to have multiple data buckets so `DataFrame.appendAll` is just moving references to memory around 
- Improve performance of Aggregations significantly by not moving around as much memory

Closes #2587


jsnoble (approved):
LGTM

PR #2590, opened 2021-03-25
Ip validation upgrade

ciorg:
* changes to validation ip to allow for ipv4 mapped ipv6 addresses
* more thorough routable ip logic based on:
 * https://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml
 *  https://www.iana.org/assignments/iana-ipv6-special-registry/iana-ipv6-special-registry.xhtml
* added comprehensive routable ip tests
* bumped version with bump script

### DIFF
--- a/packages/data-mate/package.json
+++ b/packages/data-mate/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@terascope/data-mate",
     "displayName": "Data-Mate",
-    "version": "0.25.3",
+    "version": "0.25.4",
     "description": "Library of data validations/transformations",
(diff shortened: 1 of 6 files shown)


peterdemartini:
I think we need to fix the IP validation in https://github.com/terascope/teraslice/blob/8b06aeb269a6f4f3fe90c875f67a6e0935f94e5a/packages/data-mate/src/core/IPValue.ts#L14 here too. Ideally it would use the same code but I didn't want to create a circular dependency

jsnoble (approved):
LGTM

PR #2591, opened 2021-03-26
Expand and improve Data Mate function registry

jsnoble:
- [x] Finish date functions
- [x] Finish geo functions
- [x] Add `toNumber`
- [x] Add `startsWith`
- [x] Add `endsWith`
- [x] Fix issues
- [x] Add support for dynamic arguments
- [x] Coerce arguments, so an arg with field type of `Integer` of `"1.23"` should be converted to `1`
- [x] Remove field type config arguments from `parseJSON` and other fields
- [x] Make `value` required in `equals`
- [x] Update from master
- [x] Add support for DateTuple in the other date functions
- [x] Add support for `entropy` function config 
- [ ] Change validate configs to support changing the output type, since calling `isIP` should change this column type @jsnoble 
- [x] Update function documentation
- [x] Review functions to catch any inconsistencies @jsnoble @ciorg @peterdemartini


packages/data-mate/src/adapters/argument-validator/index.ts
```diff
@@ -0,0 +1,112 @@
...
+        case FieldType.Boolean:
+            return isBoolean;
+        case FieldType.Float:
+        case FieldType.Number:
+        case FieldType.Double:
+        case FieldType.Byte:
+        case FieldType.Short:
+        case FieldType.Integer:
+        case FieldType.Long:
+            return isNumber;
```
peterdemartini, line 50: - `Long` should be validated with `isBigInt`
- `Float` and `Number ` should be validated with `isFloat`
- `Byte`, `Short` and `Integer` should be validated with `isInteger` - also we may want to support the different int size validations, see - https://github.com/terascope/teraslice/blob/master/packages/data-mate/src/builder/types/IntBuilder.ts

packages/data-mate/src/adapters/argument-validator/index.ts
```diff
@@ -0,0 +1,112 @@
...
+        case FieldType.Tuple:
+            return isNumberTuple;
+        default:
+            // equivalent to an any-builder
+            return () => true;
+    }
+}
+
+function isNumberTuple(input: unknown): boolean {
+    return Array.isArray(input) && input.length === 2;
```
peterdemartini, line 68: A tuple is not required to be a specific length, it just has to be a fixed length of at least one.

packages/data-mate/src/adapters/date-frame-adapter/index.ts
```diff
@@ -0,0 +1,210 @@
...
+    [FieldType.KeywordPathAnalyzer]: VectorType.String,
+    [FieldType.Domain]: VectorType.String,
+    [FieldType.Hostname]: VectorType.String,
+    [FieldType.NgramTokens]: VectorType.String,
+    [FieldType.IP]: VectorType.IP,
+    [FieldType.IPRange]: VectorType.IPRange,
+    [FieldType.Date]: VectorType.Date,
+    [FieldType.Boolean]: VectorType.Boolean,
+    [FieldType.Float]: VectorType.Float,
+    [FieldType.Number]: VectorType.Int,
```
peterdemartini, line 36:
```suggestion
    [FieldType.Number]: VectorType.Float,
```

packages/data-mate/src/adapters/date-frame-adapter/index.ts
```diff
@@ -0,0 +1,210 @@
...
+    [FieldType.Byte]: VectorType.Int,
+    [FieldType.Short]: VectorType.Int,
+    [FieldType.Integer]: VectorType.Int,
+    [FieldType.Long]: VectorType.Boolean,
```
peterdemartini, line 40:
```suggestion
    [FieldType.Long]: VectorType.BigInt,
```

packages/data-mate/src/adapters/date-frame-adapter/index.ts
```diff
@@ -0,0 +1,210 @@
...
+    [FieldType.Double]: VectorType.Boolean,
```
peterdemartini, line 41: We can't technically support double until BigFloat but float is fine for now

```suggestion
    [FieldType.Double]: VectorType.Float,
```

packages/data-mate/src/adapters/date-frame-adapter/index.ts
```diff
@@ -0,0 +1,210 @@
...
+    [FieldType.GeoJSON]: VectorType.GeoJSON,
+    [FieldType.GeoPoint]: VectorType.GeoPoint,
+    [FieldType.Boundary]: VectorType.GeoPoint,
+    [FieldType.Geo]: VectorType.GeoPoint,
+    [FieldType.Object]: VectorType.Object,
+    [FieldType.Any]: VectorType.Any,
+    /** FieldType tuple is not the same a VectorType Tuple */
```
peterdemartini, line 48: Why do say this?

packages/data-mate/src/field-transforms/toGeoPoint.ts
```diff
@@ -0,0 +1,87 @@
...
+                ...field_config,
+                type: FieldType.GeoPoint,
+                array
+            },
+        };
+    }
+};
+
+function arrayType(fieldConfig: DataTypeFieldConfig): boolean {
+    if (fieldConfig.type === FieldType.Number || fieldConfig.type === FieldType.Float) {
```
peterdemartini, line 74: Why is this limited to just `Number` and `Float` there are lot more number types?

packages/data-mate/test/field-validators/isEmail-spec.ts
```diff
@@ -0,0 +1,51 @@
...
+            ['Abc@def@example.com', true],
+            ['cal+henderson@iamcalx.com', true],
+            ['customer/department=shipping@example.com', true],
+            ['user@blah.com/junk.junk?a=<tag value="junk"', false],
+            ['Abc\@def  @  example.com', false],
+            ['bad email address', false],
+            [undefined, false],
+            [12345, false],
+            [true, false]
+        ].forEach(([input, expected]) => {
```
peterdemartini, line 35: You can `test.each` here instead of doing inner values, this will ensure one test failure doesn't mask other test failures

packages/data-mate/test/field-validators/isString-spec.ts
```diff
@@ -0,0 +1,46 @@
...
+
+        [
+            ['this is a string', true],
+            ['12345', true],
+            [['hello world'], false],
+            [{ hello: 'there' }, false],
+            [12345, false],
+            [undefined, false],
+            [null, false]
+        ].forEach(([input, expected]) => {
```
peterdemartini, line 31: I recommend using `it.each` so the that one test failure doesn't block other tests from being shown

packages/data-mate/src/adapters/date-frame-adapter/index.ts
```diff
@@ -0,0 +1,276 @@
...
+    [FieldType.IPRange]: VectorType.IPRange,
+    [FieldType.Date]: VectorType.Date,
+    [FieldType.Boolean]: VectorType.Boolean,
+    [FieldType.Float]: VectorType.Float,
+    [FieldType.Number]: VectorType.Int,
+    [FieldType.Byte]: VectorType.Int,
+    [FieldType.Short]: VectorType.Int,
+    [FieldType.Integer]: VectorType.Int,
+    [FieldType.Long]: VectorType.BigInt,
+    [FieldType.Double]: VectorType.BigInt,
```
peterdemartini, line 42: A double cannot be a bigint

peterdemartini:
Let's move to this folder structure:

```
function-configs/
    index.ts
    registry.ts
    string/
       index.ts
       isString.ts
       toString.ts
       parseJSON.ts
    ip/
       index.ts
       isIP.ts
       reserveIP.ts
       inIPRange.ts
    // ...
```

Also I plan on making a few changes to the function definitions in order to create documentation, see below for an example:

```
export const toLowerCaseConfig = {
    name: 'toLowerCase',
    type: 'FIELD_TRANSFORM',
    process_mode: 'INDIVIDUAL_VALUES',
    description: 'Converts a string to lower case characters',
    category: 'STRING',
    examples: [
        { input: 'foo', output: 'foo', description: 'The should not be changed' },
        { input: 'FOO', output: 'foo' }
    ],
    argument_schema: {
        locale: {
            type: 'String',
            description: 'Override the locale to the correct character conversion'
        }
    },
    accepts: ['String'],
    create() {
        return (str: unknown) => String(str).toLowerCase();
    },
};
```

jsnoble:
Wanted to leave a comment to discuss how `@cast` would behave in both the data-frame and function context. The question I have is about how this should be applied. In a sense, in data-frames, this is a no-op action, just changing some metadata which will effect when it will come out. But how would this work in the function context, would I try to create a dictionary that tries to convert the value to the type?

Should `@cast` just handle metadata changes? We partly created this in response to the changes in `@parseJSON`.  That function now returns the parsed values with the `fieldType` set to `Any`. The idea was to use `@cast` to adequately change the fieldConfig (and assign a childConfig if needed) to what it needs to be right after parsing it. This would be more in line of a metadata or data-type transform directory.

This is where the cross between the two is a little difficult. In Data-Frames, assigning a type can change the output, but it would not do so in function-context.

I believe it better to use `@cast` for mutation of the metadata, and leave function context to use toIP, or toNumber to make the hard conversions.

Though I need to bring up the thought that we may need to rethink the use of `@cast` if customers start to use this over the native `toIP`, `toNumber` directives of changing the field-types, that may work in qpl-context, but using the same query in a function context could produce different results, which breaks compatibility between adapters

ciorg:
> I believe it better to use @cast for mutation of the metadata, and leave function context to use toIP, or toNumber to make the hard conversions.

This seems right to me.

peterdemartini:
Ideally they would both function the same but for now I just changing the metadata in the function context is perfectly fine.

ciorg:
Is there any reason that `getValidDate` should not consider the offset when converting a DateTuple to a date object? 

 I feel like it should return the date object and apply the offset instead of just retuning     ` new Date(val[0]);`

peterdemartini:
That may be tricky an answer, since doing so would lose the input timezone for subsequent operations

peterdemartini (approved):
LGTM

peterdemartini:
Set to auto-merge, ship it!

Ship IT!!

PR #2610, opened 2021-03-30
Fix DataFrame.deepSelect and add DataFrame.removeEmptyRows

peterdemartini:
- Fix `DataFrame.deepSelect` to not all of object values in some cases
- Add `DataFrame.removeEmptyRows` which is a performance optimized was of removing empty objects for a given output
- Add `Vector.has` to see a non-nil value exists in a `Vector`


jsnoble (approved):
LGTM

Issue #2608, opened 2021-03-30
add external_ports keyword to teraslice job


godber:
We need a way to expose container ports in k8s on teraslice workers and execution controllers.

godber:
This is completed in the two MRs linked above.

PR #2611, opened 2021-03-31
add basic external_ports functionality

godber:
This PR adds the ability for job authors to expose external ports on their teraslice workers and execution controllers.

refs #2608


godber:
@peterdemartini ... I haven't run this in k8s to verify that the actual generated yaml is correct but this is the gist of what I need to do to expose ports.  The next question is, what would I need to do to make it possible for processors to extend this automatically?  Would all of that just be code in each processor?  Or is there an `exposePort()` method I should implement somewhere to make adding ports easier for asset authors?  We can discuss in person.

**Technically**, we could just call this good though, it would just put the burden of getting the port set on the job author.

The deployment looks like this (assuming I've got the syntax for `ports` right:

```yaml
   apiVersion: apps/v1
    kind: Deployment
    metadata:
      labels:
        app.kubernetes.io/name: teraslice
        app.kubernetes.io/component: worker
        teraslice.terascope.io/exId: e76a0278-d9bc-4d78-bf14-431bcd97528c
        teraslice.terascope.io/jobId: 7ba9afb0-417a-4936-adc5-b15e31d1edd1
        teraslice.terascope.io/jobName: example-data-generator-job
        app.kubernetes.io/instance: ts-dev1
      name: ts-wkr-example-data-generator-job-7ba9afb0-417a
      namespace: ts-dev1
      ownerReferences:
        - apiVersion: batch/v1
          controller: false
          blockOwnerDeletion: false
          kind: Job
          name: undefined
          uid: undefined
    spec:
      replicas: 2
      template:
        metadata:
          labels:
            app.kubernetes.io/name: teraslice
            app.kubernetes.io/component: worker
            teraslice.terascope.io/exId: e76a0278-d9bc-4d78-bf14-431bcd97528c
            teraslice.terascope.io/jobId: 7ba9afb0-417a-4936-adc5-b15e31d1edd1
            teraslice.terascope.io/jobName: example-data-generator-job
            app.kubernetes.io/instance: ts-dev1
        spec:
          containers:
            - image: teraslice:k8sdev
              name: ts-wkr-example-data-generator-job-7ba9afb0-417a
              ports:
                - containerPort: 45680
                - containerPort: 9090
              volumeMounts:
                - mountPath: /app/config
                  name: config
              env:
                - name: NODE_TYPE
                  value: worker
                - name: EX
                  value: >-
                    eyJuYW1lIjoiZXhhbXBsZS1kYXRhLWdlbmVyYXRvci1qb2IiLCJ3b3JrZXJzIjoyLCJqb2JfaWQiOiI3YmE5YWZiMC00MTdhLTQ5MzYtYWRjNS1iMTVlMzFkMWVkZDEiLCJleF9pZCI6ImU3NmEwMjc4LWQ5YmMtNGQ3OC1iZjE0LTQzMWJjZDk3NTI4YyIsImVudl92YXJzIjp7IkZPTyI6ImJheiJ9LCJleHRlcm5hbF9wb3J0cyI6WzkwOTBdfQ==
                - name: POD_IP
                  valueFrom:
                    fieldRef:
                      fieldPath: status.podIP
                - name: FOO
                  value: baz
                - name: EXAMPLE
                  value: test
          volumes:
            - name: config
              configMap:
                name: ts-dev1-worker
                items:
                  - key: teraslice.yaml
                    path: teraslice.yaml
          terminationGracePeriodSeconds: 60
      selector:
        matchLabels:
          app.kubernetes.io/name: teraslice
          app.kubernetes.io/component: worker
          teraslice.terascope.io/exId: e76a0278-d9bc-4d78-bf14-431bcd97528c
```

peterdemartini:
I like the idea of adding an `exposePort` function which should probably be defined on the `ConvictSchema` class, like `ensureAPIFromConfig` https://github.com/terascope/teraslice/blob/c68450c15e5138fcc26d0829e682f9bf00529818/packages/job-components/src/operations/convict-schema.ts#L69

packages/job-components/src/job-schemas.ts
```diff
@@ -238,6 +238,26 @@ export function jobSchema(context: Context): convict.Schema<any> {
             format: Boolean
         };
 
+        schemas.external_port = {
+            doc: '',
+            default: [],
```
godber, line 243: This should be `null` instead of `[]`.

packages/teraslice/lib/config/schemas/system.js
```diff
@@ -236,6 +236,23 @@ const schema = {
         default: false,
         format: Boolean
     },
+    external_ports: {
+        doc: '',
+        default: [],
```
godber, line 241: This should be `null` instead of `[]`.

godber:
I think I want to verify that this runs in k8s today, if it works as is, lets make a release with this change alone.  I can do the nice to have work later.  Just having this `external_ports` property is sufficient for internal use cases which are being blocked by this.

packages/job-components/src/job-schemas.ts
peterdemartini: Does this mean a job can only export a single port?

peterdemartini: Or maybe it's a typo and should be `external_ports` not `external_port`

godber: That is a typo, thanks.

packages/job-components/src/job-schemas.ts
```diff
@@ -238,21 +243,21 @@ export function jobSchema(context: Context): convict.Schema<any> {
...
-                // processor requirs port X this code should throw an error.
-                // We should check to see that the user doesn't specify 45680,
-                // since that is reserved by Teraslice.
+                // processor requires port X this code should throw an error.
                 if (arr != null) {
                     if (!Array.isArray(arr)) {
                         throw new Error('external_ports is required to be an array');
-                    // FIXME: improve input and error handling
+                    }
+                    if (arr.includes(45680)) {
```
peterdemartini, line 259: Why is this port request by teraslice?

packages/job-components/src/job-schemas.ts
```diff
@@ -238,6 +243,26 @@ export function jobSchema(context: Context): convict.Schema<any> {
...
+                // omitted, an empty array, or an array with numbers.  It can't
+                // contain anything other than numbers.  Processors should be able
+                // to have reserved ports.  That is, if a job has port X but a
+                // processor requires port X this code should throw an error.
+                if (arr != null) {
+                    if (!Array.isArray(arr)) {
+                        throw new Error('external_ports is required to be an array');
+                    }
+                    if (arr.includes(45680)) {
+                        throw new Error('Port 45680 cannot be included as an external_port, it is reserved by Teraslice.');
```
peterdemartini, line 260:
```suggestion
                        throw new Error('Port 45680 cannot be included in external_ports, it is reserved by Teraslice.');
```

peterdemartini (commented):
I think this looks good, just a couple minor comments

packages/job-components/src/job-schemas.ts
godber: Technically it's just the Execution controller that uses this port.

https://github.com/terascope/teraslice/blob/0ed9d6c1cfe98c1afa466df005c3ca6b6cff1e79/packages/teraslice/lib/cluster/services/cluster/backends/kubernetes/index.js#L95

For whatever reason, when I implemented this, I chose to not use the default port.  It's an internal detail not exposed to the user in any way until now.  There is probably an argument for not including this in the worker deployments.

peterdemartini: Gotcha, that is fine

peterdemartini (approved):
LGTM

PR #2613, opened 2021-04-02
Fix compatibility of setImmediate

peterdemartini:
`setImmediate` is no longer recommend for use, this fixes a couple cases where we use it. I also removed the `BigInt` warning since it was more annoying than helpful. I figured it would easier to tell it fail when using a `BigInt` instead of prematurally warning about it.


jsnoble (approved):
LGTM

PR #2628, opened 2021-04-05
Add DataFrame format serialize/deserialization

peterdemartini:
- Add `DataFrame#empty` static method (an easier API for creating empty data frames, not common but works better)
- Add `DataFrame.serialize` and `Column.serialize` method
- Add `DataFrame#deserialize` and `Column#deserialize` static method
- **BREAKING** unix timestamps now allow negative numbers
- Fix bigint parsing when number is greater than `Number.MAX_SAFE_INTEGER`
- Fix timestamp parsing in some cases when a timezone isn't specified
- Fix matching bigint in document matcher
- Remove `DateValue` and `IPValue` 
- Improve GeoJSON and GeoPoint Builder performance


peterdemartini:
Basically the serialized format ended up being like ldjson:

```json
// data frame metadata
{"size":1,"metadata":{},"config":{"version":1,"fields":{"_key":{"type":"Keyword"},"name":{"type":"Keyword"},"age":{"type":"Short"}}}}
// columns
{"name":"_key","size":1,"version":1,"config":{"type":"Keyword"},"childConfig":{},"values":["7f9b0666-c63e-4501-aa24-8b673e7af38c"]}
{"name":"name","size":1,"version":1,"config":{"type":"Keyword"},"childConfig":{},"values":["Don Montgomery"]}
{"name":"age","size":1,"version":1,"config":{"type":"Short"},"childConfig":{},"values":[56]}
```

PR #2629, opened 2021-04-07
Eslint improvements

peterdemartini:
- Adds support for ecmascript v9 (some features we use that are supported in node 10 weren't parsable by eslint)
- Remove the eslint rule that was recently re-enabled that required more types defined for private functions (that should be up to the developer but recommended).
- Disable `no-useless-return` since that would autofix functions while developing and require adding the returns back in


jsnoble (approved):
LGTM

PR #2631, opened 2021-04-08
Add a more simple DataFrame serialize/deserialize API

peterdemartini:
Make `serializeIterator` and `deserializeIterator` a more advanced use case, this also makes `serialize` and `deserialize` a more simple API since the new line joining and splitting is required knowledge.

This also fixes a bug where empty lines caused by ending new lines would cause serialization errors.


jsnoble (approved):
LGTM

PR #2646, opened 2021-04-14
Fix groupBy key generation

peterdemartini:
- Fix groupBy key generation when different fields have the same value

Previously, this would have generated the same group by key for both records since 
the field name or position was not taken into account. This is incorrect, it should have generated two 
different buckets.

```js
[{
  "field1": "hello",
  "field2": null
}, {
  "field1": null,
  "field2": "hello"
}]
```


jsnoble (approved):
LGTM

PR #2651, opened 2021-04-16
add external port object type

godber:
This changes `external_ports` from a simple array of numbers to an array that can contain numbers or objects, looks like:

```json
    "external_ports": [
        9090,
        {"name": "metrics", "port": 3333}
    ]
```

### DIFF
--- a/packages/job-components/src/job-schemas.ts
+++ b/packages/job-components/src/job-schemas.ts
@@ -272,7 +272,6 @@ export function jobSchema(context: Context): convict.Schema<any> {
                                 if (val == null || val === '') {
                                     throw new Error(`value for key "${key}" must be not empty`);
                                 }
-                                // eslint-disable-next-line no-prototype-builtins
                                 if (hasOwn(portValue, 'name') && hasOwn(portValue, 'port')) {
                                     if (!isNumber(portValue.port)) {
                                         throw new Error('The port set on an external_ports object must be a number.');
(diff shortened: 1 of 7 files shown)


packages/job-components/src/job-schemas.ts
```diff
@@ -254,17 +255,44 @@ export function jobSchema(context: Context): convict.Schema<any> {
...
+                            Object.entries(portValue).forEach(([key, val]) => {
+                                if (key == null || key === '') {
+                                    throw new Error('key must be not empty');
+                                }
+
+                                if (val == null || val === '') {
+                                    throw new Error(`value for key "${key}" must be not empty`);
+                                }
+                                // eslint-disable-next-line no-prototype-builtins
+                                if (portValue.hasOwnProperty('name') && portValue.hasOwnProperty('port')) {
```
peterdemartini, line 275: That eslint rule shouldn't be ignored, I would recommend using `hasOwn` from `@terascope/utils`.

```suggestion
                                if (hasOwn(portValue, 'name') && hasOwn(portValue, 'port')) {
```

peterdemartini (changes requested):
One small change, then we can bump this

packages/job-components/src/job-schemas.ts
godber: I've addressed this and bumped the patch version.

peterdemartini (approved):
LGTM

godber:
This PR is also relevant to https://github.com/terascope/teraslice/issues/2608

PR #2665, opened 2021-04-26
String functions

ciorg:
* added string functions to data-mate and utils 
* respective tests


packages/data-mate/src/field-validators/equals.ts
```diff
@@ -0,0 +1,50 @@
...
+    value: any;
+}
+
+export const equalsConfig: FieldValidateConfig<IncludesArgs> = {
+    name: 'equals',
+    type: FunctionDefinitionType.FIELD_VALIDATION,
+    process_mode: ProcessMode.INDIVIDUAL_VALUES,
+    description: 'Checks to see if input matches the value',
+    create({ value }) {
+        return (input: unknown) => isSame(input, value);
```
peterdemartini, line 40: Using `isSame` here might not be the correct choice, there are definity cases where the array element positions do matter. For example, `[23.99234, 39.239999]` should not be equal to `[39.239999, 23.99234]`.

packages/data-mate/src/field-validators/equals.ts
```diff
@@ -0,0 +1,50 @@
...
+    },
+    accepts: [],
+    argument_schema: {
+        value: {
+            type: FieldType.Any,
+            array: false,
+            description: 'any value'
```
peterdemartini, line 47: This description doesn't add any value. Something like `Value to use in the comparison`

jsnoble:
Thanks for the catch, I will put those changes into my commit

PR #2666, opened 2021-04-27
Add esbuild style asset creation

godber:
This is a WIP to enable bundled asset builds with teraslice cli (earl).

Here are the related asset PRs

* https://github.com/terascope/elasticsearch-assets/pull/704
* https://github.com/terascope/file-assets/pull/596
* https://github.com/terascope/kafka-assets/pull/470
* https://github.com/terascope/standard-assets/pull/460


godber:
I still need to do the following:

- [ ] add wasm support
- [ ] figure out how to do the "require test" with `execa`
- [ ] add support to other assets, I've just been working with `elasticsearch-asset`

peterdemartini:
For the require test, I think it is fine if you do, `console.dir(require(targetFilePath))`.

godber:
Note that this only generates an asset for the single specified `buildTarget` version.  Not for multiple build versions.

The node versioning situation is getting a bit messy, in the latest push I have change the filename syntax a little bit:

* bundled - `elasticsearch-v2.5.2-node12-bundle.zip`
* no bundle - `elasticsearch-v2.5.2-node-12-darwin-x64.zip`

I should probably fix that ... actually, I wonder if we have anything that expects the `node-VERSION-OS-ARCH.zip` format ... I bet we do.  Obviously I can change the `node12` to be `node-12`.

Note that in the bundled style, the `node_version` in the package comes from the `bundleTarget` value rather than the `process.version` value.  I am pretty sure this is what we want.

I think I will go back and fix the `node-12` thing.

Note the `bundleTarget` options are as follows

```
            choices: [
                'node10',
                'node12',
                'node14',
                'node16'
            ],
```

Constraining it helps me parse out the version, hopefully we don't need free formed options like `node12.21` etc.

I still have the following concerns

* `entrypoints` is a problem, I was going to assess the situation when looking at other assets.
* no WASM support
* I still don't test the generated index.js.
  * I don't think `console.dir(require(targetFilePath))` will work for cases where the `target` doesn't match the node version creating the asset.
  * calling out with `execa` is messy also since that requires us to manage the `node` version in the CI system right?
  * Ultimately, I feel like this needs to be solved at a higher level.  Managed by CI ... validating the zip assets.

packages/teraslice-cli/src/helpers/asset-src.ts
```diff
@@ -42,14 +48,22 @@ export class AssetSrc {
...
     }
 
     get zipFileName(): string {
-        const nodeVersion = process.version.split('.')[0].substr(1);
-        return `${this.name}-v${this.version}-node-${nodeVersion}-${process.platform}-${process.arch}.zip`;
+        let zipName;
+        let nodeVersion;
+        if (this.bundle) {
+            nodeVersion = this.bundleTarget?.replace('node', '');
+            zipName = `${this.name}-v${this.version}-node-${nodeVersion}-bundle.zip`;
```
godber, line 61: I am debating whether this needs to be:

```
zipName = `${this.name}-v${this.version}-node-${nodeVersion}-bundle-none.zip`;
```

for tooling compatibility reasons.

godber:
Here's my really lame asset testing script.

```
#!/usr/bin/env bash

ASSET_FILE=$1
DIR=$(mktemp -d)

unzip $ASSET_FILE -d ${DIR}
node -p "require('${DIR}/index.js').ASSETS"

echo "rm -rf ${DIR}"

```

godber:
Ok, I got this to fail like I wanted it.  In the case where the node version running cli matches the budle target it will now automatically require the bundled asset as a test.

godber:
The changes to `loadAPI` were to address namespaced operations.  When submitting this job:

```
{
    "name": "test#es-read",
    "lifecycle": "once",
    "assets": [
        "elasticsearch"
    ],
    "workers": 1,
    "operations": [
        {
            "index": "log*",
            "date_field_name": "@timestamp",
            "connection": "default"
        },
        {
            "_op": "noop"
        }
    ]
}
```

I got this error:

```
Failure loading schema from module: elasticsearch_reader_api, error: Error: Could not find operation elasticsearch_reader_api:elasticsearch_reader-0
    at OperationLoader._getBundledOperation (/Users/godber/Workspace/terascope/opensource/teraslice/packages/job-components/dist/src/operation-loader/loader.js:51:19)
    at OperationLoader.require (/Users/godber/Workspace/terascope/opensource/teraslice/packages/job-components/dist/src/operation-loader/loader.js:293:29)
    at OperationLoader.loadAPI (/Users/godber/Workspace/terascope/opensource/teraslice/packages/job-components/dist/src/operation-loader/loader.js:228:27)
    at /Users/godber/Workspace/terascope/opensource/teraslice/packages/job-components/dist/src/job-validator.js:47:46
    at Array.map (<anonymous>)
    at JobValidator.validateConfig (/Users/godber/Workspace/terascope/opensource/teraslice/packages/job-components/dist/src/job-validator.js:46:41)
    at _validateJobSpec (/Users/godber/Workspace/terascope/opensource/teraslice/packages/teraslice/lib/cluster/services/jobs.js:44:45)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Object.submitJob (/Users/godber/Workspace/terascope/opensource/teraslice/packages/teraslice/lib/cluster/services/jobs.js:55:26)
    at async /Users/godber/Workspace/terascope/opensource/teraslice/packages/teraslice/lib/utils/api_utils.js:54:28
```

godber:
OK, I think this is ready for review, I haven't bumped any versions yet.  What did I touch here ... 

* `teraslice-cli` - minor level
* `job-components`  patch level

Does that seem right?

peterdemartini:
That's right

packages/teraslice-cli/src/helpers/asset-src.ts
peterdemartini: I don't think it would be break the tooling and it makes more confusing

peterdemartini (changes requested):
I just remembered we need to bump teraslice in order for the operation loader fixes to occur.

peterdemartini (approved):

PR #2668, opened 2021-04-27
Fix spelling and interfaces

peterdemartini:
- Fix the spelling of data frame adapter (it is not date adapter)
- Fix the interfaces to work a little better in typescript (make some properties readonly)
- Add a basic function test harness (not used yet)
- Update the function definition config to also include the examples (so we can add them while we develop)


jsnoble (approved):
LGTM

PR #2671, opened 2021-04-28
Add function test harness

peterdemartini:
- Add function test harness
- Adds category to function definition config
- Improves some typescript definitions
- Ensures the function definition config works correctly with the data type config


jsnoble (approved):
LGTM

PR #2675, opened 2021-04-28
Transform functions

jsnoble:
- adding more string functions: extract, replaceLiteral, replaceRegex, splitString, reverse
- adding examples to string functions
- removing test files if they are uneccessary though the new examples testing feature


packages/data-mate/src/function-configs/string/decodeBase64.ts
```diff
@@ -13,6 +13,23 @@ export const decodeBase64Config: FieldTransformConfig = {
...
     category: FunctionDefinitionCategory.STRING,
     description: 'Converts a base64 hash back to its value',
+    examples: [
+        {
+            args: {},
+            config: {
+                version: 1,
+                fields: {
+                    testField: {
+                        type: FieldType.String
```
peterdemartini, line 23: For the field transform/validator examples, I was thinking naming the focus field should be named the same thing. I think as long as it is consistent (so the documentation read more correctly) we are fine, the `contains` example use the field `"example"` so we want to go with `"testField"` let's change the `contains` example too.

packages/data-mate/src/adapters/function-adapter/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+        }
+    } else if (isValidator) {
+        try {
+            if (!isNil(input) && fn(input)) {
+                results.push(input);
+            } else if (preserveNulls) {
+                results.push(null);
+            }
+        } catch (_err) {
+            if (preserveNulls) results.push(null);
```
peterdemartini, line 30: We don't want to hide the error here, it should be up the to individual functions to not throw. Unexpected errors will happen and we need to expose them so they humans can determine how to fix it.

packages/data-mate/src/adapters/function-adapter/utils.ts
```diff
@@ -0,0 +1,43 @@
+import {
+    isNil,
+    flatten
+} from '@terascope/utils';
+
+export function callValue(
```
peterdemartini, line 6: This function name is ambiguous, which is not deal since it is private but it should at least have jsdoc describing why it was created and how it is used.

packages/data-mate/src/function-configs/string/toString.ts
```diff
@@ -5,15 +5,48 @@ import {
     ProcessMode,
     FunctionDefinitionType,
     DataTypeFieldAndChildren,
-    FunctionDefinitionCategory
+    FunctionDefinitionCategory,
+    FunctionDefinitionExample
 } from '../interfaces';
 
+const examples: FunctionDefinitionExample<Record<string, unknown>>[] = [
+    // {
```
peterdemartini, line 13: Why is there a bunch of examples commented out here?

packages/data-mate/src/adapters/function-adapter/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+    if (Array.isArray(input)) {
+        const mappedInput = flatten(
+            input.map((newValue) => callValue(fn, newValue, preserveNulls, isValidator))
+        );
+
+        if (mappedInput.length) {
+            results.push(...mappedInput);
+        }
+    } else if (isValidator) {
+        if (!isNil(input) && fn(input)) {
```
peterdemartini, line 27: I wondering why there was this check here and I remembered that validating a nil value is useless. We should have a comment describing that since it took me a few minutes to remember that

peterdemartini (changes requested):

peterdemartini (approved):

PR #2676, opened 2021-04-28
data-mate/ utils string functions

ciorg:
* isPort
* isMimeType
* isISDN
* isPhoneNumberLIke
* isPostolCode
* isAlphaNumeric


packages/data-mate/src/function-configs/string/isMimeType.ts
```diff
@@ -0,0 +1,11 @@
+import { isMimeType } from '@terascope/utils';
+import { FieldValidateConfig, ProcessMode, FunctionDefinitionType } from '../interfaces';
+
+export const isMimeTypeConfig: FieldValidateConfig = {
+    name: 'isMimeType',
```
peterdemartini, line 5: Shouldn't this be `isMIMEType`?

packages/data-mate/src/function-configs/string/isMimeType.ts
```diff
@@ -0,0 +1,11 @@
...
+    type: FunctionDefinitionType.FIELD_VALIDATION,
+    process_mode: ProcessMode.INDIVIDUAL_VALUES,
+    description: 'Checks to see if input is a MIME Type',
```
peterdemartini, line 8: This description doesn't give you any more value than the function itself. Maybe describe what MIME is?

packages/data-mate/src/function-configs/string/isMimeType.ts
```diff
@@ -0,0 +1,11 @@
...
+    create() { return isMimeType; },
+    accepts: [],
```
peterdemartini, line 10: This should probably only accept strings since this could be optimized out if given anything else
```suggestion
    accepts: [FieldType.String],
```

packages/utils/src/phone-number.ts
```diff
@@ -13,3 +15,21 @@ export function parsePhoneNumber(input: string|number): string {
...
+    if (isString(input) || isNumber(input)) {
+        const isdn = new PhoneValidator(`+${input}`);
+
+        return isdn.isValid();
+    }
+
+    return false;
+}
+
+export function isPhoneNumberLike(input: unknown): boolean {
```
peterdemartini, line 31: What is the benefit of this function? It seems extremely loose?

ciorg: It is extremely loose, the problem is the isISDN assumes everything is in an international phone number format and throws out a lot of possibly good phone numbers. 

If a country is provided along with a number to the ISDN it becomes much more reliable, and thinking about this the country needs to be an optional arg for the isISDN function.  

But in enough cases the data didn't have accurate country info and isISDN was rejecting most of the phone number values, so the isPhoneNumber like function was created.  A basic check based on number of  digits, min and max for a valid phone number.

packages/data-mate/src/function-configs/string/isMimeType.ts
ciorg: fixed this with latest commit

packages/data-mate/src/function-configs/string/isMimeType.ts
ciorg: fixed this with latest commit

packages/data-mate/src/function-configs/string/isMimeType.ts
ciorg: I added more details to the description

ciorg:
Had a strange issue where I changed the file isMimeType -> isMIMEType and was causing an error on the build saying it couldn't find the isMIMEType file for the CI tests.  It built fine locally, but I changed the file name back to isMimeType and it passed.  Not sure  whats going on there.

peterdemartini:
That's because macos file system is case insensitive by default. You need to use `git mv` to tell git the file was moved. You might need to rename it back and then use the `git mv` command.

packages/data-mate/src/function-configs/string/isAlphaNumeric.ts
```diff
@@ -0,0 +1,33 @@
...
+            type: FieldType.String,
+            description: 'Specify locale to check for valid alpha-numeric characters, defaults to en-US if not provided'
+        }
+    },
+    accepts: [FieldType.String],
+    required_arguments: [],
+    validate_arguments({ locale }) {
+        if (locale == null) return;
+
+        if (!isString(locale) || !('this' in validator.isAlphanumericLocales)) {
```
peterdemartini, line 29:
```suggestion
        if (!isString(locale) || !(locale in validator.isAlphanumericLocales)) {
```

peterdemartini:
This needs to be updated from master, also we should be writing examples (a lot of the spec files you created can be deleted)

packages/utils/src/phone-number.ts
peterdemartini: We should probably have that described in jsdoc

peterdemartini (approved):

packages/data-mate/src/function-configs/string/isBase64.ts
```diff
@@ -58,5 +58,7 @@ export const isBase64Config: FieldValidateConfig = {
     examples,
     description: 'Checks to see if input is a valid base64 string',
     create() { return isBase64; },
-    accepts: [FieldType.String],
+    accepts: [
+        FieldType.String
```
jsnoble, line 62: does this need to be on a new line?

packages/data-mate/src/function-configs/string/isCountryCode.ts
```diff
@@ -86,5 +86,7 @@ export const isCountryCodeConfig: FieldValidateConfig = {
     examples,
     description: 'Checks to see if input is a valid ISO 3166-1 alpha-2 country code',
     create() { return isCountryCode; },
-    accepts: [FieldType.String],
+    accepts: [
+        FieldType.String
```
jsnoble, line 90: does this need to be on a new line?

packages/data-mate/src/function-configs/string/isEmail.ts
```diff
@@ -144,5 +144,7 @@ export const isEmailConfig: FieldValidateConfig = {
     create() {
         return isEmail;
     },
-    accepts: [FieldType.String],
+    accepts: [
+        FieldType.String
```
jsnoble, line 148: does this need to be on a new line?

packages/data-mate/src/function-configs/string/isFQDN.ts
```diff
@@ -114,5 +114,7 @@ export const isFQDNConfig: FieldValidateConfig = {
     description: 'Checks to see if input is a fully qualified domain name',
     examples,
     create() { return isFQDN; },
-    accepts: [FieldType.String],
+    accepts: [
+        FieldType.String
```
jsnoble, line 118: does this need to be on a new line?

jsnoble (changes requested):
Is there a new reason for the styling changes

jsnoble (approved):
LTGM

Issue #2703, opened 2021-05-10
Add BigLRUMap


peterdemartini:
We ran into "Invalid Array Size OOM errors" when using LRUMap with more than 12 million keys. We need to a `BigLRUMap` support to `@terascope/utils` that will support for more than 12 millions keys.

It essentially needs to look like this:

```ts
/**
* This used to ignore the init size of
* the array in LRUMap. Doing this will avoid
* running out of memory at 12 million records.
**/
class FlexibleArray {
    constructor() {
        return [];
    }
}

// this should come from mnemonist/lru-map
const map = new LRUMap(
    FlexibleArray, 
    // Using a typed array here should still possible
    FlexibleArray,
    cacheSize
);
map.items = new BigMap();
```

This also needs to be used in teraslice-state-storage

PR #2706, opened 2021-05-11
Add big lru map to utils

jeffmontagna:
- Added new class `BigLruMap` that extends `LRUMap` to include `BigMap` and a flexible array to avoid running out of memory with 11.8 million items in the map.

Closes https://github.com/terascope/teraslice/issues/2703

### DIFF
--- a/packages/utils/src/big-lru-map.ts
+++ b/packages/utils/src/big-lru-map.ts
@@ -1,0 +1,23 @@
+/* eslint-disable max-classes-per-file */
+
+import LRUMap from 'mnemonist/lru-map';
+import { BigMap } from '@terascope/utils';
+
+/**
+* This used to ignore the init size of
+* the array in LRUMap. Doing this will avoid
+* running out of memory at 12 million records.
+* */
+class FlexibleArray {
+    constructor() {
+        return [];
+    }
+}
+
+export class BigLruMap<V> extends LRUMap<string, V> {
+    constructor(cacheSize: number) {
+        super(FlexibleArray as any, FlexibleArray as any, cacheSize);
+        // @ts-expect-error
+        this.items = new BigMap();
+    }
+}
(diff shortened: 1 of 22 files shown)


packages/utils/src/big-lru-map.ts
```diff
@@ -0,0 +1,23 @@
...
+* the array in LRUMap. Doing this will avoid
+* running out of memory at 12 million records.
+* */
+class FlexibleArray {
+    constructor() {
+        return [];
+    }
+}
+
+export class BigLruMap<V> extends LRUMap<string|number, V> {
```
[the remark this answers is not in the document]
jeffmontagna: Changed class name to `BigLRUMap`

packages/utils/src/big-lru-map.ts
```diff
@@ -0,0 +1,23 @@
...
+    constructor(cacheSize: number) {
```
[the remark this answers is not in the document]
jeffmontagna: Added support to key and value arrays being typed.

packages/utils/src/big-lru-map.ts
```diff
@@ -0,0 +1,23 @@
...
+* running out of memory at 12 million records.
+* */
+export class FlexibleArray {
+    constructor() {
+        return [];
+    }
+}
+
+export class BigLRUMap<V> extends LRUMap<string|number, V> {
+    constructor(mapSize: number, keyArray?: FlexibleArray, valueArray?: FlexibleArray) {
```
peterdemartini, line 18: We need to make sure that the flexible array is passed in, also this should support a `TypedArray` (this comes from `@terascope/types`)

```suggestion
    constructor(mapSize: number, keyArray: FlexibleArray|TypedArray = FlexibleArray, valueArray: FlexibleArray|TypedArray = FlexibleArray) {
```

jeffmontagna: If keyArray and valueArray are not optional would it make sense to move FlexibleArray to @terascope/types?

peterdemartini: Not really, this `FlexibleArray` is really a hack and isn't a type only thing. Ideally we shouldn't even be exporting it

peterdemartini (approved):
LGTM

PR #2710, opened 2021-05-12
Date func get between times

ciorg:
submitting this to get feedback:
  * timeBetweenFormat type - I have as an enum, not sure if that is the best way to do that
  * also requiring the format in the args in the func config is causing a type error that I have not been able to figure out - ` Property 'format' is missing in type 'Record<string, unknown>' but required in type 'GetTimeBetweenArgs`, seems like it needs to be required, there could be a default like seconds, but would still like to understand this error and how to fix it.


packages/types/src/dates.ts
```diff
@@ -7,3 +7,24 @@ export enum ISO8601DateSegment {
...
+    Weeks = 'weeks',
+    CalendarWeeks = 'CalendarWeeks',
+    Months = 'months',
+    CalendarMonths = 'calendarMonths',
+    Quarters = 'quarters',
+    CalendarQuarters = 'CalendarQuarters',
+    Years = 'years',
+    CalendarYears = 'CalendarYears',
+    CalendarISOWeekYears = 'CalendarISOWeekYears',
+    IsoWeekYears = 'isoWeekYears',
```
peterdemartini, line 28:
```suggestion
    ISOWeekYears = 'ISOWeekYears',
```

packages/types/src/dates.ts
```diff
@@ -7,3 +7,24 @@ export enum ISO8601DateSegment {
...
+    IsoDuration = 'isoDuration'
```
peterdemartini, line 29:
```suggestion
    ISODuration = 'ISODuration'
```

packages/types/src/dates.ts
```diff
@@ -7,3 +7,24 @@ export enum ISO8601DateSegment {
     monthly = 'monthly',
     yearly = 'yearly',
 }
+
+export enum TimeBetweenFormats {
```
peterdemartini, line 11: Is there are reason there casing differences between the keys and values of this enum? That might lead to a missunderstanding

packages/data-mate/src/function-configs/date/getTimeBetween.ts
```diff
@@ -0,0 +1,78 @@
...
+    create(args: GetTimeBetweenArgs) {
+        return (input: unknown) => getTimeBetween(input, args);
+    },
+    accepts: [FieldType.Date, FieldType.String, FieldType.Number],
+    validate_arguments({ start, end, format }: GetTimeBetweenArgs) {
+        if (start == null && end == null) {
+            throw Error('Must provide a start or an end value');
+        }
+
+        if (format == null) {
```
peterdemartini, line 74: This should be done with `required_arguments: ['format']`

packages/data-mate/src/function-configs/date/getTimeBetween.ts
```diff
@@ -0,0 +1,78 @@
...
+    ProcessMode, FunctionDefinitionType, FunctionDefinitionCategory, FieldTransformConfig
+} from '../interfaces';
+
+export interface GetTimeBetweenArgs {
+    start?: Date | number | string;
+    end?: Date | number | string;
+    readonly format: TimeBetweenFormats;
+}
+
+export const getTimeBetweenConfig: FieldTransformConfig = {
```
peterdemartini, line 13: This will fix the type error you are getting:

```suggestion
export const getTimeBetweenConfig: FieldTransformConfig< GetTimeBetweenArgs > = {
```

packages/data-mate/src/function-configs/date/getTimeBetween.ts
```diff
@@ -0,0 +1,78 @@
...
+            type: FieldType.Date || FieldType.String || FieldType.Number,
+            description: 'End time of time range, if provided end is after input the return value will be negative'
+        },
+        format: {
+            type: FieldType.String,
+            description: 'The format of the return value'
+        }
+    },
+    create(args: GetTimeBetweenArgs) {
+        return (input: unknown) => getTimeBetween(input, args);
```
jsnoble, line 66: would it be feasible for a getTimeBetweenFP version? Would it make sense? it could help with stack traces that we are not returning an anonymous function

jsnoble (changes requested):

packages/data-mate/src/function-configs/date/getTimeBetween.ts
ciorg: added getTimeBetweenFP, I'm not 100% clear on the reasoning here.

packages/types/src/dates.ts
ciorg: ended up replacing the enum with a type 'milliseconds' | 'seconds' | 'hours' | etc...

the enum was giving me type issues, the custom type seems to work better in this case

packages/types/src/dates.ts
ciorg: no, should not have been that way

jsnoble (approved):
LGTM for my stuff

peterdemartini (approved):
LGTM

Issue #2704, opened 2021-05-10
Update supported Node versions to node 12/14


peterdemartini:
Packages are coming out dropping node 10 support, which means we need to consider removing that support as well since it out of LTS. As always the biggest challenge is updating node-rdkafka. Once we can verify the new version of node-rdkafka works, we can go ahead and do the following:

- [x] Update CI in kafka-assets to support node 12/14
- [x] Update CI in file-assets to support node 12/14
- [x] Update CI in standard-assets to support node 12/14
- [x] Update CI in elasticsearch-assets to support node 12/14
- [x] Update CI in teraslice to support node 12/14
- [x] Update the teraslice image to node 14
- [x] Update tsconfig and eslint ecmascript target to "es2019"

godber:
I think these all cover the checkboxes above:

* https://github.com/terascope/elasticsearch-assets/pull/712
* https://github.com/terascope/file-assets/pull/604
* https://github.com/terascope/kafka-assets/pull/475
* https://github.com/terascope/standard-assets/pull/465
* https://github.com/terascope/teraslice/pull/2717

The teraslice tests won't pass until we have a node14 build of the elasticsearch asset.

godber:
All of the asset PRs are green now.

peterdemartini:
Let's get on with it then haha

godber:
Assets are merged and released:

* [Elasticsearch](https://github.com/terascope/elasticsearch-assets/releases/tag/v2.6.0)
* [File-asset](https://github.com/terascope/file-assets/releases/tag/v2.1.0)
* [Standard-asset](https://github.com/terascope/standard-assets/releases/tag/v0.11.0)
* [Kafka-asset](https://github.com/terascope/kafka-assets/releases/tag/v3.2.0)

godber:
The teraslice e2e tests still fail for node 12/14

https://travis-ci.com/github/terascope/teraslice/jobs/505614200

https://travis-ci.com/github/terascope/teraslice/jobs/505614204

It says it's an asset version mismatch problem, but I don't know how/why.  I'll come back to this tomorrow.

peterdemartini:
Sorry, I forgot to tell you about this [line](https://github.com/terascope/teraslice/blob/master/e2e/test/download-assets.js#L9), this should have pulled from the `Dockerfile` or something

godber:
These have all been completed and the `v0.77.0` image on docker hub uses node 14.  We will open individual issues to address anything that comes up.

PR #2717, opened 2021-05-13
v0.77.0 - Update to node support to 12/14

godber:
This PR upgrades our CI to use node 14 as the primary node version and treats node 12 as the "old" compatibility release.  Node 10 is no longer tested at this point.

refs: #2704


godber:
Also note my first commit was some cleanup and spelling.  There were duplicate keys that didn't make any sense.

.travis.yml
```diff
@@ -2,7 +2,7 @@
 dist: bionic
 os: linux
 language: node_js
-node_js: '12.20'
+node_js: '14'
```
godber, line 5: Note, I only ever use `node_js: '14'` ... is there a minor version I should attach to these?  It was unclear what minor versions were acceptable.

peterdemartini: Maybe `14.17`, will be our base version, maybe we should the base node 12 version to `12.22`

.travis.yml
```diff
@@ -53,29 +53,52 @@ jobs:
...
       if: branch = master AND type IN (pull_request, cron) AND fork = false
       script:
         - export SERVICE_HEAP_OPTS="-Xms768m -Xmx768m"
         - yarn --silent --cwd e2e test
 
+    - name: 'Unit Test Suite (node 14)'
+      node_js: '14'
+      # run only on pull-requests or cron
+      if: branch = master AND type IN (pull_request, cron)
+      script: yarn --silent test --suite unit-b
```
godber, line 80: I don't understand the rolls of suites `unit-a` and `unit-b`.  Basically what I've done here is moved the duplicated the node 12 tests with node 14 tests.  Should some of the old node 12 tests with ES be eliminated?

peterdemartini (approved):
LGTM

PR #2720, opened 2021-05-14
Add in several date functions, and other string functions

jsnoble:
added several functions to dates, strings and added toNumber

### DIFF
--- a/packages/data-mate/src/function-configs/date/isToday.ts
+++ b/packages/data-mate/src/function-configs/date/isToday.ts
@@ -1,0 +1,38 @@
+import { isToday } from '@terascope/utils';
+import { FieldType } from '@terascope/types';
+import {
+    ProcessMode, FunctionDefinitionType, FunctionDefinitionCategory, FieldValidateConfig
+} from '../interfaces';
+
+const date = new Date().toISOString();
+
+export const isTodayConfig: FieldValidateConfig = {
+    name: 'isToday',
+    type: FunctionDefinitionType.FIELD_VALIDATION,
+    process_mode: ProcessMode.INDIVIDUAL_VALUES,
+    category: FunctionDefinitionCategory.DATE,
+    examples: [
+        {
+            args: {},
+            config: { version: 1, fields: { testField: { type: FieldType.String } } },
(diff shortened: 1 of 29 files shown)


packages/data-mate/src/function-configs/date/isToday.ts
```diff
@@ -0,0 +1,38 @@
...
+        },
+        {
+            args: {},
+            config: { version: 1, fields: { testField: { type: FieldType.String } } },
+            field: 'testField',
+            input: '2020-05-09T10:00:00.000Z',
+            output: null
+        },
+    ],
+    description: 'Determines if the given date is on the same day (server-time)',
```
peterdemartini, line 31: "server time" is not correct, servers can be configured with a time zone (even though we don't), it is better to be clear that it is UTC

peterdemartini (changes requested):
One small change then lets merge this

packages/data-mate/src/function-configs/date/isTomorrow.ts
```diff
@@ -0,0 +1,42 @@
...
+        {
+            args: {},
+            config: { version: 1, fields: { testField: { type: FieldType.String } } },
+            field: 'testField',
+            input: tomorrowDate,
+            output: tomorrowDate,
+            description: 'represents day after current time'
+        },
+    ],
+    description: 'Determines if the given date is on the next day (server-time)',
```
ciorg, line 35: should be UTC

peterdemartini (approved):
LGTM

ciorg (changes requested):

packages/data-mate/src/function-configs/date/isTomorrow.ts
peterdemartini: Haha we gave him the same comment and he just fixed it

ciorg (approved):

packages/data-mate/src/function-configs/date/isTomorrow.ts
ciorg: nice

PR #2721, opened 2021-05-14
Get years months date hours min sec msec functions

ciorg:
functions:
  * getMilliseconds
  * getSeconds
  * getMinutes
  * getHours
  * getDate
  * getMonth
  * getYear

Question on getDate - that could be confused with a complete date, not day of the month.  Maybe it should be getDayOfMonth?

### DIFF
--- a/packages/data-mate/src/function-configs/date/getHours.ts
+++ b/packages/data-mate/src/function-configs/date/getHours.ts
@@ -1,0 +1,67 @@
+import { FieldType } from '@terascope/types';
+import { getHours } from '@terascope/utils';
+import {
+    ProcessMode, FunctionDefinitionType, FunctionDefinitionCategory, FieldTransformConfig
+} from '../interfaces';
+
+export const getHoursConfig: FieldTransformConfig = {
+    name: 'getHours',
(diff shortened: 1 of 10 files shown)


packages/data-mate/src/function-configs/date/getHours.ts
```diff
@@ -0,0 +1,67 @@
...
+            output: 10
+        },
+        {
+            args: {},
+            config: {
+                version: 1,
+                fields: { testField: { type: FieldType.String } }
+            },
+            field: 'testField',
+            input: '05/22/2021 EST',
```
peterdemartini, line 41: How does this work without specifying the format?

ciorg: new Date('string') can still parse that, so the getValidDate function is able to handle that format.

peterdemartini: Gotcha.

peterdemartini (approved):
LGTM

jsnoble (approved):
LGTM

PR #2724, opened 2021-05-17
Add generic encode and misc fixes

peterdemartini:
- [x] Add generic `encode`
- [x] Add `join`
- [x] Add performance improvement to date aggregations
- [x] Rename `isSame` to `isDeepEqual`, put it in `packages/utils/src/equality` not `packages/utils/src/objects.ts`, and use the lodash `isEqual` since I don't think it worked correct with certain values. I made sure this wasn't a breaking change by adding an alias for `isSame`

### DIFF
--- a/packages/data-mate/src/function-configs/string/encode.ts
+++ b/packages/data-mate/src/function-configs/string/encode.ts
@@ -101,12 +101,12 @@ export const encodeConfig: FieldTransformConfig<EncodeConfig> = {
         algo: {
             type: FieldType.String,
             array: false,
-            description: 'Which has hashing algorithm to use'
+            description: 'Which hashing algorithm to use'
         },
         digest: {
             type: FieldType.String,
             array: false,
-            description: 'Which has digest to use, may be set to either "base64" or "hex", defaults to "hex". '
+            description: 'Which hash digest to use, may be set to either "base64" or "hex", defaults to "hex". '
             + 'Only used when algorithm is not base64, hex, or url'
         }
     },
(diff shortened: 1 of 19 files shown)


packages/data-mate/src/function-configs/string/encode.ts
```diff
@@ -0,0 +1,125 @@
...
+    argument_schema: {
+        algo: {
+            type: FieldType.String,
+            array: false,
+            description: 'Which has hashing algorithm to use'
+        },
+        digest: {
+            type: FieldType.String,
+            array: false,
+            description: 'Which has digest to use, may be set to either "base64" or "hex", defaults to "hex". '
```
ciorg, line 109: typo - `Which digest to use`

packages/data-mate/src/function-configs/string/encodeSHA.ts
```diff
@@ -69,6 +70,12 @@ export const encodeSHAConfig: FieldTransformConfig<EncodeSHAConfig> = {
             description: 'Which has digest to use, may be set to either "base64" or "hex", defaults to "hex"'
```
ciorg, line 70: `which digest to us` - typo, or is it `Which hash digest to use`?

packages/data-mate/src/function-configs/string/encode.ts
ciorg: or is it `Which hash digest to use`?

jsnoble (approved):
LGTM

ciorg (changes requested):

packages/data-mate/src/function-configs/string/encodeSHA.ts
peterdemartini: This should be fixed

packages/data-mate/src/function-configs/string/encode.ts
peterdemartini: This should be fixed

ciorg (approved):
LGTM

PR #2738, opened 2021-05-20
Geo holes

jsnoble:
- fix hole support for geo relation based queries
- added lookupTimezone, timezoneToOffset and getTimezoneOffset


packages/data-mate/src/adapters/argument-validator/index.ts
```diff
@@ -42,13 +43,14 @@ function getType(
             return isBoolean;
         case FieldType.Float:
         case FieldType.Number:
+            return isFloat;
         case FieldType.Byte:
         case FieldType.Short:
         case FieldType.Integer:
             return isNumber;
```
peterdemartini, line 50: `isNumber` would allow a float, also we should probably be more specific about the size of `Byte`, `Short` and `Integer` validation

packages/data-mate/src/adapters/argument-validator/index.ts
```diff
@@ -42,13 +43,14 @@ function getType(
...
         case FieldType.Long:
         case FieldType.Double:
```
peterdemartini, line 52: Since bigints are not floats, a `Double` should just use `isNumber`, and a `Long` should use `isBigInt`. Also we should also accept integers as `Long` to make sure it will work in QPL

packages/data-mate/src/adapters/function-adapter/interfaces.ts
```diff
@@ -12,9 +13,15 @@ export interface FunctionAdapterOptions<T extends Record<string, any>> {
...
 export interface RecordFunctionAdapterOperation {
-    rows<T extends Record<string, any>>(records: T[]): Record<string, unknown>[];
+    rows: (records: Record<string, unknown>[]) => Record<string, unknown>[];
 }
 
 export interface FieldFunctionAdapterOperation extends RecordFunctionAdapterOperation {
     column(values: unknown[]): unknown[];
 }
+
+export interface PartialArgs
```
peterdemartini, line 23: Why is this called `PartialArgs` and why does it include the input config?

packages/data-mate/src/function-configs/interfaces.ts
```diff
@@ -128,16 +129,25 @@ export interface FunctionDefinitionConfig<T extends Record<string, any>> {
     /** Used for additional custom validation of args, called after generic arg validation */
     readonly validate_arguments?: (args: T) => void;
 }
+export interface BaseArgs<T extends Record<string, any> = Record<string, unknown>> {
```
peterdemartini, line 132: This name is too generic

packages/data-mate/src/function-configs/interfaces.ts
```diff
@@ -128,16 +129,25 @@ export interface FunctionDefinitionConfig<T extends Record<string, any>> {
...
+    args: T,
```
peterdemartini, line 133: These arg properties should be marked a readonly

packages/utils/src/numbers.ts
```diff
@@ -97,6 +98,25 @@ export function bigIntToJSON(int: bigint): string|number {
     return (int - BigInt(1)).toString(10);
 }
 
+export function safeSerialize(input: unknown): any {
```
peterdemartini, line 101: Why is this part of numbers? Also this needs some jsdoc describing why this is needed, I am not completely sold on the naming here either. It is not really serializing it since it doesn't output a json string. Maybe, `valueToJSON` would be more appropriate and similar to our data frame code

peterdemartini (changes requested):
I am not sure how to verify some of the geo code so I am just going to trust you.

packages/data-mate/src/function-configs/date/timezoneToOffset.ts
```diff
@@ -0,0 +1,34 @@
...
+} from '../interfaces';
+
+export const timezoneToOffsetConfig: FieldTransformConfig = {
+    name: 'timezoneToOffset',
+    type: FunctionDefinitionType.FIELD_TRANSFORM,
+    process_mode: ProcessMode.INDIVIDUAL_VALUES,
+    category: FunctionDefinitionCategory.DATE,
+    description: `Given a timezone, it will return the offset from UTC in minutes.
+     This uses current server time as the reference for a date, so results may vary
+     depending on day lights savings time adjustments`,
```
ciorg, line 17: daylight savings

packages/data-mate/src/function-configs/string/truncate.ts
```diff
@@ -16,7 +16,7 @@ export const truncateConfig: FieldTransformConfig<TruncateConfig> = {
     type: FunctionDefinitionType.FIELD_TRANSFORM,
     process_mode: ProcessMode.INDIVIDUAL_VALUES,
     category: FunctionDefinitionCategory.STRING,
-    description: 'Truncate a string value',
+    description: 'Limit the size of string to a specific length, if the length is greater than the specified size, the excess is removed',
```
ciorg, line 19: Limit the size of a string

packages/data-mate/src/adapters/function-adapter/interfaces.ts
jsnoble: I made changes to what is passed into the create function, it now takes an objects

peterdemartini: Maybe calling it a `InitialFunctionArgs` then? Maybe jsdoc would be useful to describe why this added

packages/data-mate/src/function-configs/date/timezoneToOffset.ts
```diff
@@ -0,0 +1,34 @@
...
+    accepts: [
+        FieldType.String,
+    ],
+    output_type(inputConfig) {
+        const { field_config } = inputConfig;
+
+        return {
+            field_config: {
+                ...field_config,
+                type: FieldType.Number
```
peterdemartini, line 30: Keep this size as small as possible, we will eventually have memory optimizations for smaller numbers

```suggestion
                type: FieldType.Short
```

packages/data-mate/src/builder/types/IntBuilder.ts
```diff
@@ -11,25 +11,17 @@ export const INT_SIZES = {
...
 
 export class IntBuilder extends Builder<number> {
+    private numValidator: (input: unknown) => number
+
     constructor(
         data: WritableData<number>,
         options: BuilderOptions
     ) {
         super(VectorType.Int, data, options);
+        this.numValidator = validateNumberType(this.config.type as FieldType);
```
peterdemartini, line 21: This can be simplified

```suggestion
        this._valueFrom = validateNumberType(this.config.type as FieldType);
```

peterdemartini (changes requested):

packages/data-mate/src/builder/types/IntBuilder.ts
jsnoble: hmm I tried doing that but I was getting type errors, typescript seemed to think that _valueFrom did not exist properly and was not being implemented properly.

peterdemartini: What is the exact error?

jsnoble: class IntBuilder
Non-abstract class 'IntBuilder' does not implement inherited abstract member '_valueFrom' from class 'Builder<number>'.

jsnoble:
```
export class IntBuilder extends Builder<number> {
    constructor(
        data: WritableData<number>,
        options: BuilderOptions
    ) {
        super(VectorType.Int, data, options);
        this._valueFrom = validateNumberType(this.config.type as FieldType);
    }
}
```

peterdemartini (approved):
LGTM

PR #2739, opened 2021-05-20
Lookup function

ciorg:
lookup function - 

There's a few spots here that I'm not sure of:
  * seems like the utils function should check if the lookup object is an actual object, or should the type determine that?
  * also I have the arg as the lookup object and the input as the key to support the fp style, but that feels backwards to me like the arg should be the key and the input should be the lookup arg

### DIFF
--- a/packages/data-mate/src/function-configs/object/lookup.ts
+++ b/packages/data-mate/src/function-configs/object/lookup.ts
@@ -1,0 +1,63 @@
+import { lookup } from '@terascope/utils';
+import { FieldType } from '@terascope/types';
+import {
+    ProcessMode, FunctionDefinitionType, FunctionDefinitionCategory, FieldTransformConfig,
+    DataTypeFieldAndChildren
+} from '../interfaces';
+
+export interface LookupArgs {
+    readonly lookupObj: Record<string, unknown>
+}
+
+export const lookupConfig: FieldTransformConfig<LookupArgs> = {
+    name: 'lookup',
+    type: FunctionDefinitionType.FIELD_TRANSFORM,
+    process_mode: ProcessMode.INDIVIDUAL_VALUES,
+    category: FunctionDefinitionCategory.OBJECT,
+    description: 'Returns value based on key value lookup ',
+    examples: [
+        {
+            args: { lookupObj: { key1: 'value1', key2: 'value2'}},
+            config: {
+                version: 1,
+                fields: { testField: { type: FieldType.String } }
+            },
+            field: 'testField',
+            input: 'key1',
+            output: 'value1'
+        },
+        {
+            args: { lookupObj: { 123: 4567, 8910:1112 }},
+            config: {
+                version: 1,
+                fields: { testField: { type: FieldType.Number } }
+            },
+            field: 'testField',
+            input: 8910,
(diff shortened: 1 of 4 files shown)


packages/data-mate/src/function-configs/object/lookup.ts
```diff
@@ -0,0 +1,63 @@
+import { lookup } from '@terascope/utils';
+import { FieldType } from '@terascope/types';
+import {
+    ProcessMode, FunctionDefinitionType, FunctionDefinitionCategory, FieldTransformConfig,
+    DataTypeFieldAndChildren
+} from '../interfaces';
+
+export interface LookupArgs {
+    readonly lookupObj: Record<string, unknown>
```
peterdemartini, line 9: We discussed naming this arg `in` so when using this function, it would be `lookup(in: { foo: 'bar' })` which is more natural than `lookup(lookupObj: { foo: 'bar' })`

packages/data-mate/src/function-configs/object/lookup.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export const lookupConfig: FieldTransformConfig<LookupArgs> = {
+    name: 'lookup',
+    type: FunctionDefinitionType.FIELD_TRANSFORM,
+    process_mode: ProcessMode.INDIVIDUAL_VALUES,
+    category: FunctionDefinitionCategory.OBJECT,
+    description: 'Returns value based on key value lookup ',
```
peterdemartini, line 17: This description can be improved, maybe `By matching the input to key in the table, the corresponding value is returned` (I am not compleletly sold on that description but I think it is a better description.)

peterdemartini (changes requested):

packages/data-mate/src/function-configs/object/lookup.ts
ciorg: yeah - I'll think about this a little more

packages/data-mate/src/function-configs/object/lookup.ts
ciorg: updated arg to `in`

packages/data-mate/src/function-configs/object/lookup.ts
ciorg: description is `Matches the input to a key in a table and returns the corresponding value.`

peterdemartini (approved):
LGTM

packages/data-mate/src/function-configs/object/lookup.ts
```diff
@@ -0,0 +1,73 @@
...
+                version: 1,
+                fields: { testField: { type: FieldType.String } }
+            },
+            field: 'testField',
+            input: 'key3',
+            output: undefined
+        }
+    ],
+    create(args: LookupArgs): any {
+        return lookup(args.in);
```
jsnoble, line 51: this needs a 'required_arguments' value to guarantee that "in" is set

jsnoble (changes requested):
This needs more validation to make sure that the arg "in" exists

jsnoble (approved):
LGTM

PR #2741, opened 2021-05-21
Add timezone support to column

peterdemartini:
- [x] Add support for a `DateTuple` to a DataFrame date column
- [x] Fix spelling of ISO 8061 to the correct ISO 8601
- [deferred] Ensure functions can work with this new type


peterdemartini:
I think we are going to defer updating the date functions for now, I am going to mark this as ready for review so we can get this merged

ciorg (approved):
I don't understand all of this, but I didn't see anything that stood out.

jsnoble (approved):
LGTM

PR #2750, opened 2021-05-25
Ensure that DataFrame.appendAll will add varying fields if needed

peterdemartini:
Ensure that `DataFrame.appendAll` will correctly deal with varying fields


jsnoble (approved):
LGTM

Issue #2748, opened 2021-05-25
Feature: Make asset build support javascript assets for bundled assets


godber:
Our recent changes only supported typescript assets.  We need to support javascript assets as well.

godber:
This was resolved in https://github.com/terascope/teraslice/pull/2751

PR #2751, opened 2021-05-25
add support for JS asset bundle refs: 2748

godber:
This implements the JS asset build.

### DIFF
--- a/packages/teraslice-cli/src/helpers/asset-src.ts
+++ b/packages/teraslice-cli/src/helpers/asset-src.ts
@@ -172,7 +172,20 @@ export class AssetSrc {
             // can accept this limitation now.
             let entryPoint = '';
             try {
-                entryPoint = require.resolve(path.join(tmpDir.name, 'asset', 'src', 'index.ts'));
+                if (this.assetPackageJson?.entryPoint) {
+                    // FIXME: why should we treat JS and TS differently?  If I
+                    // do it like this ... we do the same thing two different
+                    // ways:
+                    // * JS - explicitly specify entry point in package.json
+                    // * TS - implicitly find entry point by looking for file
+                    // we should chose one or the other and use the same for
+                    // both cases.
+                    entryPoint = require.resolve(
+                        path.join(tmpDir.name, 'asset', this.assetPackageJson.entryPoint)
+                    );
+                } else {
+                    entryPoint = require.resolve(path.join(tmpDir.name, 'asset', 'src', 'index.ts'));
+                }
                 reply.warning(`entryPoint: ${entryPoint}`);
             } catch (err) {
                 reply.fatal(`Unable to resolve entry point due to error: ${err}`);
--- a/packages/teraslice-cli/src/helpers/asset-src.ts
+++ b/packages/teraslice-cli/src/helpers/asset-src.ts
@@ -166,27 +166,19 @@ export class AssetSrc {
             reply.info('* running yarn --prod --no-progress');
             await this._yarnCmd(path.join(tmpDir.name, 'asset'), ['--prod', '--no-progress']);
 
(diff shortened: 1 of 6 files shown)


godber:
refs: https://github.com/terascope/teraslice/issues/2748

godber:
Please note my FIXME comment.

I think we should just require the `entryPoint` for both JS and TypeScript.  Alternatively, we can assume a specific path for JS.  Regardless, we should do one or the other, not both under different conditions.  I'd be happy to go back and add the entryPoints to the public assets.

godber:
On drawback of this whole static registry thing is that asset developers can write assets, which include passing unit tests, that still fail to include processors because they have forgotten to include them in the registry.

godber:
> Please note my FIXME comment.
> 
> I think we should just require the `entryPoint` for both JS and TypeScript. Alternatively, we can assume a specific path for JS. Regardless, we should do one or the other, not both under different conditions. I'd be happy to go back and add the entryPoints to the public assets.

After thinking overnight, I think that since we are now requiring that asset registries be implemented we can easily extend that requirement as follows ...

"All assets must implement registries that provide a single entrypoint to all of the processors/functionality contained in that asset.  If the asset is implemented in Javascript, that registry must be `asset/index.js` and if the asset is implemented in TypeScript, the registry must be `asset/src/index.ts`."

If we require this, then we can make the assumption that one of those index files will be present.  We can eliminate the `entryPoint` (or `srcMain`) property in the `package.json`.

This is what I will do.

godber:
I haven't run bump yet, but I think this is ready.

packages/teraslice-cli/src/helpers/asset-src.ts
```diff
@@ -166,14 +166,19 @@ export class AssetSrc {
...
-            // can accept this limitation now.
+            // Since we now require bundled assets to implement a registry, we
+            // require that Javascript assets place it at `asset/index.js` and
+            // TypeScript assets place it at `asset/src/index.ts`
             let entryPoint = '';
             try {
-                entryPoint = require.resolve(path.join(tmpDir.name, 'asset', 'src', 'index.ts'));
-                reply.warning(`entryPoint: ${entryPoint}`);
+                if (await fs.pathExists(path.join(tmpDir.name, 'asset', 'index.js'))) {
+                    entryPoint = require.resolve(path.join(tmpDir.name, 'asset', 'index.js'));
```
peterdemartini, line 175: This doesn't need to be changed but `path.join` should be sufficient and `require.resolve` does not add any value here.

peterdemartini (changes requested):
This looks good for the most part, there are a couple log messages (one was previously already there) but I think they could be improved by not including the tmp directory since it makes it more difficult to understand.

Screen Shot 2021-05-28 at 7 49 09 AM

packages/teraslice-cli/src/helpers/asset-src.ts
godber: The main point of the logs is to provide the temp directory just in case something goes wrong the user can find the temp directories and hopefully track down the problem.  If I removed the temp directories the logs would always be one of two possibilities:

```
* entryPoint: asset/src/index.ts
* doing a test require of index.js
```

or


```
* entryPoint: asset/index.js
* doing a test require of index.js
```

peterdemartini: The entryPoint line is more clear without the temp directory and the temp directory doesn't add any value there. The require test line is maybe a different story, I personally think it would be better to output the asset registry, if the require test fails it should definitely output the full file path so it can be debugged

godber: Yeah, let me remove that and confirm it still works.

peterdemartini (approved):

PR #2755, opened 2021-05-27
Documentation changes

ciorg:
* Tried to use same verbiage for functions
    * basic formula was `Returns x if y` where applicable
* Tried to simplify where obvious 
* Found some typos, there are probably more
* Focused on the function descriptions and args
* Changed all the `set` date function arg to `value` instead of `second` or `year`, etc...


peterdemartini (approved):
LGTM

jsnoble (approved):
LGTM

PR #2760, opened 2021-05-27
Dynamic arg support

jsnoble:
- allow dynamic arg generation by specifying args as a function
- moved arg type interfaces so that they can be exported


packages/data-mate/src/adapters/data-frame-adapter/index.ts
```diff
@@ -1,21 +1,22 @@
...
     FieldValidateConfig, DataTypeFieldAndChildren, FunctionDefinitionConfig,
-    FunctionContext, TransformContext
+    FunctionContext, DynamicFrameFunctionContext
 } from '../../function-configs/interfaces';
 import { Builder } from '../../builder';
 import { WritableData } from '../../core';
 
 export interface DataFrameAdapterOptions<T extends Record<string, any>> {
-    args?: T,
+    args?: T | ((index: number, column: Column<unknown>) => T),
```
peterdemartini, line 19: I am not sure if we want to just have column here, it would be more useful to have the whole data frame if available. For our use case though, we probably won't even need this argument.

packages/data-mate/src/adapters/function-adapter/interfaces.ts
```diff
@@ -1,11 +1,12 @@
 import { DataTypeConfig } from '@terascope/types';
+import { FunctionContext, DynamicFunctionContext } from '../../function-configs/interfaces';
 
 export interface FunctionAdapterOptions<T extends Record<string, any>> {
     /**
      * Required with using the data field config
     */
     readonly field?: string,
-    readonly args?: T,
+    readonly args?: T | ((index: number, column: unknown[]) => T),
```
peterdemartini, line 9: I am not sure if we want to just have column here, it would be more useful to have the whole data frame if available. For our use case though, we probably won't even need this argument.

ciorg (approved):
LGTM

sorry you had to move all those small interfaces!

packages/data-mate/src/function-configs/interfaces.ts
```diff
@@ -1,9 +1,309 @@
...
+    DateFormat, GeoInput, GeoPointInput, GeoShapeRelation,
+    TimeResolution
 } from '@terascope/types';
+import { BinaryToTextEncoding } from 'crypto';
+import validator from 'validator';
 import { Column } from '../column';
 
+export type { AdjustDateArgs, GetTimeBetweenArgs, InNumberRangeArg } from '@terascope/utils';
+
+export interface FormatDateArgs {
```
peterdemartini, line 13: Lets either move these arg definitions to a different interfaces file or export them from the individual function files using:

`export { FormatDateArgs } from './formatDate';`

peterdemartini (changes requested):

packages/utils/src/ip.ts
```diff
@@ -2,14 +2,26 @@ import _isIP from 'is-ip';
...
 import { toInteger, isNumberLike, toBigIntOrThrow } from './numbers';
+import { getTypeOf } from './deps';
 
-export function isIP(input: unknown): boolean {
+export function isIP(input: unknown): input is FieldType.IP {
     return isString(input) && _isIP(input);
 }
 
+export function isIPOrThrow(input: unknown): FieldType.IP {
+    const ipValue = primitiveToString(input);
```
peterdemartini, line 16: I don't think we want to call this here

packages/utils/src/ip.ts
```diff
@@ -2,14 +2,26 @@ import _isIP from 'is-ip';
...
 import ip6addr from 'ip6addr';
+import { FieldType } from '@terascope/types';
 import validateCidr from 'is-cidr';
-import { isString } from './strings';
+import { isString, primitiveToString } from './strings';
 import { toInteger, isNumberLike, toBigIntOrThrow } from './numbers';
+import { getTypeOf } from './deps';
 
-export function isIP(input: unknown): boolean {
+export function isIP(input: unknown): input is FieldType.IP {
```
peterdemartini, line 11:
```suggestion
export function isIP(input: unknown): input is string {
```

packages/utils/src/ip.ts
```diff
@@ -2,14 +2,26 @@ import _isIP from 'is-ip';
...
     return isString(input) && _isIP(input);
 }
 
+export function isIPOrThrow(input: unknown): FieldType.IP {
```
peterdemartini, line 15:
```suggestion
export function isIPOrThrow(input: unknown): string {
```

packages/data-mate/src/adapters/argument-validator/index.ts
```diff
@@ -103,7 +36,7 @@ export function validateFunctionArgs<T extends Record<string, any>>(
 
         for (const [field, typeConfig] of Object.entries(fnDef.argument_schema)) {
             if (Object.hasOwnProperty.call(config, field)) {
-                const typeValidator = getType(typeConfig);
+                const typeValidator = getValidatorForFieldType(typeConfig);
```
peterdemartini, line 39: This needs to actually coerce the argument values, currently this is just used as a validator. Just to note, this should create new copy of the arguments and NOT mutate the input

packages/data-mate/src/function-configs/json/cast.ts
```diff
@@ -0,0 +1,108 @@
...
+    description?: string;
+    indexed?: boolean;
+    locale?: string;
+    format?: string;
+    is_primary_date?: boolean;
+    time_resolution?: TimeResolution;
+    child_config?: DataTypeFields
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
peterdemartini, line 28: Why is this disabled here?
```suggestion
```

packages/data-mate/src/function-configs/json/cast.ts
```diff
@@ -0,0 +1,108 @@
...
+const castConfig: FieldTransformConfig<CastArgs> = {
+    name: 'cast',
+    type: FunctionDefinitionType.FIELD_TRANSFORM,
+    process_mode: ProcessMode.FULL_VALUES,
+    category: FunctionDefinitionCategory.JSON,
+    description: 'Converts the field type of the entity to another type',
+    create() {
+        return (input: unknown) => {
```
peterdemartini, line 36: This implementation should just be noop (a function that returns the input)

packages/data-mate/src/function-configs/json/cast.ts
```diff
@@ -0,0 +1,108 @@
...
+    child_config?: DataTypeFields
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const castConfig: FieldTransformConfig<CastArgs> = {
+    name: 'cast',
+    type: FunctionDefinitionType.FIELD_TRANSFORM,
+    process_mode: ProcessMode.FULL_VALUES,
+    category: FunctionDefinitionCategory.JSON,
+    description: 'Converts the field type of the entity to another type',
```
peterdemartini, line 34:
```suggestion
    description: 'Converts the field to another type, this is also useful changing the metadata of field',
```

packages/data-mate/src/function-configs/json/cast.ts
```diff
@@ -0,0 +1,108 @@
...
+        child_config: {
+            type: FieldType.Object,
+            description: 'If parsing an object, you can specify the DataTypeFields of the key/values of the object.'
+            + ' This is an object whose keys are the name of the fields, whose value is an object with all of the other properties listed above'
+            + ' (ie type, array, locale, format but not child_config).'
+        }
+    },
+    validate_arguments(args) {
+        if (args && args.child_config) {
+            for (const [field, config] of Object.entries(args.child_config)) {
```
peterdemartini, line 87: This should probably be validated using using a function from the data types package

packages/data-mate/src/function-configs/json/cast.ts
```diff
@@ -0,0 +1,108 @@
...
+    indexed?: boolean;
+    locale?: string;
+    format?: string;
+    is_primary_date?: boolean;
+    time_resolution?: TimeResolution;
+    child_config?: DataTypeFields
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const castConfig: FieldTransformConfig<CastArgs> = {
```
peterdemartini, line 29: Shouldn't this be exported? It is possible we are missing the reference to this in the index.ts file too

packages/data-mate/src/function-configs/json/cast.ts
```diff
@@ -0,0 +1,108 @@
...
+} from '@terascope/types';
+import {
+    FieldTransformConfig,
+    ProcessMode,
+    FunctionDefinitionType,
+    DataTypeFieldAndChildren,
+    FunctionDefinitionCategory,
+} from '../interfaces';
+
+export interface CastArgs {
```
peterdemartini, line 16:
```suggestion
export type CastArgs = DataTypeFieldConfig & { child_config?: DataTypeFields }
```

packages/data-mate/test/function-configs/dynamic-args-spec.ts
```diff
@@ -0,0 +1,267 @@
...
+
+                const api = dataFrameAdapter(getTimezoneOffsetConfig, { args: dynamicArgs, field });
+                const newFrame = api.frame(frame);
+
+                expect(newFrame.getColumn(field)?.toJSON()).toEqual(expectedResults);
+            });
+        });
+    });
+
+    // describe('with functionAdapter', () => {
```
peterdemartini, line 103: We should remove the commented tests

packages/data-mate/test/function-configs/functionTestHarness.ts
```diff
@@ -40,67 +39,69 @@ export function functionTestHarness<T extends Record<string, any>>(
...
-                            field: testCase.field,
-                            config: testCase.config
-                        }).rows([input]);
-                    }).toThrowError(
-                        testCase.output ? String(testCase.output) : undefined
-                    );
-                }
-            });
-        });
+        // disabling tests till we have a clearer use case
```
peterdemartini, line 42: We should remove these because by the time we need this to work, the code will be out of sync what it should be and distract us from getting it done right. It would be better to use git for a record of this change

packages/data-mate/test/function-configs/geo/toGeoPoint-spec.ts
```diff
@@ -41,157 +36,166 @@ describe('toGeoPointConfig', () => {
...
-
-                        return accum;
-                    },
-                    []
-                );
-                expect(api.rows(rows)).toEqual(results);
-            });
-        });
-    });
+    // describe('when paired with fieldFunctionAdapter', () => {
```
peterdemartini, line 52: We should remove the commented code

packages/utils/src/field-type.ts
```diff
@@ -0,0 +1,415 @@
...
+
+    return getTransformerForFieldType<T>(fieldConfig, childConfig) as CoerceFN<T>;
+}
+
+// We convert Int, Byte and Short to Integers
+function _shouldConvertToInteger(type: FieldType) {
+    return [FieldType.Integer, FieldType.Byte, FieldType.Short].includes(type);
+}
+
+export function isIPOrThrow(input: unknown): string {
```
peterdemartini, line 64: This should be the `./ip.ts` file, and also it is missing some jsdoc

packages/utils/src/field-type.ts
```diff
@@ -0,0 +1,415 @@
...
+} from './arrays';
+import { getTypeOf, isPlainObject } from './deps';
+import { isFunction } from './functions';
+import { isNotNil } from './empty';
+
+export const HASH_CODE_SYMBOL = Symbol('__hash__');
+
+type CoerceFN<T = unknown> = (input: unknown) => T
+
+export function coerceToType<T = unknown>(
```
peterdemartini, line 38: This is missing some jsdoc

packages/utils/src/field-type.ts
```diff
@@ -0,0 +1,415 @@
...
+            return coerceToGeoPoint;
+        case FieldType.GeoJSON:
+            return coerceToGeoJSON;
+        case FieldType.Object:
+            return coerceToObject(argFieldType, childConfig);
+        case FieldType.Tuple:
+            return coerceToTuple(argFieldType, childConfig);
+        default:
+            // equivalent to an any-builder
+            return (input) => input;
```
peterdemartini, line 358: This would also potentially miss configurations to pass through, might be better to explicitly add a case for `FieldType.Any` and then throw in the default case to avoid a missing field type case

packages/utils/src/field-type.ts
```diff
@@ -0,0 +1,415 @@
...
+        case FieldType.Double:
+            return toNumberOrThrow;
+        case FieldType.Byte:
+            return coerceToNumberType(FieldType.Byte);
+        case FieldType.Short:
+            return coerceToNumberType(FieldType.Short);
+        case FieldType.Integer:
+            return coerceToNumberType(FieldType.Integer);
+        case FieldType.Long:
+            return toBigIntOrThrow;
```
peterdemartini, line 345: It seems to me like `coerceToNumberType` should either be renamed to `coerceIntergerType` or be changed to handle all number types.

```suggestion
        case FieldType.Float:
        case FieldType.Number:
        case FieldType.Double:
        case FieldType.Byte:
        case FieldType.Short:
        case FieldType.Integer:
        case FieldType.Long:
            return coerceToNumberType(argFieldType.type);
```

packages/utils/src/ip.ts
```diff
@@ -5,11 +5,20 @@ import ip6addr from 'ip6addr';
...
 import { isString } from './strings';
 import { toInteger, isNumberLike, toBigIntOrThrow } from './numbers';
+import { getTypeOf } from './deps';
 
-export function isIP(input: unknown): boolean {
+export function isIP(input: unknown): input is string {
     return isString(input) && _isIP(input);
 }
 
+export function isIPRangeOrThrow(input: unknown): string {
```
peterdemartini, line 14: This is missing jsdoc

peterdemartini (changes requested):

packages/utils/src/field-type.ts
jsnoble: I first had it in the ip file, and then you stated `I don't think we want to call this here`, so I moved it back to fieldType. If your meaning was something else what was it then?

peterdemartini: Ahh yes, I should be more specific, I was referring to the `primitiveToString` function call since we only allow strings as input it was useless call

packages/data-mate/test/function-configs/dynamic-args-spec.ts
```diff
@@ -0,0 +1,102 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
peterdemartini, line 1: Do we need this eslint comment?

peterdemartini (changes requested):
This is close, there is still the argument validator change that is needed

peterdemartini (approved):
Overall I think this looks good there is just one comment regarding an eslint ignore

ciorg (approved):
LGTM

PR #2765, opened 2021-05-28
Revert "Bump typescript from 4.2.4 to 4.3.2"

peterdemartini:
Reverts terascope/teraslice#2757


peterdemartini:
Typescript 4.3 doesn't work with typedoc yet so we will need to revert upgrading this

PR #2790, opened 2021-06-07
Allow dev docker image override

peterdemartini:
Add support for overriding the docker dev image and registry to ts-scripts


jsnoble (approved):
LGTM

PR #2799, opened 2021-06-08
v0.77.1 - Add better error handling for unhandled errors or rejections

peterdemartini:
- Add better exit log message with status code
- Ensure that the slice gets marked as failed if there is an expected error
- Ensure that older versions of node work on newer versions of teraslice


godber (approved):
This looks good.

PR #2803, opened 2021-06-09
Date tuple

ciorg:
* added DateTuple support to date functions in utils and function-configs in data-mate and respective tests


packages/data-mate/src/function-configs/date/getMonth.ts
```diff
@@ -7,7 +7,7 @@ import {
 export const getMonthConfig: FieldTransformConfig = {
     name: 'getMonth',
     type: FunctionDefinitionType.FIELD_TRANSFORM,
-    process_mode: ProcessMode.INDIVIDUAL_VALUES,
+    process_mode: ProcessMode.FULL_VALUES,
```
peterdemartini, line 10: This should be `INDIVIDUAL_VALUES`

packages/data-mate/src/column/utils.ts
```diff
@@ -40,9 +41,12 @@ export function mapVectorEachValue<T, R = T>(
     builder: Builder<R>,
     fn: (value: T, index: number) => Maybe<R>,
 ): Vector<R> {
+    const containsArray = vector.type === VectorType.Tuple || vector.config.array
+        || vector.type === VectorType.Any;
+
     function _mapValue(value: T|readonly Maybe<T>[], index: number): Maybe<R>|readonly Maybe<R>[] {
-        if (isArrayLike<readonly Maybe<T>[]>(value)) {
-            return value.map((v): Maybe<R> => (
+        if (containsArray && value != null) {
```
peterdemartini, line 48: To support the `Any` we need this logic, since `Any` doesn't have to be any array
```suggestion
        if (containsArray && Array.isArray(value)) {
```

packages/types/src/dates.ts
```diff
@@ -16,3 +16,11 @@ export type TimeBetweenIntervals = 'milliseconds' | 'seconds' | 'minutes' | 'hou
  * (minutes relative to UTC)
 */
 export type DateTuple = readonly [epoch_millis: number, timezone: number];
+
+export type DateTypes = Date | string | number | DateTuple;
```
peterdemartini, line 20: Not a big fan of this name. It could also use some jsdoc describing what it is used for.

```suggestion
export type DateInputType = Date | string | number | DateTuple;
```

peterdemartini (changes requested):
This also needs to be updated from the feature branch

packages/utils/src/dates.ts
```diff
@@ -110,6 +116,28 @@ export function getValidDateOrThrow(val: unknown): Date {
     return date;
 }
 
+/**
+ * Returns a valid date with the timezone applied or throws{@see getValidDate}
+ */
+export function getValidDateWithTimezoneOrThrow(val: unknown): Date {
+    if (isDateTuple(val)) {
+        return getValidDateOrThrow(toISO8601(val));
```
peterdemartini, line 124: Converting this to ISO8601 string is excessive work. Doing something like `new Date(input[0] + (input[1] * 60_000))` would be less work and easier to process

peterdemartini (approved):
LGTM, except I have one comment

jsnoble (approved):
LGTM besides peters comments

PR #2807, opened 2021-06-10
WIP: added string entropy to function-configs

jsnoble:
- added entropy directive (defaults to use shannon)
-

### DIFF
--- a/packages/utils/src/strings.ts
+++ b/packages/utils/src/strings.ts
@@ -596,3 +596,44 @@ export function joinList(
         return `${acc}${sep} ${curr}`;
     }, '');
 }
+
+export type StringEntropyFN = (input: unknown) => number
+
+function shannonEntropy(input: unknown) {
+    if (!isString(input)) {
+        throw new Error(`Invalid input ${input}, must be of type String`);
+    }
+
+    return [...new Set(input)]
+        .map(
+            (chr) => {
+                const matching = input.match(new RegExp(chr, 'g')) as RegExpMatchArray;
+                return matching.length;
+            }
+        )
+        .reduce((sum, frequency) => {
+            const p = frequency / input.length;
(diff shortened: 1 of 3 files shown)


packages/utils/src/strings.ts
```diff
@@ -596,3 +596,58 @@ export function joinList(
...
+        throw new Error(`Invalid input ${input}, must be of type String`);
+    }
+
+    let sum = 0;
+    const len = input.length;
+    const dict: Record<string, number> = Object.create(null);
+
+    // get number of chars per string
+    for (const char of input) {
+        if (dict[char]) {
```
peterdemartini, line 615:
```suggestion
        if (dict[char] != null) {
```

packages/utils/src/strings.ts
```diff
@@ -596,3 +596,58 @@ export function joinList(
...
+            dict[char]++;
+        } else {
+            dict[char] = 1;
+        }
+    }
+
+    for (const num of Object.values(dict)) {
+        const p = num / len;
+        const pLogCalc = p * Math.log(p);
```
peterdemartini, line 624: The previous implementation used `log2`, why the change?

peterdemartini (changes requested):

packages/utils/src/strings.ts
```diff
@@ -612,7 +612,7 @@ export function shannonEntropy(input: unknown): number {
 
     // get number of chars per string
     for (const char of input) {
-        if (dict[char]) {
+        if (dict[char] !== null) {
```
peterdemartini, line 615: This should be `!=` not `!==`

peterdemartini (approved):
LGTM as long as you fix the `!=` usage

ciorg (approved):
LGTM

PR #2808, opened 2021-06-10
function export check

ciorg:
* added maxValues to numeric index.ts
* added test to check that all function configs are exported

### DIFF
--- a/packages/data-mate/test/function-configs/function-configs-spec.ts
+++ b/packages/data-mate/test/function-configs/function-configs-spec.ts
@@ -1,5 +1,5 @@
 import 'jest-extended';
-import fs from 'fs-extra';
+import { promises as fsp } from 'fs';
 import path from 'path';
 import { uniq } from '@terascope/utils';
 import {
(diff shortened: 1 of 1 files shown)


packages/data-mate/test/function-configs/function-configs-spec.ts
```diff
@@ -23,3 +25,38 @@ describe('function configs', () => {
...
+
+    for (const item of configDirs) {
+        // ignore non-directories
+        if (item.includes('.ts')) continue;
+
+        const functionPath = path.join(dirPath, item);
+
+        const imports = parseIndexFile(path.join(functionPath, 'index.ts'));
+
+        const configFiles = fs.readdirSync(functionPath).filter((i) => !(i.includes('utils') || i === 'index.ts'));
```
peterdemartini, line 42: We should use async functions here to not block the event loop

```suggestion
        const configFiles = fs.readdirSync(functionPath).filter((i) => !(i.endsWith('utils.ts') || i === 'index.ts'));
```

peterdemartini (changes requested):
LGTM

peterdemartini (approved):

jsnoble (approved):
LGTM

Issue #2769, opened 2021-06-02
Dynamically generate entrypoints for JS assets


godber:
I think we can do something like this to dynamically generate the asset registry for Teraslice assets:

```
'use strict';

const glob = require('glob');

glob('**/{processor,schema}.js', { ignore: ['node_modules/**', '__lib/**'] }, (er, files) => {
    console.log(files);
});
```

This would force the following requirements on these assets:

* All asset processor code be in `processor.js` or `schama.js` (I think this is a fixed, well known set of files, right? I haven't checked the loader code).
* All helper code would need to be in `__lib/` or be certain to not contain `.js` files that match those listed above

If the asset registry is ONLY used to supply esbuild entrypoints, then I think this should work.  Code like this could be added to the `teraslice-cli asset build` code to populate entrypoints.  If the registry is used in some other way, then perhaps this will be insufficient.

peterdemartini:
I don't like the idea of this, mainly because this file won't exist until build time leading problems that won't be caught in CI. I think we'd be better off having a `teraslice-cli assets init --registry` command that would generate it for you. This way it will be ran as needed so the file will be checked-in and tested via the test-harness. Plus since it uses `yeoman` we get file-system diffing capabilities when the file already exists.

peterdemartini:
As for some of the file conventions, ignore a directory or filename that starts with `__`, `.` or equals `node_modules`. And yes there is a strict file naming convention so requiring the files like that should be fine

godber:
> I don't like the idea of this, mainly because this file won't exist until build time leading problems that won't be caught until CI publish step. I think we'd be better off having a `teraslice-cli assets init --registry` command that would generate it for you. This way it will be ran as needed so the file will be checked-in and tested via the test-harness. Plus since it uses `yeoman` we get file-system diffing capabilities when the file already exists.

That solution had not occurred to me and it seems like a good idea ... it adds friction to the asset development workflow though.  Would we expect that the generated `index.js` be committed to git?

godber:
Is the registry used for anything other than providing an entrypoint for `esbuild`?

godber:
BTW, this is motivated by two main things:

* avoiding writing boilerplate code
* (more importantly) help prevent mistakes where processors are accidentally omitted from a manually generate registry.

peterdemartini:
>  ... it adds friction to the asset development workflow though.

Yes it would be a little bit more work for development but overall I think it gives the developer more control.

[...]

PR #2810, opened 2021-06-11
Add asset registry generator

godber:
This is still mostly a sketch with only bits implemented, but I think I can pull this together soon.

I am trying to dynamically generate the asset registry.

This addresses #2769


godber:
What are the names of these things:

https://github.com/terascope/file-assets/blob/master/asset/src/index.ts#L40-L49

```
export const ASSETS = {
            file_exporter: {
                Processor: FileBatcher,
                Schema: FileExporterSchema
            },
            file_reader: {
                Fetcher: FileFetcher,
                Schema: FileReaderSchema,
                Slicer: FileSlicerOperation
```

Do these things have generic names?
* `file_exporter` =?  Operation? Processor?
*  `FileBatcher` =? `FileExporterSchema` ... the things UNDER the `file_exporter` ... do they have a name?

peterdemartini:
I am not sure how to answer that but maybe a simple example will be more helpful (it will need to be changed to work with multiple operations and the files.


```ts
// NOTE toPascalCase comes from @terascope/utils
import ${toPascalCase(`${op_directory}_${op_file_name}`)} from `${op_directory}/${op_file_name}`;

export const ASSETS = {
       [op_directory]: {
            [toPascalCase(op_file_name)]:  ${toPascalCase(`${op_directory}_${op_file_name}`)},
       }
}
```

```js
// NOTE toPascalCase comes from @terascope/utils
const ${toPascalCase(`${op_directory}_${op_file_name}`)} = require(`${op_directory}/${op_file_name}`);

module.exports = {
      ASSETS: {
            [op_directory]: {
                 [toPascalCase(op_file_name)]:  ${toPascalCase(`${op_directory}_${op_file_name}`)},
            }
     }
}
```

We only care about a few select file name combinations in a operation, the easiest way for you would be to just find any file with the name `processor`, `schema`, `api`, `fetcher`, or `slicer`.

peterdemartini:
If you have any more questions we can do a call

godber:
@ciorg this is mostly functional I think ... I've only tried it on one internal asset, usage as follows:

```
node ./bin/teraslice-cli.js assets init --registry --base-dir /Users/godber/Workspace/asset_dir
```

There are a few cleanup things (console.error -> reply.foo) and a few other improvements.

Feel free to take this over and improve or just use this branch temporarily.  If nothing else, this PR serves as a good start.

packages/teraslice-cli/src/generators/new-processor/index.ts
```diff
@@ -69,7 +69,7 @@ export default class extends Generator {
 
         this.fs.copyTpl(this.templatePath(`base-op/${type}.js`),
             this.destinationPath(`asset/${name}/processor.js`),
-            { name: capitolizeFirstLetter(camelCase(name)) });
+            { name: capitalizeFirstLetter(camelCase(name)) });
```
peterdemartini, line 72: This should just use `toPascalCase` from `@terascope/utils`

packages/teraslice-cli/package.json
```diff
@@ -44,8 +44,11 @@
         "esbuild": "^0.12.7",
         "execa": "^5.1.0",
         "fs-extra": "^9.1.0",
+        "glob": "^7.1.7",
+        "glob-promise": "^4.2.0",
         "got": "^11.8.2",
         "js-yaml": "^4.1.0",
+        "lodash.set": "^4.3.2",
```
peterdemartini, line 51: We shouldn't need this since this adds an extra dependency, just `set` from `@terascope/utils` which is the same thing

ciorg:
* added functionality to copy `__state_assets` dir if it's in the top level asset directory for bundled assets
* tested this  with an internal asset and job and it worked, both generating the registry and coping the static assets

packages/teraslice-cli/generator-templates/registry/index.ejs
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
```
peterdemartini, line 2: We should disable eslint for the file:

```suggestion

/* eslint-disable */
```

packages/teraslice-cli/package.json
```diff
@@ -61,6 +63,7 @@
         "@types/decompress": "^4.2.3",
         "@types/easy-table": "^0.0.32",
         "@types/js-yaml": "^4.0.1",
+        "@types/lodash.set": "^4.3.6",
```
peterdemartini, line 66: We don't need this dependency

packages/teraslice-cli/src/generators/registry/index.ts
```diff
@@ -0,0 +1,43 @@
...
+    async default(): Promise<void> {
+        let asset;
+        let registry;
+
+        // find operators in asset
+        try {
+            asset = new AssetSrc(this.options.asset_path);
+            registry = await asset.generateRegistry();
+        } catch (error) {
+            throw new Error(`ERROR: ${error}`);
```
peterdemartini, line 30: This try catch is not adding any value

packages/teraslice-cli/src/generators/registry/index.ts
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        // generate registry object
+        try {
+            this.fs.copyTpl(this.templatePath('index.ejs'), this.destinationPath('index.js'),
+                {
+                    registry
+                });
+        } catch (error) {
+            console.error(`ERROR: ${error}`);
```
peterdemartini, line 40: Why does it do this?

packages/teraslice-cli/src/helpers/asset-src.ts
```diff
@@ -3,16 +3,33 @@ import prettyBytes from 'pretty-bytes';
...
     bytes: string;
 }
 
+interface AssetRegistry {
+    [dir_name: string]: {
+        [property in OpType]: string
+    }
+}
+
+type OpType = 'Api' | 'Fetcher' | 'Processor' | 'Schema' | 'Slicer';
```
peterdemartini, line 31:
```suggestion
type OpType = 'API' | 'Fetcher' | 'Processor' | 'Schema' | 'Slicer';
```

packages/teraslice-cli/src/helpers/asset-src.ts
```diff
@@ -95,9 +109,53 @@ export class AssetSrc {
...
+     *   api.js
+     *   fetcher.js
+     *   processor.js
+     *   schema.js
+     *   slicer.js
+     * @returns {Array} array of paths to all of the operator files
+     */
+    async operatorFiles(): Promise<string[]> {
+        const matchString = path.join(this.srcDir, 'asset', '**/{api,fetcher,processor,schema,slicer}.js');
+        return glob(matchString, { ignore: ['**/node_modules/**', '**/__lib/**'] });
```
peterdemartini, line 123:
```suggestion
        return glob(matchString, { ignore: ['**/node_modules/**', '**/_*/**', '**/.*/**'] });
```

packages/teraslice-cli/src/helpers/utils.ts
```diff
@@ -30,6 +30,10 @@ export function camelCase(str: string): string {
         .replace(/\W+(.)/g, (match, chr) => chr.toUpperCase());
 }
 
+export function capitalizeFirstLetter(string: string): string {
```
peterdemartini, line 33: We should remove this since it is a duplicate to `firstToUpper` in `@terascope/utils`

peterdemartini (changes requested):

peterdemartini (commented):
I manually tested this and this LGTM, I have a couple of minor comments that can optional be addressed

peterdemartini (approved):
Sorry I meant to approve that (with additional suggestions)

peterdemartini:
Don't forget to bump too

peterdemartini:
Don't forget to update from master before bumping

packages/teraslice-cli/src/cmds/assets/init.ts
```diff
@@ -22,27 +24,43 @@ export = {
...
         const assetBaseDir = cliConfig.args.baseDir;
-        // if just adding a new processor AssetBaseDir needs to have an asset dir
-        if (argv.proc && !fs.pathExistsSync(path.join(assetBaseDir, 'asset'))) {
-            reply.fatal('Execute the command in the base directory of an asset or use the --base-dir with the asset\'s full path');
-        }
 
         try {
-            if (argv.proc) {
+            if (cliConfig.args.proc) {
+                // FIXME: manually verify that this check behaves the same after my change here.
```
godber, line 48: I never double checked this.  Did you verify that the other modes work as well?  e.g. `init` without `--registry`?

peterdemartini: When testing this PR, I found https://github.com/terascope/teraslice/issues/2824. However I thought it was unrelated to these changes

PR #2811, opened 2021-06-11
Utils fix

jsnoble:
- remove usage of node crypto usage in utils, now it uses crypto-js which is browser compatible.


ciorg (approved):
LGTM

PR #2815, opened 2021-06-14
Data mate functions

jsnoble:
- [x] Add Cast function config
- [x] Fix SetTimezone to take in a timezone as well as an offset
- [ ] Add tests for timezoneToOffset

### DIFF
--- a/packages/data-mate/test/function-configs/function-configs-spec.ts
+++ b/packages/data-mate/test/function-configs/function-configs-spec.ts
@@ -10,7 +10,7 @@ import { functionTestHarness } from './functionTestHarness';
 
 describe('function configs', () => {
     Object.entries(functionConfigRepository).forEach(([key, fnDef]) => {
-        if (key === 'setTimezone') functionTestHarness(fnDef as FunctionDefinitionConfig<any>, key);
+        functionTestHarness(fnDef as FunctionDefinitionConfig<any>, key);
     });
 
     it('should not have any duplicate names', () => {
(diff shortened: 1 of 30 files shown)


packages/data-mate/test/function-configs/function-configs-spec.ts
```diff
@@ -10,7 +10,7 @@ import { functionTestHarness } from './functionTestHarness';
 
 describe('function configs', () => {
     Object.entries(functionConfigRepository).forEach(([key, fnDef]) => {
-        functionTestHarness(fnDef as FunctionDefinitionConfig<any>, key);
+        if (key === 'setTimezone') functionTestHarness(fnDef as FunctionDefinitionConfig<any>, key);
```
peterdemartini, line 13:
```suggestion
        functionTestHarness(fnDef as FunctionDefinitionConfig<any>, key);
```

peterdemartini: I have seen you leave this in a few times, I strongly encourage you to use the `-t` flag when running the test so, `yarn test:debug test/function-configs/function-configs-spec.ts -t 'setTimezone'` or `yarn test:watch test/function-configs/function-configs-spec.ts -t 'setTimezone'`. Running tests via jest watch is very powerful and it is very quick to add additional filters by test name, run only the failed tests, or switch to a different file, just type `w` when all the tests have complete in watch mode for help options

jsnoble:
I am having a hard time coming up with a way to tests this function as it innately relies on side effects that change over the span of a year, and changes based off of where the tests are being run.

This will become a little more reliable when the servers are run in Arizona, but the CI and others could run this test in any place as this is open source. I am not sure if there is a proper way to tests this function. 

Also, our interface to it is very simplistic, so it would mainly be testing the function from another library itself, which I am not sure is needed

peterdemartini (approved):
LGTM

Issue #2819, opened 2021-06-16
Jest remove success flag if no tests are run/found


lesleydreyer:
If you run a test file that does not exist, it gives you a 'success: All tests completed' message at the bottom.

PR #2825, opened 2021-06-17
Fix no tests in ts-scripts

peterdemartini:
- Only download the bundled assets in the e2e tests - this should speed them up
- Fix validation of `JEST_MAX_WORKERS` env
- Don't allow the tests to pass if none ran
- Fixes to EventLoop when logger gets deferenced
- Make tree shaking a little easier when requiring date-fns

Closes https://github.com/terascope/teraslice/issues/2819


jsnoble (approved):
LGTM

PR #2865, opened 2021-07-06
Fix index change detection in elasticsearch-store

peterdemartini:
- Update tests to have an indexed nested property
- Fix spelling in index-model-spec
- Correctly detect index field mapping changes, mainly when a nested field changes


peterdemartini:
I tested this in our internal repo and it seems to have detected the changes correctly

jsnoble (approved):
LGTM

PR #2882, opened 2021-07-14
Smarter split, error message fix and round toFahrenheit and toCelsius to 2 decimal places

### LINKED ISSUE #2852: DataMate toFahrenheit or toCelsius should return only 2 decimal places
When calling `toFahrenheit` or `toCelsius` in data mate, you get back too many decimal places since most the time you don't care about 10 decimal places, just 2 to get a rough idea.

peterdemartini:
- Be smarter about how we split strings as performance improvement (I am sure it is minor but it easy to add and might work better with larger strings)
- Fix error message in `toIntergerOrThrow`
- Round `toFahrenheit` and `toCelsius` to 2 decimal places

Closes #2852


jsnoble (approved):
LGTM

PR #2887, opened 2021-07-20
Teraslice cli asset error message improvement

godber:
I improve the error message to give the user a clue that the problem may be that there is not a released asset that matches their request, and they can specify the version if they want.

### DIFF
--- a/packages/teraslice-cli/src/cmds/assets/deploy.ts
+++ b/packages/teraslice-cli/src/cmds/assets/deploy.ts
@@ -120,7 +120,11 @@ export = {
             try {
                 assetPath = await asset.download(cliConfig.assetDir, cliConfig.args.quiet);
             } catch (err) {
-                reply.fatal(`Unable to download ${cliConfig.args.asset}, you can download a prerelease asset by specifying its version.\n\nExample:\n\t${cliConfig.args.asset}@v1.2.3`);
+                if (!asset.version) {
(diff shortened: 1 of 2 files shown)


packages/teraslice-cli/src/cmds/assets/deploy.ts
```diff
@@ -120,7 +120,7 @@ export = {
             try {
                 assetPath = await asset.download(cliConfig.assetDir, cliConfig.args.quiet);
             } catch (err) {
-                reply.fatal(`Unable to download ${cliConfig.args.asset} asset: ${err.stack}`);
+                reply.fatal(`Unable to download ${cliConfig.args.asset}, you can download a prerelease asset by specifying its version.\n\nExample:\n\t${cliConfig.args.asset}@v1.2.3`);
```
peterdemartini, line 123: We should probably only log this error if they don't have a specific version specified. Although that is a pretty minor issue

godber: Good suggestion, looks like that is accessible on `asset.version` too.  I'll see if I can fool with other failure scenarios too.

godber: This will have to do for now.  Really to do anything more we will want to expose the [`getReleases`](https://github.com/terascope/fetch-github-release/blob/master/src/getReleases.js) from `fetch-github-release`:

https://github.com/terascope/fetch-github-release/blob/master/src/index.js

peterdemartini (approved):

PR #2914, opened 2021-08-05
Improve Geo Point Performance

peterdemartini:
- Remove cached check for geo points that uses `WeakSet` since doesn't scale with the V8 garbage collector
- Improve performance of `parseGeoPoint` by creating a "hot path" for our ideal geo point data structure


jsnoble (approved):
LGTM

PR #2917, opened 2021-08-06
Improve object column performance

peterdemartini:
- Removes the custom unique hash stuff that created way too much memory (this was the biggest performance again)
- Create building functions can be v8 inlined optimize
- No longer freeze objects stored in a column (we just need to be smart about not touching it)

Previously building a column with an object array took 4.451 seconds and used 688MiB of RAM. Now it takes 365ms and used 119 MiB of RAM.  So it uses over 5x less RAM previously required and is a 12x faster.

### DIFF
--- a/packages/data-mate/bench/build-column-perf-test.js
+++ b/packages/data-mate/bench/build-column-perf-test.js
@@ -1,0 +1,26 @@
+'use strict';
+
+const { config, data } = require('./fixtures/data.json');
+const { Column } = require('./src');
+
+const name = 'results';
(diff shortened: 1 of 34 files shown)


peterdemartini:
I am running all of the benchmarks again right now to compare to make sure there aren't any serious issues but I don't think it should stop this PR from going out because we need to see the benefit in production to really tell

kstaken (approved):
Moving forward for further validation.

jsnoble (approved):
LGTM

PR #2923, opened 2021-08-09
More DataFrame empty row performance improvements

peterdemartini:
- Allow setting null values to override existing ones
- Remove `BuilderWithCache` since that creates more overhead than it is worth
- Add `DataFrame->countEmptyRows`, `DataFrame->hasEmptyRows`, `DataFrame->hasNilValues`, `Vector->hasNilValues`, `Vector->countValues` and `Vector->isEmpty`
- Add the ability to reverse search the Vector data buckets so it will perform better when there are a lot of them
- Misc performance improvements


jsnoble (approved):
LGTM

PR #2924, opened 2021-08-09
stand up rabbitmq for integration tests in scripts

jsnoble:


packages/scripts/src/helpers/test-runner/services.ts
```diff
@@ -20,6 +20,11 @@ const logger = ts.debugLogger('ts-scripts:cmd:test');
 
 const serviceUpTimeout = ms('2m');
 
+const rabbitConfigPath = `${getRootDir()}/.ts-test-config/rabbitmq.conf`;
+
+// we create the rabbitmq.conf file for tests
+fs.outputFileSync(rabbitConfigPath, 'loopback_users = none\nloopback_users.guest = false');
```
peterdemartini, line 26: We should only do this if they request RabbitMQ and this should use an async request

packages/scripts/src/helpers/test-runner/services.ts
```diff
@@ -20,6 +20,11 @@ const logger = ts.debugLogger('ts-scripts:cmd:test');
 
 const serviceUpTimeout = ms('2m');
 
+const rabbitConfigPath = `${getRootDir()}/.ts-test-config/rabbitmq.conf`;
```
peterdemartini, line 23: We should use `path.join` here:

```suggestion
const rabbitConfigPath = path.join(getRootDir(), '.ts-test-config/rabbitmq.conf');
```

peterdemartini (changes requested):

peterdemartini (approved):

PR #2929, opened 2021-08-11
Bundle asset bundles keep names

peterdemartini:
Fix issue with asset bundle renaming functions, this broke ts-transforms


ciorg:
used this branch build standard assets, ran the same test job that was failing, and this fixed the issue

ciorg (approved):
Fixed the issue!

PR #2945, opened 2021-08-18
Improve aggregation/data frame performance

peterdemartini:
- Improve aggregation performance by doing less
- Improve column lookup by name performance
- Improve date performance (minor)
- Fix `BigMap` and `BigSet` API to be more compatible with `Set` and `Map`


jsnoble (approved):
LGTM

PR #2954, opened 2021-08-20
DataFrame improvements

peterdemartini:
- Improve data frame deserialize performance (not much)
- Use `getColumnOrThrow` in more places to utilize the cache
- Hide the internal storage for Vector data so we can change how it is being stored in the future

### DIFF
--- a/packages/data-mate/bench/src.js
+++ b/packages/data-mate/bench/src.js
@@ -1,4 +1,3 @@
 'use strict';
 
 module.exports = require('../dist/src');
-// module.exports = require(`${process.env.HOME}/tmp/data-mate/node_modules/@terascope/data-mate`);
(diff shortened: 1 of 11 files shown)


packages/data-mate/bench/src.js
```diff
@@ -1,3 +1,4 @@
 'use strict';
 
 module.exports = require('../dist/src');
+// module.exports = require(`${process.env.HOME}/tmp/data-mate/node_modules/@terascope/data-mate`);
```
jsnoble, line 4: does this need to be in here?

jsnoble (changes requested):

packages/data-mate/bench/src.js
peterdemartini: Nope

jsnoble (approved):
LGTM

PR #2955, opened 2021-08-20
Add ListVector.getRowIterator and improve iterator support in utils

peterdemartini:
- Add `ListVector.getRowIterator` so we can iterate over values in a given row
- Add `isIterator` and `isAsyncIterator` to utils
- Fix `isEmpty` thinking that the iterators are empty


jsnoble (approved):
LGTM

PR #2968, opened 2021-08-25
Faster getHashCodeFrom

peterdemartini:
Improve the performance of `getHashCodeFrom` by relying less on md5 and extra function calls. I think this also reduces the likelihood of collisions. We used `md5` when the string length was greater than 35 characters, I changed that to `1024`.

### DIFF
--- a/packages/utils/src/field-type.ts
+++ b/packages/utils/src/field-type.ts
@@ -9,8 +9,9 @@ import { isIPRangeOrThrow, isIPOrThrow } from './ip';
 import { toEpochMSOrThrow } from './dates';
 import { toBooleanOrThrow } from './booleans';
 import {
-    isValidateNumberType, toBigIntOrThrow, toFloatOrThrow,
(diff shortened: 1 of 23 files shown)


packages/utils/src/field-type.ts
```diff
@@ -89,30 +91,47 @@ function coerceToGeoPoint(input: unknown): GeoPoint {
...
+            ? _mapToString(value)
+            : value
+    );
+}
 
-    const hash = typeof value === 'object'
-        ? _mapToString(value)
-        : primitiveToString(value);
+/**
+ * If we has a hash a really long value we want to ensure that
```
jsnoble, line 121: Description here is a little wierd

jsnoble (changes requested):

packages/utils/src/field-type.ts
peterdemartini: I think I fixed that

jsnoble (approved):
LGTM

PR #2990, opened 2021-09-02
Fix index mapping and setPrecision

peterdemartini:
- Fix non-timeseries template index pattern to not include wildcard
- Allow NaN and Infinity with setPrecision


jsnoble (approved):
LGTM

PR #3011, opened 2021-09-13
Fix support for unindexed field types 

peterdemartini:
- Fix support for unindexed field types
- Allow overriding the key, updated and created dates in the index model


jsnoble (approved):
LGTM

PR #3032, opened 2021-09-21
Fix boundary support

peterdemartini:
We were incorrectly using the geo point parsing logic to process geo boundary data types, this fixes that.


packages/utils/test/type-coercion-spec.ts
```diff
@@ -54,4 +55,26 @@ describe('type-coercion', () => {
             });
         });
     });
+
+    describe('coerceToGeoBoundary', () => {
+        const validCases = [
+            [{ lat: 12, lon: 13 }, [{ lat: 12, lon: 13 }]],
+            [
+                [{ lat: 12, lon: 13 }, { lat: 13, lon: 14 }],
+                [{ lat: 12, lon: 13 }, { lat: 13, lon: 14 }]
```
jsnoble, line 64: I believe this is an invalid test. If I remember correctly, a geo boundary is just a geobox.  If that is so then there can only be two geopoint in the array, one representing the top left of the box, and the next is the bottom right. Any more than two and it becomes a geopolygon, any less its a geopoint. It needs to have exactly two

packages/data-mate/test/vector/vector-spec.ts
```diff
@@ -150,6 +150,17 @@ describe('Vector', () => {
                 null
             ],
         ],
+        [
+            FieldType.Boundary,
+            [[[90, 60], ['90.123', '60.456']], { lat: '89.002', lon: '20.034990' }, null, undefined],
+            undefined,
+            [
+                [{ lat: 60, lon: 90 }, { lat: 60.456, lon: 90.123 }],
+                [{ lat: 89.002, lon: 20.034990 }],
```
jsnoble, line 159: this sould throw, as listed for the reasons below

jsnoble (changes requested):

packages/utils/test/type-coercion-spec.ts
peterdemartini: Ahh that makes sense. Good catch, I updated, will you verify I did it correctly

jsnoble (approved):
LGTM

PR #3120, opened 2021-11-05
Change DataFrame toJSON to return DataEntities

peterdemartini:
Change DataFrame.toJSON and similar functions to output DataEntities for better interop between or other libraries


jsnoble (approved):
LGTM

PR #3131, opened 2021-11-10
Make `RouteSenderAPI->send` return the number of affected records

peterdemartini:
- Make `RouteSenderAPI->send` return the number of affected records
- Make `RouteSenderAPI->send` accept an iterable so we can stream process records in the future
- Add comments to the interface
- Move the interface to utils so it can be defined without needing to import job-components, this is useful for some our api packages

These changes should be backwards compatible since the number of affected records can default to the number of input records

### DIFF
--- a/packages/data-mate/package.json
+++ b/packages/data-mate/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@terascope/data-mate",
     "displayName": "Data-Mate",
-    "version": "0.36.2",
+    "version": "0.36.3",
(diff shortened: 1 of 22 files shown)


jsnoble (approved):
This is fine, though would it be more appropriate for it to live in the types package?

peterdemartini:
> This is fine, though would it be more appropriate for it to live in the types package?

I wanted to do that too but the DataEntity type doesn't live in the types package

PR #3142, opened 2021-11-15
Breaking changes to `elasticsearch-api->buldSend` to fix retry support

peterdemartini:
Makes breaking changes to the elasticsearch api `bulkSend` method so that the retries work correctly.

### DIFF
--- a/docs/packages/elasticsearch-api/overview.md
+++ b/docs/packages/elasticsearch-api/overview.md
@@ -240,24 +240,7 @@ const name = 'logs_template';
 elasticsearch.putTemplate(template, name)
 ```
 
-### bulkSend [DEPRECATED]
-
-Uses the client bulk functionality with exponential back-off retries. Use bulkSendImproved instead
-
-Query requires:
-
-- data  (formatted to work with elasticsearch bulk queries)
-
-```js
-const elasticsearch = require('@terascope/elasticsearch-api')(client, logger, opConfig);
-elasticsearch.bulkSend(data)
-  .then(function(){
-      //all done sending data
-   });
-```
-
-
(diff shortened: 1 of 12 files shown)


jsnoble (approved):
LGTM, though is there any reason why this could not be added directly to bulkSend? why does it have to be another method?

peterdemartini:
> LGTM, though is there any reason why this could not be added directly to bulkSend? why does it have to be another method?

I actually we could do that it mean we would have to do major update on that package

ciorg (approved):
LGTM

jsnoble:
> > LGTM, though is there any reason why this could not be added directly to bulkSend? why does it have to be another method?
> 
> Actually we could do that it mean we would have to do major update on that package

I would prefer that, better to have a well behaved method then a good one and a one that does not retry. It would simply things. I don't know if this needs a major as nothing is broken nor has the interface changed. It has an additional feature to retry, but it behaves exactly how it did before on a regular query

peterdemartini:
> > > LGTM, though is there any reason why this could not be added directly to bulkSend? why does it have to be another method?
> > 
> > 
> > Actually we could do that it mean we would have to do major update on that package
> 
> I would prefer that, better to have a well behaved method then a good one and a one that does not retry. It would simply things. I don't know if this needs a major as nothing is broken nor has the interface changed. It has an additional feature to retry, but it behaves exactly how it did before on a regular query

I am not sure if you read the code right because this is very much a breaking change. The input now takes `{ action: Metadata, data?: Record<string, unknown> }[]` vs the raw flat bulk request functions and the return value returns a number instead of that last success retry response.

PR #3146, opened 2021-11-16
Add chunk iterator util and fix affected rows

peterdemartini:
- Adds `chunkIter` util
- Fix affected row count by not counting 404 items in the result in `bulkSend`


jsnoble (approved):
LGTM

PR #3153, opened 2021-11-18
Expand bulk request API in IndexStore to support upserts with scripts

peterdemartini:
- [x] Expand the bulk request API in the `IndexStore` to support upserts with scripts
- [x] Fix the support reducing duplicates bulk requests by record key to avoid certain issues with bulk requests


jsnoble (approved):
LGTM

Issue #2202, opened 2020-09-17
Teraslice k8s workers should be able to have different cpu request and limit


godber:
Right now, the k8s deployment for the teraslice worker sets CPU request and limit to the same thing.  There should be a way that these can be set differently.  We are seeing some jobs that read from kafka who's workers lag more than others and eventually fall behind.  I suspect that if these workers could burst a little bit, they would probably catch up.

Initially I am thinking something like make `limit = request + limit_offest` where `limit_offset = 1` by default but can be overridden by job.

peterdemartini:
Wouldn't it be easier to just use the same k8s config options, like `limit` and `request`?

godber:
That would be most straight forward, sure.  I guess I could maintain backwards compatibility by doing:

* if `cpu` is a number, do what we do now `request` = `limit` = `value`
* if `cpu` is `{ request: requestValue, limit: limitValue}`, use those values as you would expect.

PR #3157, opened 2021-11-19
[k8s] add support for separate req/lim

godber:
This PR adds the ability for a teraslice job to specify separate requests and limits for the teraslice worker pods in Kubernetes.

Todo:

- [x] manually verify that this works e2e in k8s, share results here
- [x] extend the teraslice config schema for separate resource settings
- [x] cleanup or fix my commented out code
- [x] refactor the env var thing maybe

refs #2202


godber:
I think I've missed part of the job config code.  In manual testing I get this message when submitting the job

```
Validation failed for job config: example-data-generator-job-resource - configuration param 'cpu' missing from config, did you override its parent?
configuration param 'memory' missing from config, did you override its parent?
```

godber:
After my coffee, I remembered there are two schemas that need updating for jobs.

godber:
Even after updating the schema I still get some form of the error

```
configuration param 'memory' missing from config, did you override its parent?
```

It turns out this is because the parameters get flattened in two different ways as shown below:

```
flatInstance: {..."cpu":1,"memory":805306368}
```

versus

```
    flatInstance: {...,"cpu.requests":0.5,"cpu.limits":1,"memory.requests":805306368,"memory.limits":905306368}
```

peterdemartini:
That makes sense, can we just add a new property for the `cpu` and `memory` and throw if the old was are specified at the same time. Like what about:
```json
{
    "resources": {
        "cpu": {
             "requests": 0.5,
             "limits": 1
         },
        "memory": {
             "requests": 805306368,
             "limits": 905306368
        }
     }
}
```

godber:
There are two minor usability issues with

```
{
    "resources": {
        "cpu": {
             "requests": 0.5,
             "limits": 1
         },
        "memory": {
             "requests": 805306368,
             "limits": 905306368
        }
     }
}
```

That is k8s specifies it as `resources.requests.cpu`, and we're pluralizing something that isn't an array.  So in the similar vein to your suggestion, how about:

```
{
    "resources": {
        "requests": {
             "cpu": 0.5,
             "memory": 805306368
         },
        "limits": {
             "cpu": 1,
             "memory": 905306368
        }
     }
}
```

peterdemartini:
Sure I think that makes sense

godber:
@jsnoble this PR is "ready" if it weren't for the data-mate test that is failing.  Have you seen this before?

jsnoble (approved):
LGTM, would be nice to see the units in the description around memory, but I wont hold that up for it

godber:
Ok, assuming that the `packages/data-mate/src/function-configs/date/setTimezone.ts` functions are just used in tests, I think I've completed the changes, doc updates and bumped the right packages.  Pretty sure this is ready to do.

PR #3183, opened 2021-12-01
Performance improvement to DataFrames and add appendOne

peterdemartini:
- Minor performance improvements to DataFrame so it doesn't iterate over the columns too much when creating a DataFrame
- Add `DataFrame.appendOne` for a faster alternative to `appendAll` since it does less
- Add memory allocation optimizations to the Vector, ReadableData and WritableData which fixes memory explosions with lots of data buckets


jsnoble (approved):
LGTM

PR #3254, opened 2022-04-07
Async client creation

jsnoble:
- base foundation to create async elasticsearch client (only enough for spaces)
- updated packages


packages/elasticsearch-store/src/elasticsearch-client/create-client.ts
```diff
@@ -0,0 +1,85 @@
...
+
+        if (body.version.distribution === 'opensearch') {
+            // TODO: clean this up
+            const openConfig = {
+                ...config,
+                node: gotOptions.url as string
+            };
+            const client = new open.Client(openConfig);
+            const meta = {
+                client_type: 'opensearch',
```
godber, line 54: Change `client_type` to `distribution`.

godber (changes requested):
We reviewed this on zoom and have requested the changes listed here as well as changing the conditionals on version to version and `distribution`.

packages/teraslice/lib/storage/backends/elasticsearch_store.js
```diff
@@ -84,7 +84,7 @@ module.exports = function elasticsearchStorage(backendConfig) {
 
         if (fields) {
             const esVersion = elasticsearch.getESVersion();
-            if (esVersion > 6) {
+            if (esVersion !== 6) {
```
kstaken, line 87: Does this need to include a check on the `distribution` as well?

jsnoble: hmm, good spot you found, is should be able to use the helper method isElasticsearch6 for this

PR #3268, opened 2022-06-15
Scripts fix

jsnoble:
- renamed a directory that would accidently be removed during a post install inside assets (it would attempt to whittle down uneeded files to make the asset size smaller)


godber:
Would constraining the `doc` or `docs` in here be possible or a better fix?

https://github.com/terascope/elasticsearch-assets/blob/master/.yarnclean

godber:
> Would constraining the `doc` or `docs` in here be possible or a better fix?

I guess renaming that isn't that bad.

ciorg (approved):
LGTM

PR #3269, opened 2022-06-21
Allow scripts test suite to better run with asset directory structures

jsnoble:
- Improve compatibility usage on how scripts test-suites can be used with our asset packages, will be first used with elasticsearch-assets


ciorg (approved):
LGTM

PR #3271, opened 2022-07-13
Dataframe missing variable field fix

jsnoble:
- undefined variables will turn the logic node to return false as it is inoperable
- this allows dataframes and document matcher to behave more like elasticsearch, so it wont blow up or return wrong data


ciorg (approved):
LGTM

PR #3274, opened 2022-07-26
fix issue of test failing depending on the date proivided and the tim…

jsnoble:
…ezone offsets, added more tests


ciorg (approved):
Wrapping times in UTC time func seems like a good solution.

LGTM

PR #3278, opened 2022-08-10
Bigint interger bug fix

jsnoble:
- javascript numbers go up to ^51 while an integer in another system is ^31 which causes friction when a value is coerced into a number, this fix makes it more permissable


ciorg (approved):
LGTM

PR #3279, opened 2022-08-15
update script error handling and remove xpack commands from opensearch

ciorg:
* error handling and tests for pulling a bad docker image
* removed xpack commands on opensearch


jsnoble (approved):
LGTM

PR #3282, opened 2022-08-22
Teraslice store startup fix

jsnoble:
- fix's the issue around using the new client type on elasticsearch store startup. It should now be able to wait and recover if teraslice is up before elasticsearch is up.


ciorg (approved):
LTGM

PR #3300, opened 2022-09-14
Add minio and s3_exporter example job

briend:
- Adds a minio instance to a new docker-compose file in the examples/es-to-s3 folder
- Adds an example s3 job (`s3_exporter` op) to copy data from elasticsearch into minio w/ a query option
- Makes a correction for the `elasticsearch-next` teraslice configs

### DIFF
--- a/config/docker-teraslice-master.yml
+++ b/config/docker-teraslice-master.yml
@@ -9,11 +9,18 @@ terafoundation:
         elasticsearch-next:
             default:
                 node:
-                    - "http://localhost:9200"
+                    - "http://elasticsearch:9200"
         kafka:
             default:
                 brokers:
                     - "kafka:9092"
+        s3:
+          minio:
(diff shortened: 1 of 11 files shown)


godber:
Let's avoid making changes to any other existing config.  Instead, restructure this so everything one needs is in `examples/es-to-s3`.  So anything you're modifying, copy to the new location and revert.

godber:
The changes to `config/docker-teraslice-master.yml` and `config/docker-teraslice-worker.yml` are no longer necessary, right?  Those should be reverted.

briend:
I think localhost was a mistake, it wasn’t working for me so I left that
correction in this PR

https://github.com/terascope/teraslice/commit/9f133c7101eddbd738bb92d9327a83bbd5a302fb

godber (approved):
Looks good

PR #3309, opened 2022-10-19
WIP: Convert to esm modules

jsnoble:
- [ ]  check spots where require.main is used
- [ ] look into tests hacks that were done for teraslice tests
- [ ] look into tree shaking


jsnoble:
Notes so far:

- must have direct require suffix  `./src` => `./src/index.js`
- had to use `lodash-es` vs `lodash` for module support (its their native esm version that they support)
- cannot have deep require for modules `lodash\get` => `import { get } from 'lodash-es' , same with date-fns

jsnoble:
- Tree shaking only really works if we use a bundler like webpack or other similar tools. As noted [here](https://lodash.com/per-method-packages), individual packages will not be in lodash 5, and in some cases can make it worse. After looking at our code, we already import the full lodash library, so having additional side packages will actually make it worse. Proper imports will allow tree shaking if we use a tool for it, so we should not bloat it more
- If lodash is a real problem, which I don't think so, we can look into this [cli](https://lodash.com/custom-builds) to do that, but we would need to override downloading of this at install and use this instead.
- If we did use a tool, we should add a clause in package.json to indicate if its pure or not so it can determine if it can drop files when shaking
- Only advatage I have read is that the compiler when executing can ignore files if things are properly imported from esm modules

jsnoble:
- Typescript made a decision on how to parse the type files with models, makiing many libraries that have types to not work as is https://github.com/microsoft/TypeScript/issues/49160

PR #3310, opened 2022-11-15
fix to not override original includes/excludes

lesleydreyer:
Fix Query Access to respect the original included/excluded fields


jsnoble (approved):
LGTM

PR #3325, opened 2022-12-15
Teraslice cli assets init fix

ciorg:
* fixes`Callback support have been removed`  bug for teraslice-cli assets init and it's options
* bumped cli version


jsnoble (approved):
LGTM

PR #3329, opened 2023-01-04
Test roll bug2

godber:
This isn't meant to be merged, it's just here as a convenience for testing:

https://github.com/terascope/elasticsearch-assets/issues/970

### DIFF
--- a/config/docker-teraslice-master.yml
+++ b/config/docker-teraslice-master.yml
@@ -6,10 +6,22 @@ terafoundation:
             default:
                 host:
                     - "elasticsearch:9200"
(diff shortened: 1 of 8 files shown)


godber:
Actually, I think we should modify and merge this.  However, lets not overwrite any of the existing files.  Lets move all of these changes into a subdirectory like `examples/multiple-es-clusters` and adapt the files so they run in place there.  It will be useful to have a more complex/realistic example laying around.

godber:
I think we've got all of the good stuff out of this PR, I'm going to close it.

PR #3330, opened 2023-01-11
Fix bulk connection failure bug when cluster rolls 

jsnoble:
- Added check in elasticsearch-api handle errors better on when to retry


jsnoble:
Don't pull in till I double-check changes to elasticsearch-asset

ciorg (approved):
LGTM

PR #3334, opened 2023-01-27
Update facade client tests

jsnoble:
- added tests for the ten missing tests
- fix bad reference in client facade to underlining client
- fix type references


ciorg (approved):
LGTM

Issue #3270, opened 2022-06-21
Upgrade Teraslice to node 16 docker image


godber:
We need to upgrade Teraslice to the node 16 docker image.  If node 18 works, that's fine too.  But I think this is complicated by the asset build process.  The assets need to build for appropriate target node versions.

godber:
I finally got this to build 

https://app.travis-ci.com/github/terascope/base-docker-image/builds/260894118

We now have a node `v16.19.1-core` image:

https://hub.docker.com/layers/terascope/node-base/16.19.1-core/images/sha256-d9f26668e2ce785330ec9c071ffa3ee660d73ff67701d699bf79f2831ebc0a89?context=explore

I've gotten rid of the non-core images on the assumption that the teraslice images no longer need the kafka connector built into them.

godber:
The e2e test fails with a kafka error:

https://app.travis-ci.com/github/terascope/teraslice/jobs/596587644#L2645

```
{"name":"teraslice","assignment":"worker","hostname":"6515946ff32f","pid":369,"level":50,"err":{"message":"Could not find connector implementation for: kafka\n","name":"Error","stack":"Error: Could not find connector implementation for: kafka\n\n    at createConnection (/app/source/packages/terafoundation/dist/src/connector-utils.js:102:15)\n    at Object.getConnection (/app/source/packages/terafoundation/dist/src/api/index.js:88:75)\n    at KafkaReaderApi.create (/app/assets/****************************************/index.js:56506:49)\n    at Object.create (/app/assets/****************************************/index.js:51324:51)\n    at KafkaFetcher.initialize (/app/assets/****************************************/index.js:56292:32)\n    at processTicksAndRejections (internal/process/task_queues.js:95:5)\n    at async /app/source/packages/job-components/dist/src/execution-context/base.js:61:17\n    at async /app/source/node_modules/p-map/index.js:57:22"},"msg":"Teraslice Worker shutting down due to failure!","time":"2023-02-22T23:06:49.473Z","v":0}
```

stack trace:

```
Error: Could not find connector implementation for: kafka

    at createConnection (/app/source/packages/terafoundation/dist/src/connector-utils.js:102:15)
    at Object.getConnection (/app/source/packages/terafoundation/dist/src/api/index.js:88:75)
    at KafkaReaderApi.create (/app/assets/****************************************/index.js:56506:49)
    at Object.create (/app/assets/****************************************/index.js:51324:51)
[...]

PR #3338, opened 2023-02-22
release: (minor) teraslice@0.82.0

godber:


godber:
refs #3270

godber:
Ok, yeah, that was silly of me, we can't just do node 16 since we don't make node 16 assets:

`Could not find a release for terascope/elasticsearch-assets (linux x64)`

So I will dial this back to the latest node14.  Then we can ask the node16 question later.  The branch name will be a misnomer though.

godber:
If this passes tests, we need to validate that the kafka asset still works with this build since I've removed it from the base image.

godber:
This passed again after putting the kafka check back

https://app.travis-ci.com/github/terascope/teraslice/builds/260932765

jsnoble (approved):

PR #3346, opened 2023-03-10
WIP: updated es-api types

ciorg:
Idea is to return the non-retryable records back to the bulk_sender if the dead letter action is `kakfka_dead_letter`, for example records where a value is different than it's es index mapping.

Currently the bulkSender returns a count, this would change the returned value an object with the count and an array of deadLetter records if there are any.  So I need to track down what functions this would affect where they are expecting just a number to be returned.

The other part of this would be to change the es_bulk processor so that if deadLetter records are returned it could use  the `rejectRecord` function to send the records to the kafka dead letter topic.

If `_dead_letter_action` is not set to `kafka_dead_letter` in the opConfig it should behave like normal.

### DIFF
--- a/packages/elasticsearch-api/index.js
+++ b/packages/elasticsearch-api/index.js
@@ -278,6 +278,7 @@ module.exports = function elasticsearchApi(client, logger, _opConfig) {
     */
     function _filterRetryRecords(actionRecords, result) {
         const retry = [];
+        const deadLetter = [];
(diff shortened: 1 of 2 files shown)


jsnoble (approved):
LGTM

jsnoble (commented):
This needs to be a minor update, not a patch since the api has changed

PR #3348, opened 2023-03-17
K8s update mar 2023

godber:
I had this working with the example job with the `initContainers` added, that ran and everything so it worked.  I've made some changes since then so it's possible that I broke it again, but its close.  I need to add tests and implement the `kubernetes_overrides_enabled` config setting

refs: 3347

### DIFF
--- a/docs/configuration/clustering-k8s.md
+++ b/docs/configuration/clustering-k8s.md
@@ -95,7 +95,7 @@ The table below shows the Teraslice Master configuration settings added to
 support k8s based Teraslice deployments.
 
 |         Configuration          |                                                                        Description                                                                         |  Type   |  Notes   |
-| :----------------------------: | :--------------------------------------------------------------------------------------------------------------------------------------------------------: | :-----: | :------: |
+|:------------------------------:|:----------------------------------------------------------------------------------------------------------------------------------------------------------:|:-------:|:--------:|
 |         assets_volume          |                               Name of kubernetes volume to be shared across all pods, where Teraslice assets will be stored                                | String  | optional |
 |    cpu_execution_controller    |                                           CPU resources to use for Execution Controller request and limit values                                           | Number  | optional |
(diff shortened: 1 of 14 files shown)


godber:
I think that I should reject a job a job submission time if `kubernetes_overrides_enabled: false` and the job has a non null (`{}`) `pod_spec_override`.  Can I do that in `format` in `job-schemas.ts`:

https://github.com/terascope/teraslice/pull/3348/files#diff-501767f567d0fd519fa00553a9d316cef76c0675f54f2a3443ccda88951af053R314

Or is there  job validation elsewhere, I feel like there was other job validation.

docs/configuration/clustering-k8s.md
```diff
@@ -436,6 +437,39 @@ property should be the name of the Kubernetes `persistentVolumeClaim`.  The
...
+
+Teraslice provides the `pod_spec_override` job property which allows the user to
+specify any arbitrary Kubernetes configuration to be overlaid on the
+[pod](https://kubernetes.io/docs/concepts/workloads/pods/) `.spec` for both the
+execution controller and worker pods.  This allows users to "sneak" in
+functionality that we have not yet explicitly implemented.
+
+For instance, this feature can be used to add `initContainers` or `hostAliases`.
+A `pod_spec_override` must be in the form of JSON that maps directly to the
+appropriate YAML for the Kubernetes resource.  And example of a
```
ciorg, line 450: Typo - `An example`

ciorg (approved):
Approving, found one small typo

docs/configuration/clustering-k8s.md
godber, line 450:
```suggestion
appropriate YAML for the Kubernetes resource.  An example of a
```

godber: Thanks, I've got that fixed, I merge when it completes.

PR #3349, opened 2023-03-21
Dlq refactor

ciorg:
Reverted changes to apis and tests from the previous dead-letter-queue related pr and changed the dead letter queue action to update the docs metadata instead of returning the docs with non-retryable errors.  This allows the class apis and tests to remain the same while still supplying the dlq info for each doc sent to elasticsearch.


ciorg:
changed metadata field to `_bulk_sender_rejection`

PR #3351, opened 2023-03-27
fix yarn issue and bundled asset fixture for tests

jsnoble:


jsnoble:
This is not needed anymore

PR #3353, opened 2023-04-05
Fix broken tests

jsnoble:
- made changes to travis.ci to better work with scripts
- removed TestHarness auto loading of ./test/cache asset files to prevent race conditions, it now requires explicit asset paths to be passed in
- made modifications to scripts and how tests are run to make sure they are all run


packages/teraslice-test-harness/test/job-test-harness-spec.ts
```diff
@@ -10,7 +10,7 @@ import {
 import { JobTestHarness } from '../src';
 
 describe('JobTestHarness', () => {
-    const assetDir = path.join(__dirname, 'fixtures');
+    const assetDir = path.join(__filename, '../fixtures');
```
godber, line 13: This and all the rest of the changes from `__dirname` to `__filename` ... are they necessary?  They seem arbitrary.

godber (approved):
I just have the one question about `__dirname`

packages/teraslice-test-harness/test/job-test-harness-spec.ts
jsnoble: they are essentially the same, it was part of the debugging process as I originally would get different results depending on where the tests were called (from top of teraslice vs the package themselves). The error was a little vague but talked about errors that where not referenced in the folder being run. I thought somehow the pathing was getting wrong depending on how it was called. However, in the end  the main culprit was due to the assets auto loading I believe. I could go back and switch that if needed though it should essentially be the same value

godber: Yeah, I think it would be best if you changed those back.  The changes are needless, and just slightly harder to understand.  `path.join(__dirname, 'fixtures')` makes sense whereas `path.join(__filename, '../fixtures')` is confusing since it relies on weird behavior from `path.join()`.

PR #3355, opened 2023-04-20
xlucene mode to filter undefined variable nodes

lesleydreyer:
Adds the ability to filter out variable nodes if variables are undefined. Must pass "filterNilVariables" to the xlucene-parser as well as the variables object for it to work.


jsnoble (changes requested):
So I looked through the changes, I was thinking of this leaning more towards a more recursive approach, there is some, but it is a relying heavily on mapping and filtering which takes up a lot more memory and cpu. This is fine since the ast are never that big, but this is something to consider to measure in our join queries.  Its probably fine.

However the bigger issue is semver, it is all in patch. This is a new major feature, it should not be auto included like a patch will. Please change from patch to minor semver update

jsnoble (approved):

PR #3361, opened 2023-06-07
first steps to move tests to github actions

jsnoble:


.github/workflows/publish.yml
```diff
@@ -0,0 +1,40 @@
+name: Publish Assets
+run-name: Publish assets to npm
```
godber, line 2:
```suggestion
name: Publish Teraslice NPM Packages and Docker Image
```

Changing the name to get rid of assets and removing `run-name` entirely.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version:  18.16.0
+          registry-url: 'https://registry.npmjs.org'
+          cache: 'yarn'
```
godber, line 25: This is good, I have seen this `cache` setting work.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,40 @@
...
+          YARN_SETUP_ARGS: "--prod=false --silent"
+
+      - name: Build documentation
+        run: ./scripts/publish-documentation.sh
+
+      - name: Publish to npm
+        run: yarn ts-scripts publish -t tag npm
+
+      - name: Publish to docker
+        run: yarn ts-scripts publish -t tag docker
```
godber, line 39: I don't know if this will work, hopefully it has the correct credentials.

.github/workflows/test.yml
```diff
@@ -0,0 +1,254 @@
+name: Teraslice CI Tests
+run-name: Teraslice Tests
```
godber, line 2:
```suggestion
```

Let's leave `run-name` empty since it will copy useful context from the action if left blank.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#run-name

> If run-name is omitted or is only whitespace, then the run name is set to event-specific information for the workflow run. For example, for a workflow triggered by a push or pull_request event, it is set as the commit message.

packages/elasticsearch-api/package.json
```diff
@@ -14,11 +14,11 @@
...
         "test": "ts-scripts test . --",
-        "test:7": "ELASTICSEARCH_VERSION='7.9.3' yarn run test",
-        "test:8": "ELASTICSEARCH_VERSION='8.1.2' yarn run test",
+        "test:7": "ELASTICSEARCH_VERSION='7.9.3' yarn run test:debug",
+        "test:8": "ELASTICSEARCH_VERSION='8.1.2' yarn run test:debug",
         "test:debug": "ts-scripts test --debug . --",
-        "test:legacy": "LEGACY_CLIENT=true yarn run test",
-        "test:opensearch": "TEST_OPENSEARCH='true' RESTRAINED_OPENSEARCH='true' yarn run test",
+        "test:legacy": "LEGACY_CLIENT=true yarn run test:debug",
+        "test:opensearch": "TEST_OPENSEARCH='true' RESTRAINED_OPENSEARCH='true' yarn run test:debug",
```
godber, line 21: Do you really mean to leave these set to `:debug` mode?

I'm not objecting, just making sure this isn't something you've overlooked.

.github/workflows/publish.yml
godber: Thinking about this ... I don't know how the docker password and username, stored as secrets in github will get used by this publish docker script.  This PR doesn't do anything to provide those credentials at this point, so I suspect it will fail.

godber (changes requested):
I am pretty sure the docker publish will fail.

godber (approved):

Issue #3363, opened 2023-06-15
fix ts-scripts opensearch timeout


godber:
About 1 in 10 runs of the opensearch container by ts-scripts service runner results in a 2min timeout error like this:

```
✔  success   elasticsearch@6.8.6 is running at http://10.1.0.100:49200/, took 30.65sec

✖  fatal     TSError: Opensearch service (***10.1.0.100:49210) timeout after 2min
    at Timeout.<anonymous> (/home/runner/work/elasticsearch-assets/elasticsearch-assets/node_modules/@terascope/utils/dist/src/promises.js:166:24)
    at listOnTimeout (internal/timers.js:557:17)
    at processTimers (internal/timers.js:500:7)
Caused by: Opensearch service (***10.1.0.100:49210) timeout after 2min

error Command failed with exit code 1.
```

This is a retry-able error but a big nuisance.

godber:
This is reproducable locally and the status of things when it happens is as follows:

The container starts

```
docker ps -a
CONTAINER ID   IMAGE                                COMMAND                  CREATED          STATUS          PORTS                                                              NAMES
513876417efe   opensearchproject/opensearch:1.3.0   "./opensearch-docker…"   38 seconds ago   Up 37 seconds   9200/tcp, 9300/tcp, 9600/tcp, 9650/tcp, 0.0.0.0:49210->49210/tcp   ts_test_opensearch
1fd0e14e3882   blacktop/elasticsearch:6.8.6         "/elastic-entrypoint…"   38 seconds ago   Up 37 seconds   9200/tcp, 9300/tcp, 0.0.0.0:49200->49200/tcp                       ts_test_elasticsearch
```

there is no response on the http endpoint

```
curl http://admin:admin@192.168.1.198:49210
curl: (52) Empty reply from server
```

The logs on the docker container show a problem starting up:

```
docker logs -f 513876417efe
Disabling execution of install_demo_configuration.sh for OpenSearch Security Plugin
Disabling OpenSearch Security Plugin
Killing opensearch process 10
Killing performance analyzer process 11
Killing performance analyzer process 11
OpenSearch exited with code 143
Performance analyzer exited with code 143
```

godber:
Here's Peter seeing one of the issues I'm researching here:

https://github.com/jestjs/jest/issues/9659#issuecomment-623692945

Here's a link to an endless Jest bug on noise ... https://github.com/jestjs/jest/issues/13576

godber:
[...]

PR #3364, opened 2023-06-16
Minor fixes and possible opensearch fix

godber:
This branch tries to fix some minor deprecation noise and other things like that.  I don't think it really has much of a chance to fix the opensearch timeout bug.

### DIFF
--- a/packages/scripts/src/helpers/config.ts
+++ b/packages/scripts/src/helpers/config.ts
@@ -62,7 +62,7 @@ export const OPENSEARCH_HOSTNAME = process.env.OPENSEARCH_HOSTNAME || HOST_IP;
 export const OPENSEARCH_PORT = process.env.OPENSEARCH_PORT || '49210';
 export const OPENSEARCH_USER = process.env.OPENSEARCH_USER || 'admin';
 export const OPENSEARCH_PASSWORD = process.env.OPENSEARCH_PASSWORD || 'admin';
-export const OPENSEARCH_VERSION = process.env.OPENSEARCH_VERSION || '1.3.0';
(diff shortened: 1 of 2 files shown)


packages/scripts/src/helpers/config.ts
```diff
@@ -62,7 +62,7 @@ export const OPENSEARCH_HOSTNAME = process.env.OPENSEARCH_HOSTNAME || HOST_IP;
 export const OPENSEARCH_PORT = process.env.OPENSEARCH_PORT || '49210';
 export const OPENSEARCH_USER = process.env.OPENSEARCH_USER || 'admin';
 export const OPENSEARCH_PASSWORD = process.env.OPENSEARCH_PASSWORD || 'admin';
-export const OPENSEARCH_VERSION = process.env.OPENSEARCH_VERSION || '1.3.0';
+export const OPENSEARCH_VERSION = process.env.OPENSEARCH_VERSION || '1.3.6';
```
godber, line 65: The hope is that this version doesn't fail, the rationale for accepting this change even if it doesn't make a difference is that this version is used more.

godber:
refs #3363

godber:
Oh, I should probably fork this off Jared's branch.

packages/scripts/src/helpers/test-runner/services.ts
```diff
@@ -355,8 +355,7 @@ async function checkRestrainedOpensearch(
             const actual: string = body.version.number;
             const expected = options.opensearchVersion;
 
-            const satifies = semver.satisfies(actual, `^${expected}`);
```
godber, line 358: This variable was misspelled and also unnecessary.

jsnoble (approved):
LGTM

PR #3372, opened 2023-06-29
fix test helper method

jsnoble:
- populateIndex was broken and not tested, made sure that tests are using it and fixed bug


ciorg (approved):
LGTM

PR #3374, opened 2023-07-12
support to test opensearch v2 and elasticsearch7 for e2e tests

jsnoble:
- add support to test against opensearch v2 and elasticsearch v7 for e2e tests


godber (approved):
I like how this has simplified things Jared.

PR #3376, opened 2023-07-13
test runner resource fix

jsnoble:
- explicitly list test resources to be used for each test instead of spawning in unused instances


godber:
Closes #3379

godber (approved):
Looks great Jared!

PR #3383, opened 2023-07-26
Store schema chagnes

jsnoble:
- fix s3 schema in terafoundation
- change default kafka version for testing to 3.1

### DIFF
--- a/packages/terafoundation/src/connectors/s3.ts
+++ b/packages/terafoundation/src/connectors/s3.ts
@@ -50,12 +50,12 @@ export default {
                 default: '',
                 format: String
             },
-            s3ForcePathStyle: {
+            forcePathStyle: {
                 doc: '',
                 default: false,
                 format: Boolean
             },
-            s3BucketEndpoint: {
+            bucketEndpoint: {
                 doc: '',
                 default: false,
                 format: Boolean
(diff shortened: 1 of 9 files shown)


packages/terafoundation/src/connectors/s3.ts
```diff
@@ -50,12 +50,12 @@ export default {
...
                 format: String
             },
-            s3ForcePathStyle: {
+            forcePathStyle: {
                 doc: '',
                 default: false,
                 format: Boolean
             },
-            s3BucketEndpoint: {
+            bucketEndpoint: {
```
godber, line 58: What is the difference between `endpoint` and `bucketEndpoint`?

godber (commented):
The rest of this PR looks good, I'm just concerned that we're overlooking something WRT `endpoint` and `bucketEndpoint`.  How is each of those used?

packages/terafoundation/src/connectors/s3.ts
jsnoble: you can check the [migration](https://github.com/aws/aws-sdk-js-v3/blob/main/UPGRADING.md) here and see about forcePathStyle and see what bucketEndpoint does

godber (approved):
Approved, thanks Jared!

Issue #3402, opened 2023-09-06
add feature to --replace flag within terascope-CLI


sotojn:
I'm using the following teracope-cli command to build and deploy to my cluster.

``` $ teracscope-cli assets localhost --build --replace ```

My issue is that I expect the ``` --replace ``` flag to overwrite my old **.zip** file within the ``` build ``` folder in my asset directory.

I'm aware that if I use this command --> ``` $ teracscope-cli assets build --overwrite ``` that it correctly overwrites the file, but I feel that the ``` replace flag ``` should be expected to do the same and would be more practical in some instances.

Here is the current result when trying to use the ``` replace ``` flag.
```  
% earl assets deploy localhost --build --replace                                          
Beginning asset build.
Failure to build asset, caused by Error: Zipfile already exists "/Users/jsoto/Workspace/teraslice/joseph-asset-v-2/build/joseph-asset-v-2-v0.1.0-node-14-bundle.zip"  
```

Here is my expected output:
```
% earl2.0 assets deploy localhost --build --replace
Beginning asset build.
* copying files to the tmp directory "/var/folders/ht/_plsbr191vz0jl0br8cwypy80000gn/T/tmp-25310-RrnoUyx6zLNP"
[NOTE] Automatically added node_version, platform, arch restrictions for the asset Use --dev to temporarily disable this, or put false for the values node_version, platform, arch in the asset.json
* making tmp bundle directory "/var/folders/ht/_plsbr191vz0jl0br8cwypy80000gn/T/tmp-25310-Caa7x1NYZ0Li"
* running yarn --prod --no-progress
* entryPoint: /var/folders/ht/_plsbr191vz0jl0br8cwypy80000gn/T/tmp-25310-RrnoUyx6zLNP/asset/index.js
* doing a test require of /private/var/folders/ht/_plsbr191vz0jl0br8cwypy80000gn/T/tmp-25310-Caa7x1NYZ0Li/index.js
* overwriting /Users/jsoto/Workspace/teraslice/joseph-asset-v-2/build/joseph-asset-v-2-v0.1.0-node-14-bundle.zip
* zipping the asset bundle
Asset created:
	/Users/jsoto/Workspace/teraslice/joseph-asset-v-2/build/joseph-asset-v-2-v0.1.0-node-14-bundle.zip (158 kB)
*** Warning ***
The --replace option is intended for asset development only.
Using it for production asset management is a bad idea.
But Joseph thinks its fine.
[...]

PR #3403, opened 2023-09-07
add feature to terasclice-cli replace flag

sotojn:
I added an ```If``` statement on line 138 inside ```teraslice/packages/teraslice-cli/src/cmds/assets/deploy.ts```

```diff
File: deploy.ts
  130:         } else if (cliConfig.args.build || assetJsonExists) {
  131:             const asset = new AssetSrc(
  132:                 assetJsonExists ? '.' : cliConfig.args.srcDir,
  133:                 cliConfig.args.dev
  134:             );
  135: 
  136:             try {
  137:                 reply.green('Beginning asset build.');
+ 138:                 if (cliConfig.args.replace) {
+ 139:                     asset.overwrite = true;
+ 140:                 }
  141:                 const buildResult = await asset.build();
  142:                 assetPath = buildResult.name;
  143:                 reply.green(`Asset created:\n\t${buildResult.name} (${buildResult.bytes})`);
  144:             } catch (err) {
  145:                 reply.fatal(new TSError(err, {
  146:                     reason: 'Failure to build asset'
  147:                 }));
  148:                 return;
  149:             }
  150: 
```
I ran all the ```teraslice-cli``` tests and they are all passing.

```cmd
 PASS   teraslice-cli  packages/teraslice-cli/test/lib/jobs-spec.ts (6.593 s)
  jobs
    ✓ should return a job object (1 ms)
  jobs.workers should
    ✓ return correct response if ts-client response is an object (11 ms)
    ✓ return correct response if ts-client response is a string (5 ms)
  jobs.awaitStatus should
    ✓ return status when desired status is reached (5015 ms)

Test Suites: 38 passed, 38 total
Tests:       129 passed, 129 total
Snapshots:   0 total
Time:        6.747 s, estimated 7 s
```

This should solve the issue referenced here #3402

### DIFF
--- a/packages/teraslice-cli/package.json
+++ b/packages/teraslice-cli/package.json
@@ -1,7 +1,7 @@
 {
     "name": "teraslice-cli",
     "displayName": "Teraslice CLI",
-    "version": "0.55.0",
+    "version": "0.55.1",
     "description": "Command line manager for teraslice jobs, assets, and cluster references.",
     "keywords": [
         "teraslice"
--- a/packages/teraslice-cli/src/cmds/assets/deploy.ts
+++ b/packages/teraslice-cli/src/cmds/assets/deploy.ts
@@ -135,6 +135,9 @@ export = {
 
             try {
                 reply.green('Beginning asset build.');
+                if (cliConfig.args.replace) {
+                    asset.overwrite = true;
+                }
                 const buildResult = await asset.build();
                 assetPath = buildResult.name;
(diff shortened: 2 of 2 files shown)


godber (changes requested):
Please bump the minor version on this package.

Note: there are no tests that directly test this change so there's no concrete proof that your change works.  But the existing code coverage is weak and the code itself isn't well factored to enable testing.  It probably isn't worth resolving in this PR.

godber (approved):

PR #3408, opened 2023-09-13
add ts-scripts bump feature for assets

sotojn:
@busma13 and I added an asset bump feature to the ```ts-scripts``` bump command.

You can now run the bump command inside the ```root``` directory of an ```asset```. 

For example here is running ```yarn bump``` inside of the root directory of ```standard assets```

```cmd
User@standard-assets-root % yarn bump                          
ts-scripts.js bump [packages..]

Update a package to specific version and its dependencies. This should be run in the root of the workspace.

Positionals:
  packages  Run scripts for one or more package (if specifying more than one make sure they are ordered by dependants)
                                                                                                                   [array] [default: []]

Options:
      --prerelease-id  Specify the prerelease identifier, defaults to RC                                                 [default: "rc"]
      --skip-reset     Skip resetting the packages to latest from NPM                                         [boolean] [default: false]
  -d, --deps           Bump the child dependencies recursively, (ignores the monorepo's main package)          [boolean] [default: true]
      --skip-asset     If in an asset repository, bump the package version without updating the asset version [boolean] [default: false]
      --patch          Bump release by patch                                                       [boolean] [required] [default: false]
      --minor          Bump release by minor                                                       [boolean] [required] [default: false]
      --major          Bump release by major                                                       [boolean] [required] [default: false]
      --prerelease     Bump release by prerelease                                                  [boolean] [required] [default: false]
      --prepatch       Bump release by prepatch                                                    [boolean] [required] [default: false]
      --preminor       Bump release by preminor                                                    [boolean] [required] [default: false]
      --premajor       Bump release by premajor                                                    [boolean] [required] [default: false]
  -r, --release        Specify the release change for the version, see https://www.npmjs.com/package/semver.
                       [DEPRECATED] Use --patch, --minor, --major, etc
                                         [string] [choices: "patch", "minor", "major", "prerelease", "prepatch", "preminor", "premajor"]
  -h, --help           Show help                                                                                               [boolean]
  -v, --version        Show version number                                                                                     [boolean]

Examples:
  ts-scripts.js bump  example // 0.20.0 => 0.20.1
  ts-scripts.js bump  --patch example // 0.20.0 => 0.20.1
  ts-scripts.js bump  --minor example // 0.5.0 => 0.6.0
  ts-scripts.js bump  --prepatch example // 0.20.0 => 0.20.1-rc.0
  ts-scripts.js bump  --premajor example // 0.15.0 => 1.0.0-rc.0
[...]

PR #3409, opened 2023-09-14
Bump asset feature dev fix

sotojn:
I made some updates based off of our final pr with the main branch here #3408

The main updates are: 

- Refactored code by removing ```console.logs``` and unnessesary comments
- Updating tests inside [bump-spec.ts](https://github.com/terascope/teraslice/blob/bump-asset-feature-dev-fix/packages/scripts/test/bump-spec.ts) to use temp files while testing "Bump Assets" instead of creating files in the root directory.
- Fixed a bug that added "asset:false" inside of the root ```package.json``` inside of teraslice causing "verify build" to fail on GitHub actions.


busma13 (approved):

busma13:
LGTM

PR #3414, opened 2023-09-20
Date tuple fixes

jsnoble:
- change behaviour around date tuples, the time in milliseconds will not be offset until presentation time. Most date functions will now be timezone aware and not automatically convert to UTC. If given a tuple with a non-zero offset, it will be considered as a local time and show as such
- moved TZ=utc from proliferating in package.json. Moved it to service.js and to test-runner so that all tests and teraslice will have correct settings locally so it properly reflects what the server is set to.


ciorg (approved):
I read through everything and looked for functions that needed the `getValidDateWithTimezoneOrThrow` function but may have been overlooked and didn't find any and also double checked the tests.

LGTM

PR #3426, opened 2023-10-04
Update the clustering-k8s readme and example asset 

busma13:
- Add an updated example asset to `examples/k8s`
- Remove original asset that no longer had a valid format
- Update all example jobs to use the new asset
- Update the instructions in `clustering-k8s.md`


docs/configuration/clustering-k8s.md
```diff
@@ -501,8 +501,86 @@ There is a `Makefile` I use to help bootstrap Teraslice and do repetitive tasks,
...
+```bash
+minikube -n ts-dev1 service teraslice-master --url
+```
+Return to the original terminal.
+
+```bash
+export ES_URL=http://path/to/elasticsearch/tunnel # copy the URL returned in terminal 2
+export TERASLICE_MASTER_URL=http://path/to/teraslice-master/tunnel # copy the URL returned in terminal 3
+
+earl aliases remove ts-minikube-dev1
```
godber, line 536: This step and the one following it, basically redo something that's done above in `setup-all`.  We should probably ... I'm not sure ... just pull out the `earl alias add ...` from `setup-all` and then remove this `earl aliases remove ...`  from the instructions.  The dynamic nature of the URLs have forced this setup into a two part process.  I wonder if there are more options on `minikube service ...` like one that lets us CHOOSE a stable port number for each service we setup a tunnel for.  That would allow us to go back to doing everything in the setup.

godber: We've decided to just leave this as is for now.  We'll either come back to this and reconsider it or replace everything with KIND.

examples/k8s/example-asset/asset/example/processor.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+const { BatchProcessor } = require('@terascope/job-components');
+
+class Example extends BatchProcessor {
+    onBatch(dataArray) {
+        // example code, processor code goes here
+        dataArray.forEach((doc) => {
+            doc.type = this.opConfig.type;
+        });
```
godber, line 10: Did you actually run the job with this asset and it worked?  The original asset simply passed through the data array like `return (data) => data;`.  This processor will be setting `doc.type = string` on each doc unless the job is modified to add a `type: 'foo'` option to the processor.  This is fine as long as you know it runs.

godber (approved):

godber (approved):

examples/k8s/example-asset/asset/example/processor.js
busma13: I ran example-job.json, but didn't take a close look at what ended up in Elasticsearch. Let me run it again real quick.

busma13: It works properly if Minikube is running k8s v1.23.17.

packages/terafoundation/test/validate-configs-spec.ts
```diff
@@ -144,7 +144,7 @@ describe('Validate Configs', () => {
                             default: {
                                 accessKeyId: null,
                                 certLocation: '',
-                                endpoint: '127.0.0.1:80',
+                                endpoint: 'http://127.0.0.1:80',
```
busma13, line 147:
```suggestion
                                endpoint: '127.0.0.1:80',
```

godber (approved):

PR #3438, opened 2023-10-18
add mergeBy aggregation

lesleydreyer:
- [x] add mergeBy aggregation to data-mate aggregation frame

MergeBy merges unique values from the specified fields together into a single record... if there are multiple values for the same field, the merged record will be the last value in the list of records.

Decided to not allow a frame to call groupBy and mergeBy in the same aggregation for now because would need to know whether to allow other aggregations (sum, count, select, etc) on both or just one and currently you can call those before or after groupBy so we would need to rework

### DIFF
--- a/e2e/package.json
+++ b/e2e/package.json
@@ -37,7 +37,7 @@
     },
     "devDependencies": {
         "bunyan": "^1.8.15",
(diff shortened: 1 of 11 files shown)


jsnoble (approved):
LGTM

PR #3443, opened 2023-10-20
re-enable node 18 tests, remove node 14 tests

jsnoble:
This PR is the start of the migration away from node 14 and to set node 18 as the primary node version of teralice. Node 18 tests failed due to the got library having a bug in streaming redirects (used in asset processes) so we updated the library to use the ESM version which forces the cut of node 14

### DIFF
--- a/.github/workflows/publish-tag.yml
+++ b/.github/workflows/publish-tag.yml
@@ -42,7 +42,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [14.21.3, 16.19.1, 18.16.0]
+        node-version: [16.19.1, 18.16.0]
     steps:
       - name: Check out code
         uses: actions/checkout@v3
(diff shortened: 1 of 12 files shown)


godber:
@jsnoble can you bump the minor version on teraslice in this PR?

godber (changes requested):
This looks great Jared.  I think there are two more places we want to change the node version though:

* In the Dockerfile (https://github.com/terascope/teraslice/blob/3300119015bf3b214c47a487d303c7f35134206c/Dockerfile#L3)
* In the ts-scripts helper (https://github.com/terascope/teraslice/blob/3300119015bf3b214c47a487d303c7f35134206c/packages/scripts/src/helpers/config.ts#L160)

Unless you have specific reasons not to change those.  The first one certainly needs changing.  I haven't thought that hard about the one in ts-scripts.

godber (approved):

PR #3451, opened 2023-11-02
add restriction for regexp queries when prevent_prefix_wildcard

lesleydreyer:
- [x] Add restriction for prevent_prefix_wildcard on regexp queries ensuring they aren't prefixed with a regexp operator and also have a guaranteed match (i.e. ab* has to match a but a* doesn't so could search the whole index


jsnoble (approved):
LGTM

PR #3453, opened 2023-11-02
Fix ts scripts opensearch timeout

sotojn:
I have followed the instructions and guidance of issue #3363 to hopefully fix the opensearch timeout problem. 

- Updated the default opensearch version to be  `v1.3.10`
> > This is to have access to an environment variable that can turn off the performance analyzer.

- Set DISABLE_PERFORMANCE_ANALYZER_AGENT_CLI to 'true' for opensearch container
> > This will ideally stop opensearch from failing to startup randomly 

References to this solution [here](https://github.com/terascope/teraslice/issues/3363#issuecomment-1646317566)


busma13 (approved):
Looks good Joseph.

PR #3476, opened 2023-11-27
Convert to k8s js client from kubectl in k8senv

busma13:
This Pr includes the following changes:
- adds the @kubernetes/client-node package to ts-scripts
- replace the use of kubectl with the k8s node client when starting teraslice with the k8s-env command

Future improvements:
- replace kubectl command when starting services
- use the k8s node client in k8s e2e tests


packages/scripts/src/helpers/k8s-env/k8s.ts
```diff
@@ -0,0 +1,226 @@
...
+        try {
+            /// Creates configmap for teraslice-master
+            const masterConfigMap = baseConfigMap;
+            const masterTerafoundation: object = k8sClient.loadYaml(fs.readFileSync(`${path.join(e2eK8sDir, 'masterConfig', 'teraslice.yaml')}`, 'utf8'));
+            masterConfigMap.data = { 'teraslice.yaml': k8sClient.dumpYaml(masterTerafoundation) };
+            masterConfigMap.metadata = { name: 'teraslice-master' };
+            const response = await this.k8sCoreV1Api
+                .createNamespacedConfigMap(this.terasliceNamespace, masterConfigMap);
+            logger.debug('deployK8sTeraslice masterConfigMap:', response.body);
+        } catch (err) {
```
godber, line 90: This catch catches two possible error prone things, the `loadYaml` and the call to the k8s API `createNamespacedConfigMap`.   Wrapping each of these in their own try/catch would be ideal, so the error message can be explicit about exactly what failed.  That being said, since none of this is user supplied, the likelihood of an error is somewhat lower.

PR #3478, opened 2023-11-29
Typescript updates

jsnoble:
- convert teraslice and its unit/integration tests to typescript
- fix few bugs and refactored various modules
- added more types to out types package


godber:
We should probably adjust our Docker image creation to exclude `/app/source/packages/teraslice/src/` though I have been thinking of other ways to build the release image altogether ... e.g. `pkg` or something.

packages/teraslice/src/lib/workers/execution-controller/index.ts
```diff
@@ -631,56 +671,57 @@ class ExecutionController {
...
+        const metaData = this.executionStorage.executionMetaData(executionStats);
         this.logger.info(`execution ${this.exId} has completed`);
-        await exStore.setStatus(this.exId, 'completed', metaData);
+        await this.executionStorage.setStatus(this.exId, 'completed', metaData);
     }
 
     _logFinishedJob() {
         const endTime = Date.now();
-        const elapsed = endTime - this.startTime;
+        const elapsed = endTime - (this.startTime ?? 0);
```
godber, line 724: I doubt this is right, when would a default startTime of `0` be correct?  That would result in `elapsed` being like 50 years.  If `startTime` is undefined, the log should probably say unknown time or something.

godber (approved):

PR #3480, opened 2023-12-01
Add the --rebuild option to the k8s-env command

busma13:
This PR includes the following changes:
- Add the `--rebuild` option to `ts-scripts k8s-env`
- Use `yarn k8s:rebuild` script from root of teraslice
- This allows you to build a new Teraslice docker image and restart Teraslice in your k8s cluster. Previously you had to delete the cluster and re-run `yarn k8s`. Now services and the cluster do not need to be rebuilt.
- Changes to how teraslice and elasticsearch ports are determined. 
- Elasticserach store data is preserved between rebuilds. We may need to add the option to reset the ES store.

### DIFF
--- a/e2e/test/global.setup.js
+++ b/e2e/test/global.setup.js
@@ -13,10 +13,8 @@ const signale = require('./signale');
 const setupTerasliceConfig = require('./setup-config');
 const downloadAssets = require('./download-assets');
 const {
-    CONFIG_PATH, ASSETS_PATH, TEST_PLATFORM
+    CONFIG_PATH, ASSETS_PATH, TEST_PLATFORM, TERASLICE_PORT
 } = require('./config');
-
-const TERASLICE_PORT = 45678;
 
 module.exports = async () => {
     const teraslice = new TerasliceHarness();
--- a/e2e/test/global.setup.js
+++ b/e2e/test/global.setup.js
@@ -16,6 +16,8 @@ const {
     CONFIG_PATH, ASSETS_PATH, TEST_PLATFORM
 } = require('./config');
 
+const TERASLICE_PORT = 45678;
+
 module.exports = async () => {
     const teraslice = new TerasliceHarness();
     await teraslice.init();
@@ -39,9 +41,9 @@ module.exports = async () => {
     ]);
 
     if (TEST_PLATFORM === 'kubernetes') {
-        await deployK8sTeraslice();
+        await deployK8sTeraslice(TERASLICE_PORT);
         await teraslice.waitForTeraslice();
-        await setAliasAndBaseAssets();
+        await setAliasAndBaseAssets(TERASLICE_PORT);
     } else {
         await Promise.all([setupTerasliceConfig(), downloadAssets()]);
         await dockerUp();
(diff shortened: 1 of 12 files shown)


godber:
> Elasticserach store data is preserved between rebuilds. We may need to add the option to reset the ES store.

Yeah, there will be cases for both.

e2e/test/teraslice-harness.js
```diff
@@ -22,12 +22,15 @@ const { cleanupIndex } = ElasticsearchTestHelpers;
 
 const generateOnly = GENERATE_ONLY ? parseInt(GENERATE_ONLY, 10) : null;
 
+const TERASLICE_PORT = 45678;
```
godber, line 25: I'm not sure if this is accessible in this scope, but here's another place this port is defined:

https://github.com/terascope/teraslice/blob/7d9c89308d463832dd3470411b09c9caa3b7c93c/packages/scripts/src/helpers/scripts.ts#L24

In the scripts package, so maybe it does't make sense to have this dependency, or maybe it should go the other way.

e2e/test/global.setup.js
```diff
@@ -16,6 +16,8 @@ const {
     CONFIG_PATH, ASSETS_PATH, TEST_PLATFORM
 } = require('./config');
 
+const TERASLICE_PORT = 45678;
```
godber, line 19: You have this variable set elsewhere too.

godber: Is is a problem if it gets updated in one place but not the other?  If so, you have a problem.

packages/scripts/src/helpers/k8s-env/k8s.ts
```diff
@@ -188,39 +168,86 @@ export class K8s {
...
         } catch (err) {
-            logger.error('Error creating priorityClass: ', err);
+            logger.debug('Teraslice namespace cannot be deleted. It might not yet exist: ', err.response.body);
         }
+
+        await this.confirmDeletion(terasliceNamespace, this.k8sCoreV1Api);
     }
 
-    async deleteTerasliceNamespace() {
+    async confirmDeletion(terasliceNamespace: string, coreV1Api: k8sClient.CoreV1Api) {
```
godber, line 217: This is a pretty vague function name, if it means `confirmNamespaceDeletetion` you might name it that.

packages/scripts/src/helpers/scripts.ts
```diff
@@ -21,7 +21,6 @@ import { getE2eK8sDir } from '../helpers/packages';
 import { yamlDeploymentResource, yamlServiceResource } from './k8s-env/interfaces';
 
 const logger = debugLogger('ts-scripts:cmd');
-let TS_PORT = '45678';
```
godber, line 24: Now I'm confused about this TS_PORT thing.

godber (approved):
Approved, most of my comments are either general advice or stream of conscious comments.

e2e/test/teraslice-harness.js
busma13: - The teraslice port was a magic number in `e2e/test/teraslice-harness.js` and `e2e/test/global.setup.js`, so I just made it a constant so things were more explicit. I should be able to pass the value from one to the other.
- In my first attempt to declare the teraslice port, I placed it in `helpers/scripts.ts`, then modified it if we were in k8s-env mode, not testing mode. I then removed the declaration in favor of passing the port in from either the `global.setup.js` constant or the `k8s-env` command options.
- Ports are also hard coded into some yaml files. It would be nice to set a variable once and pass it everywhere needed!

godber (approved):

Issue #2670, opened 2021-04-28
k8s - Add force parameter to Teraslice Job stop API that force deletes things


godber:
It would be nice if k8s pods could be forcibly deleted as an option using the Teraslice job stop API.

Currently when stopping jobs it's like

```
POST /job/JOBID/_stop
```

perhaps an option like this:

```
POST /job/JOBID/_stop?force
```

I'd have to take a look at the k8s library I am using and see how I am using it and if force parameters can be added to these API calls, but the `kubectl` equivalant would be:

```
kubectl delete pods PODNAME --grace-period=0 --force
```

I'm not sure offhand how that works when deleting k8s jobs and deployments.

busma13:
I've spent some time playing with the `POST /job/JOBID/_stop` api endpoint as well as using `kubectl` with different combinations of `force` and `grace-period`.  Here are my thoughts so far:

- `POST /job/JOBID/_stop` is equivalent to `kubectl delete jobs`, not `kubectl delete pods`.  No combination of `force` and `grace-period` on a job will send a `SIGKILL` to worker and execution pods.  The worker pods will shutdown gracefully or timeout after `sysconfig.teraslice.shutdown_timeout` milliseconds. In prod this time is set to 5 minutes.
-  Calling `kubectl delete pods --grace-period=0` will send a `SIGKILL` to the pods specified, killing them instantly.  But without shutting down the jobs or/and resources first, the pods may restart automatically.
- Adding a `force` option to the  `POST /job/JOBID/_stop` endpoint would require adding a second request to the `jobs` section of the `k8s` client `delete` function that would delete all pods associated with the job.  This feels like the wrong approach, as we are fighting against the design principles of kubernetes. There may be possible race conditions where new worker pods are created to replace the killed pods if jobs/deployments are not yet deleted.

There are multiple scenarios that we are trying to fix:
1. A job is in a terminal state and is not allowed to be stopped. We don't need to shut pods down immediately. This should coincide with the `force` option?
2. A job is not in a terminal state, but we want the worker pods to shutdown immediately, without waiting `sysconfig.teraslice.shutdown_timeout` milliseconds. This should coincide with the `grace-period` option?
3. A combination of 1 and 2: terminal state that we want to shut down immediately.

busma13:
I missed something earlier:
- When using `force` and `grace-period` together grace period is ignored.
- This command will achieve scenario 2:
[...]

PR #3491, opened 2023-12-08
Add force option to POST /jobs/jobId/_stop and POST /ex/exId/_stop endpoints

busma13:
This PR adds a `force` option to the `POST /jobs/{jobId}/_stop` and `POST /ex/{exId}/_stop` endpoints
- include `force='true'` as a request parameter to forcefully stop all jobs, deployments, replicaSets and pods labeled with the execution ID in the url or the execution ID associated with the jobID in the url.
- This overrides the default behavior of not allowing a job with a status of `'completed', 'stopped', 'rejected', 'failed', 'terminated', or 'stopping'` to be stopped.
- Endpoint documentation updated.
- Adds a test for deleting with force set to true.
- Modifies `delete` job, service, and deployment tests to expect the complete response, not the response.body.

Ref: #2670

### DIFF
--- a/packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetes/k8s.ts
+++ b/packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetes/k8s.ts
@@ -301,20 +301,20 @@ export class K8s {
      *                                    'deployments', 'services', 'jobs'
      * @param  {Object}  forcePodList     List of all related pod, deployment, and job resources
      *                                    to be forcefully stopped.
-     * @return {Array}                    Array of k8s delete response objects.
+     * @return {Object}                   k8s delete response.
      */
     async delete(name: string, objType: string, forcePodList?: any) {
-        const responses = [];
+        let response;
 
         try {
             if (objType === 'services') {
-                responses.push(await pRetry(() => this.client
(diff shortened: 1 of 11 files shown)


packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetes/k8s.ts
```diff
@@ -315,16 +317,36 @@ export class K8s {
...
+                        kind: 'DeleteOptions',
+                        propagationPolicy: 'Background'
+                    }
+                };
+
+                const deletePodResponses = [];
+                if (forcePodList) {
+                    deleteOptions.body.gracePeriodSeconds = 1;
+
+                    for (const pod of forcePodList.items) {
```
busma13, line 335: I am reconsidering the choice to delete all of these pods and the job at once.  I could add a delete pod option to this function and loop through the pods names in the `_deleteObjByExId` function, calling `delete` on each of them. This would clean up this function and make the testing much easier.

godber (approved):

godber:
I'm going to merge this, no need to add the comments, perhaps in a later PR.  Note, I've bumped the Teraslice version to `v0.89.0` so I can make a release.

Issue #3489, opened 2023-12-08
Fix documentation for Teraslice


godber:
Merging the typescript PR broke our documentation CI.

I have temporarily disabled the CI in this PR:

https://github.com/terascope/teraslice/pull/3488

We should make the docs work ... preferably without snowballing into an upgrade of docusaurus.

sotojn:
When I run `yarn docs` in the root of teraslice I am getting a type error. This only seems to occur when building docs for @terascope/utils. It correctly builds docs for every package before it. 

```
$ tsc --build

★  star      Building docs for @terascope/eslint-config

★  star      Building docs for @terascope/types
…  awaiting  building typedocs for package @terascope/types
Loaded plugin /Users/jsoto/Workspace/teraslice/node_modules/typedoc-plugin-markdown
warning You are running with an unsupported TypeScript version! If TypeDoc crashes, this is why. TypeDoc supports 4.6, 4.7, 4.8, 4.9, 5.0
Documentation generated at /Users/jsoto/Workspace/teraslice/docs/packages/types/api
✔  success   generated docs for package @terascope/types
$ tsc --build

★  star      Building docs for @terascope/utils
…  awaiting  building typedocs for package @terascope/utils
Loaded plugin /Users/jsoto/Workspace/teraslice/node_modules/typedoc-plugin-markdown
warning You are running with an unsupported TypeScript version! If TypeDoc crashes, this is why. TypeDoc supports 4.6, 4.7, 4.8, 4.9, 5.0
ts-scripts docs [packages..]

Update documentation for all packages or a specific package

Positionals:
  packages  Run scripts for one or more packages                                                                                 [array] [default: []]

Options:
  -h, --help     Show help                                                                                                                   [boolean]
  -v, --version  Show version number                                                                                                         [boolean]

TypeError: Cannot read properties of undefined (reading 'name')
    at /Users/jsoto/Workspace/teraslice/node_modules/typedoc/dist/lib/converter/types.js:559:100
    at Array.map (<anonymous>)
    at Object.convertType (/Users/jsoto/Workspace/teraslice/node_modules/typedoc/dist/lib/converter/types.js:559:34)
    at convertType (/Users/jsoto/Workspace/teraslice/node_modules/typedoc/dist/lib/converter/types.js:105:34)
    at /Users/jsoto/Workspace/teraslice/node_modules/typedoc/dist/lib/converter/types.js:427:53
    at Array.map (<anonymous>)
[...]

PR #3495, opened 2023-12-12
 Update typedoc to be compatible with current typescript version.

busma13:
Typedoc v0.23.28 was not compatible with typescript v5.2.2.
The following packages were upgraded:
- typedoc from v0.23.28 to v0.25.4
- typedoc-plugin-markdown from 3.14.0 to 3.17.1

The `typedoc.Application` class now requires a call to `bootstrapWithPlugins()` to initialize.

Ref: #3489

### DIFF
--- a/.github/workflows/publish-master.yml
+++ b/.github/workflows/publish-master.yml
@@ -44,39 +44,39 @@ jobs:
       - name: Publish to docker
         run: yarn ts-scripts publish -t dev docker
 
-  # build-docs:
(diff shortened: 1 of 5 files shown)


busma13:
I manually checked the functionality by deleting docs out of the teraslice-cli package then running `yarn docs teraslice-cli`.  All files were regenerated. I spot checked individual files and those I checked have the same content.

godber (approved):

PR #3496, opened 2023-12-12
copy over types from other packages so we can start the migration over the future

jsnoble:
- set the foundation so we can start pulling from a single source of truth, this will be an ongoing process
- Namespace a few of the important types to help organize and prevent possible collisions of type name

`
import { Teraslice, Foundation } from '@terascope/types';

Foundation.Config
Teraslice.SliceRequest
`


godber (approved):

godber:
Should we be bumping Teraslices version so we can get a release with all of these related changes?

Issue #3482, opened 2023-12-06
K8s - parameterize k8s version in kind (e2e tests and k8s-env command)


busma13:
It would be helpful and probably necessary to be able to dynamically change which version of Kubernetes we use in the k8s e2e tests and when running a k8s environment for development.

- When kind creates a Kubernetes cluster, it will use a default node image determined by the current kind version, unless configured to do otherwise. Something like this: `kindest/node:v1.27.3@sha256:3966ac761ae0136263ffdb6cfd4db23ef8a83cba8a463690e98317add2c9ba72`
To see the default image for a given release see the release notes here: https://github.com/kubernetes-sigs/kind/releases

- To use a specific Kubernetes version you must add an image property to each node in your kind configuration yaml file. https://kind.sigs.k8s.io/docs/user/configuration/#kubernetes-version
  Example:
    ```kind: Cluster
    apiVersion: kind.x-k8s.io/v1alpha4
    nodes:
    - role: control-plane
     image: kindest/node:v1.16.4@sha256:b91a2c2317a000f3a783489dfb755064177dbc3a0b2f4147d50f04825d016f55
    - role: worker
      image: kindest/node:v1.16.4@sha256:b91a2c2317a000f3a783489dfb755064177dbc3a0b2f4147d50f04825d016f55
    ```

PR #3500, opened 2023-12-14
Add K8s version parameter to kind cluster creation

busma13:
- This PR allows for the setting of a `K8S_VERSION` env variable when running `yarn test:k8s` or `yarn k8s` (#3482)
- The env variable is converted to a command option which is used by the `Kind` class to create a kind cluster using that k8s version. 
- A map of `kind` versions to available images lives in the `Kind` class and will need to be updated with new kind versions.
- K8s images are limited to one minor version per kind version, with the patch level equal to the latest version at release time. For example:
  - kind v0.20.0 --> kindest/node:v1.26.6
  - kind v0.19.0 --> kindest/node:v1.26.4
  - kind v0.19.0 --> kindest/node:v1.26.3
- Create Kind class
- add `clusterName` option to k8s

### DIFF
--- a/packages/scripts/src/helpers/kind.ts
+++ b/packages/scripts/src/helpers/kind.ts
@@ -1,0 +1,75 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import execa from 'execa';
+import yaml from 'js-yaml';
+import { Logger, debugLogger } from '@terascope/utils';
+import signale from './signale';
+import { getE2eK8sDir } from '../helpers/packages';
+import { kindCluster } from './interfaces';
+
+export class Kind {
+    clusterName: string;
+    logger: Logger;
+
+    constructor(clusterName = 'default') {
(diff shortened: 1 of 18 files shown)


packages/scripts/src/helpers/kind.ts
```diff
@@ -0,0 +1,136 @@
...
+            const subprocess = await execa.command('kind version');
+            const version = subprocess.stdout.split(' ')[1].slice(1);
+            return version;
+        } catch (err) {
+            throw new Error('Kind version could not be determined.');
+        }
+    }
+}
+
+const map = {
```
godber, line 95: I would avoid every using a name as generic as `map`.  You don't need to change this now, but next time your come across this I'd give it a more descriptive name.

godber (approved):

godber:
I want to merge this, please bump the version and let me know after you do and I will merge it.

Issue #3498, opened 2023-12-13
Add ability to override `log_level` at job level


godber:
The logging level of the teraslice master and all other components is currently controlled in the master configuration using the `log_level`:

https://terascope.github.io/teraslice/docs/configuration/overview#terafoundation-configuration-reference

It should be possible to specify the `log_level` (with the same set of options) on the Teraslice job and have that setting override the global setting.

godber:
This capability should be in https://github.com/terascope/teraslice/releases/tag/v0.90.0

PR #3505, opened 2023-12-18
Add log_level parameter to teraslice job.

busma13:
This PR adds the `log_level` parameter to a teraslice job. 
- Export `loglevels` from `utils/src/Logger`
- Add `log_level` to `job-components/src/job-schema`
- Within the `Worker` and `ExecutionController` constructors, use the log level from the `executionContext` to set the log level of `context.logger` before it is used to create a new child logger.
- This child logger will be inherited by all other classes that have a logger within `Worker` and `ExecutionController` instances.

Ref: #3498


packages/job-components/src/job-schemas.ts
```diff
@@ -208,6 +209,19 @@ export function jobSchema(context: Context): convict.Schema<any> {
                     }
                 });
             },
+        },
+        log_level: {
+            default: 'info',
```
godber, line 214: I think this needs to be `undefined` or something, since your conditional in `worker-service.js` tests for existence, it will override the terafoundation config in all cases.

godber (changes requested):
Fix the default problem, and is there a way we could have implemented a test to catch that scenario.

godber (approved):

packages/job-components/src/job-schemas.ts
```diff
@@ -208,6 +209,19 @@ export function jobSchema(context: Context): convict.Schema<any> {
...
                 });
             },
+        },
+        log_level: {
+            default: undefined,
+            doc: 'the log level to be set on all loggers associated with the job',
+            format(level: unknown) {
+                if (typeof level !== 'string') {
+                    throw new Error('must be of type string');
+                }
```
godber, line 219: Given the check against `logLevels` I think this test for string might be redundant and less useful.  Imaging a user provides a number `5`, then they will be told to use a `string`.  Then they put in the string `noisy`, then it finally tells them the acceptible set of strings.  Maybe the next check should test for string and includes(level).

godber (changes requested):
Bump sub package versions and teraslice version to `0.90.0` (both top level and subpackage) and consider the inline suggestion

godber:
There's not a whole lot of code here, but since it cross cuts packages and schemas and things we might ask for @jsnoble's input.

packages/job-components/src/job-schemas.ts
busma13: Good point. I will combine the two checks and just have one error message.

busma13:
I added an explanatory comment on each of the constructors that overwrites the log level. I also confirmed that the log_level indicated in the job is written to a log file (all log streams are overwritten with new log level).

godber (approved):

godber (approved):

PR #3506, opened 2023-12-19
add constant for index required field types

lesleydreyer:
- [x] Add a constant for index required types so packages using es mapping can validate
- [x] Rework the toESMapping methods to use the above constant
- [x] bit of tree shaking


jsnoble (approved):
LGTM

PR #3508, opened 2023-12-20
Check for orphaned processes by job_id when starting a job

busma13:
- Add check for orphaned resources by job_id before starting job.
- Add listResourcesForJobId function to execution service.
- Add listResourcesForJobId function to k8s and native cluster backends.

- Improve messaging to user when using force option on the POST /jobs/{job_id}/_stop endpoint
- Remove check for native clustering when using force stop.
- Update docs for _stop endpoint to show an example using force

### DIFF
--- a/docs/management-apis/endpoints-json.md
+++ b/docs/management-apis/endpoints-json.md
@@ -372,7 +372,8 @@ Remove orphaned pods from a failed job:
 ```sh
 $ curl -XPOST 'localhost:5678/v1/jobs/5a50580c-4a50-48d9-80f8-ac70a00f3dbd/_stop?force=true'
 {
-    Force stop complete for exId 041a00a9-a474-4355-96aa-03e5ecf9b246, job status: failed
+    "message": "Force stop complete for exId: 041a00a9-a474-4355-96aa-03e5ecf9b246",
+    "status": "failed"
 }
 ```
 
@@ -762,7 +763,8 @@ Remove orphaned pods from a failed job:
 ```sh
 $ curl -XPOST 'localhost:5678/v1/ex/863678b3-daf3-4ea9-8cb0-88b846cd7e57/_stop?force=true'
 {
-    Force stop complete for exId 863678b3-daf3-4ea9-8cb0-88b846cd7e57, job status: failed
+    "message": "Force stop complete for exId: 863678b3-daf3-4ea9-8cb0-88b846cd7e57",
+    "status": "failed"
 }
 ```
 
--- a/docs/management-apis/endpoints-json.md
+++ b/docs/management-apis/endpoints-json.md
@@ -368,6 +368,14 @@ $ curl -XPOST 'localhost:5678/v1/jobs/5a50580c-4a50-48d9-80f8-ac70a00f3dbd/_stop
 }
 ```
 
+Remove orphaned pods from a failed job:
+```sh
+$ curl -XPOST 'localhost:5678/v1/jobs/5a50580c-4a50-48d9-80f8-ac70a00f3dbd/_stop?force=true'
+{
+    Force stop complete for exId 041a00a9-a474-4355-96aa-03e5ecf9b246, job status: failed
+}
+```
+
 ## POST /v1/jobs/{jobId}/_pause
 
 Issues a pause command, this will prevent the execution controller from invoking slicers and also prevent the allocation of slices to workers, marks the job execution context state as paused.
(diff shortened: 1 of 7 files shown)


docs/management-apis/endpoints-json.md
```diff
@@ -368,6 +368,14 @@ $ curl -XPOST 'localhost:5678/v1/jobs/5a50580c-4a50-48d9-80f8-ac70a00f3dbd/_stop
 }
 ```
 
+Remove orphaned pods from a failed job:
+```sh
+$ curl -XPOST 'localhost:5678/v1/jobs/5a50580c-4a50-48d9-80f8-ac70a00f3dbd/_stop?force=true'
+{
+    Force stop complete for exId 041a00a9-a474-4355-96aa-03e5ecf9b246, job status: failed
+}
```
godber, line 376: The response should be valid `json`.

godber: Please respond to this with what the stop command returns with (without stop).  We probably need to think about what the key will be.

busma13:
```
{
    "status": "completed"
}
```

busma13: Maybe something like this:
```
{
    "msg": "Force stop complete for exId 041a00a9-a474-4355-96aa-03e5ecf9b246",
    "status": "completed"
}
```

packages/teraslice/src/lib/cluster/services/jobs.ts
```diff
@@ -138,6 +138,12 @@ export class JobsService {
             });
         }
 
+        const clusterType = this.context.sysconfig.teraslice.cluster_manager_type;
+        const currentResources = await this.executionService.listResourcesForJobId(jobId);
+        if (currentResources.length > 0 && clusterType === 'kubernetes') {
```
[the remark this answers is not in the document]
godber: I don't really care, but if it does the job, lets use it.  Otherwise I'd rather leave that method stubbed out in `native`.

docs/management-apis/endpoints-json.md
godber: Yeah, that's along the right line, my question is, does precedence for a message like this already exist in the code, if so, what is it?  Lets try and be consistent.

godber (changes requested):
Lets sort out the response and the native method thing before accepting.  Otherwise this looks great.

docs/management-apis/endpoints-json.md
busma13: The JSON APIs return JSON objects or arrays of objects. It looks like `message` is used as a property name on error responses.
```
curl -XPOST localhost:5678/v1/jobs/52861e62-d0b7-46fd-8904-b94422011208/_workers\?add=2
{
    "error": 404,
    "message": "no active execution context was found for ex_id: 620e1385-7944-4986-86e1-662cf44d77ff"
}
```

packages/teraslice/src/lib/cluster/services/jobs.ts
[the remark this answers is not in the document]
busma13: I was tentative to use it in native because I don't fully understand how jobs are stopped in native. I don't see any issues with my implementation, but I would like @jsnoble to look it over.

packages/teraslice/src/lib/cluster/services/jobs.ts
```diff
@@ -138,6 +138,23 @@ export class JobsService {
...
+            for (const resource of currentResources) {
+                exIdsSet.add(resource.metadata.labels['teraslice.terascope.io/exId']);
+            }
+            const exIdsArr = Array.from(exIdsSet);
+            const exIdsString = exIdsArr.join(', ');
+            throw new TSError(`There are orphaned resources for job: ${jobId}, exId: ${exIdsString}.
+            To remove orphaned resources:
+                curl -XPOST <teraslice host>/v1/jobs/${jobId}/_stop?force=true
+                Or:
+                curl -XPOST <teraslice host>/v1/ex/${exIdsArr[0]}/_stop?force=true`);
```
godber, line 155: Please remove this option, we don't ever want to encourage people to interact with the `ex` directly, at this point I'd argue for removal of the APIs that enable doing anything other than `GET` on `ex` because it is just going to confuse things.

godber, line 155:
```suggestion
```
this ...

godber (approved):

Issue #3494, opened 2023-12-11
New asset generator unit test failures


busma13:
This is the second time I've seen these two unit tests fail.

```
 FAIL  packages/teraslice-cli/test/generators/new-asset-spec.ts
  ● new asset generator should › should create the correct asset dir tree and put files in the correct dir

    assert(received)

    Expected value to be equal to:
      true
    Received:
      false
    
    Message:
      /home/runner/work/teraslice/teraslice/packages/teraslice-cli/test/fixtures/generate-new-asset/generated-asset/new_asset/package.json, no such file or directory

      30 |
      31 |     it('should create the correct asset dir tree and put files in the correct dir', () => {
    > 32 |         assert.file([
         |                ^
      33 |             path.join(rootAssetPath, 'package.json'),
      34 |             path.join(rootAssetPath, '.eslintrc'),
      35 |             path.join(rootAssetPath, '.editorconfig'),

      at node_modules/yeoman-assert/index.js:56:12
          at Array.forEach (<anonymous>)
      at Function.Object.<anonymous>.assert.file (node_modules/yeoman-assert/index.js:54:26)
      at Object.<anonymous> (packages/teraslice-cli/test/generators/new-asset-spec.ts:32:16)

  ● new asset generator should › should modify the file content with the name and description

    ENOENT: no such file or directory, open '/home/runner/work/teraslice/teraslice/packages/teraslice-cli/test/fixtures/generate-new-asset/generated-asset/new_asset/package.json'

      43 |
      44 |     it('should modify the file content with the name and description', () => {
    > 45 |         assert.jsonFileContent(path.join(rootAssetPath, 'package.json'), {
         |                ^
      46 |             name: 'new_asset-bundle',
      47 |             description: 'this is a new asset',
      48 |             version: '0.1.0'

      at readFile (node_modules/yeoman-assert/index.js:32:19)
      at Function.Object.<anonymous>.assert.JSONFileContent.assert.jsonFileContent (node_modules/yeoman-assert/index.js:263:24)
      at Object.<anonymous> (packages/teraslice-cli/test/generators/new-asset-spec.ts:45:16)
```

godber:
I just saw this too
https://github.com/terascope/teraslice/actions/runs/7197386821/job/19604401446?pr=3497#logs

busma13:
It looks like the errors are related to the timing of directories being created and destroyed.  Creating temp directories with random names may fix this.

busma13:
```
[...]

PR #3509, opened 2023-12-20
Change teraslice-cli tests to use temp dir

sotojn:
- Changed tests inside `teraslice-cli` to use temp directories instead of sharing directories
  - This separates state between tests and guarantee directories even when tests are ran in parallel 

This fixes an issue posted here #3494

### DIFF
--- a/packages/teraslice-cli/package.json
+++ b/packages/teraslice-cli/package.json
@@ -1,7 +1,7 @@
 {
     "name": "teraslice-cli",
     "displayName": "Teraslice CLI",
-    "version": "0.59.1",
+    "version": "0.59.2",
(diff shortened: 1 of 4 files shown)


busma13 (changes requested):
The new-processor-spec also has directories that are shared. I'm thinking we should take care of that now, before it becomes a problem.

godber (dismissed):

sotojn:
> The new-processor-spec also has directories that are shared. I'm thinking we should take care of that now, before it becomes a problem.

I have updated `new-processor-spec.ts` to also use temp directories. Great feedback!

godber (approved):

busma13 (approved):

PR #3510, opened 2023-12-21
change docker image to not add "nodevX" if a custom image was provided

lesleydreyer:
- [x] change docker image to not add "nodevX" if a custom image was provided (DEV_DOCKER_IMAGE)


godber (approved):
I think this will work in all the use cases.

PR #3511, opened 2023-12-22
teraslice esm migration

jsnoble:
- migrate teraslice to use esm modules natively

Issues:
- eslint needs to be updated to be able to correctly parse typescripte >5.2. It throws warning
- eslint is having a problem linting an esm module based package. The primary issue is that this is a monorepo with several packages and we are slowly converting code over. For now I am ignoring linting on teraslice until I can figure out a way to do it independently for each package, assuming its worth the dev effort.
- using `@swc/jest` for typescript compiling for jest testing is causing errors with class decorators, so I am keeping ts-jest for the rest of the tests and using the new one until it can be properly configured


e2e/jest.config.js
```diff
@@ -8,4 +8,30 @@ const config = require('../jest.config.base')(__dirname);
 config.testPathIgnorePatterns = process.env.TEST_PLATFORM === 'kubernetes' ? ['data/recovery-spec', 'cluster/worker-allocation-spec', 'cluster/state-spec'] : [];
 config.collectCoverage = false;
 delete config.transform;
+// config.moduleNameMapper = {
+//     '^(\\.{1,2}/.*)\\.js$': '$1',
+// };
```
godber, line 13:
```suggestion
```

Is this comment still useful?  Or just left over?

e2e/jest.config.js
```diff
@@ -8,4 +8,30 @@ const config = require('../jest.config.base')(__dirname);
...
+//         },
+//         target: 'es2022'
+//     },
+//     module: {
+//         type: 'commonjs',
+//         strictMode: false,
+//         noInterop: false
+//     }
+// }];
+
```
godber, line 36:
```suggestion
```

Is this comment still useful or just left over?

packages/teraslice/jest.config.js
```diff
@@ -1,3 +1,21 @@
...
 
-module.exports = require('../../jest.config.base')(__dirname);
+const dirPath = fileURLToPath(new URL('.', import.meta.url));
+const configModulePath = path.join(dirPath, '../../jest.config.base.js');
+
+const module = await import(configModulePath);
+
+const config = module.default(dirPath);
+
+config.preset = "ts-jest/presets/default-esm";
```
godber, line 11: Didn't we get rid of ts-jest, is this config still necessary?

packages/teraslice/src/lib/config/default-sysconfig.ts
```diff
@@ -1,14 +1,3 @@
-// TODO: see if I can fix this
-function hasKafkaConnector() {
```
godber, line 2: Do you know why this was originally added?

packages/teraslice/src/lib/config/index.ts
```diff
@@ -1,10 +1,12 @@
 import path from 'node:path';
 import { get } from '@terascope/utils';
+import { fileURLToPath } from 'node:url';
 import { formats, TerasliceConfig } from '@terascope/job-components';
 import type { FoundationConfig } from 'terafoundation';
-import { configSchema } from './schemas/system';
+import { configSchema } from './schemas/system.js';
 
-export const terasliceOpPath = path.join(__dirname, '..');
+const filePath = fileURLToPath(new URL('.', import.meta.url));
```
godber, line 8: What is going on here?

packages/teraslice/src/lib/storage/backends/elasticsearch_store.ts
```diff
@@ -8,13 +8,13 @@ import {
...
-import executionSchema from './mappings/ex.json';
-import jobsSchema from './mappings/job.json';
-import stateSchema from './mappings/state.json';
+import { makeLogger } from '../../workers/helpers/terafoundation.js';
+import { timeseriesIndex } from '../../utils/date_utils.js';
+import analyticsSchema from './mappings/analytics.json' assert { type: 'json' };
+import assetSchema from './mappings/asset.json' assert { type: 'json' };
+import executionSchema from './mappings/ex.json' assert { type: 'json' };
+import jobsSchema from './mappings/job.json' assert { type: 'json' };
+import stateSchema from './mappings/state.json' assert { type: 'json' };
```
godber, line 17: Hmmm, ok, magic to import a JSON file?

packages/teraslice/src/lib/workers/metrics/index.ts
```diff
@@ -30,22 +30,25 @@ export class Metrics extends EventEmitter {
...
         this._typesCollectedAt = {};
+    }
 
+    async initialize() {
         // never cause an unwanted error
         try {
-            this.gcStats = require('gc-stats')();
+            // @ts-expect-error
+            const stats = await import('gc-stats');
+            console.log('stats', stats);
```
godber, line 40:
```suggestion

```

Do we still want this `console.log`?

packages/teraslice/test/workers/worker/slice-spec.ts
```diff
@@ -114,7 +114,7 @@ describe('Slice', () => {
                 expect(results).toEqual(times(10, () => ({ hi: true })));
 
                 // should have have the analytics data
-                expect(slice).not.toHaveProperty('analyticsData');
+                expect(slice.analyticsData).toBeUndefined();
```
godber, line 117: Is `expect(slice.analyticsData).toBeUndefined();` really the equivalent of `expect(slice).not.toHaveProperty('analyticsData')`

godber (commented):
I have some questions here.

e2e/jest.config.js
jsnoble: This may or may not be needed when I convert e2e to modules and typescript, was initially going to leave it for documentation for what needs to be done but I can clean it up

e2e/jest.config.js
jsnoble: same as above

packages/teraslice/src/lib/config/default-sysconfig.ts
jsnoble: to conditionally add default configuration if the library is being used, I just made it so its always there

packages/teraslice/src/lib/storage/backends/elasticsearch_store.ts
jsnoble: so `require` used to automatically convert json files into their object based form, import does not do that automatically. I would have to `readFile` the json file then `JSON.parse()` it, this is some sugar to tell it to do that for us

packages/teraslice/src/lib/workers/metrics/index.ts
jsnoble: this is a good catch, looking into it

packages/teraslice/test/workers/worker/slice-spec.ts
jsnoble: essentially yes, this change happened because jest was throwing when testing with esm modules that it did not have those methods available. Jest is really weird with esm and has to jump through a lot of hoops and I think that the version of jest we are using might have a bug with that. There is no global jest object in esm module form so there is some issues about that.

e2e/jest.config.js
godber: You can leave it if it's a clue for the future.

godber (approved):

PR #3515, opened 2024-01-03
add nodeSuffix flag to scripts publish

sotojn:
- Added a `node-suffix` flag in `ts-scripts publish` to allow the ability to include or exclude a node version suffix in a docker image tag 
  - By default it's set to `true` to maintain original functionality

### DIFF
--- a/packages/scripts/package.json
+++ b/packages/scripts/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@terascope/scripts",
     "displayName": "Scripts",
-    "version": "0.64.2",
+    "version": "0.65.0",
     "description": "A collection of terascope monorepo scripts",
     "homepage": "https://github.com/terascope/teraslice/tree/master/packages/scripts#readme",
     "bugs": {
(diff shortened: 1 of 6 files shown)


godber (changes requested):
Please add a yargs `.example()` for this option and bump the minor on the scripts package.

godber:
The `getDevDockerImage()` has a suffix too and should probably honor your new flag no?

https://github.com/terascope/teraslice/blob/5f65294f40d13a76cb0b25bf14d79bf5eb1a1dbd/packages/scripts/src/helpers/misc.ts#L131-L137

Consider this and then I want to talk to you about it, I think I want to push on, merge this PR as is.  Then you can later deal with `getDevDockerImage` and refactor a bit.  I think it's a sign that we can do some rethinking here, but it wasn't clear to me what to do.

busma13 (approved):

godber:
We are going to ignore the `getDevDockerImage` thing for now.  If anyone ends up wanting this we can implement it.

godber (approved):

Issue #3501, opened 2023-12-14
Add force option to teraslice-cli jobs and tjm stop commands


jeffmontagna:
Add a force option to teraslice-cli to the `jobs` and `tjm` stop commands.
### Documentation
https://terascope.github.io/teraslice/docs/management-apis/endpoints-json#post-v1jobsjobid_stop

### Commands
- `teraslice-cli jobs stop`
- `teraslice-cli tjm stop`
### Options
- `-f` and  `--force`
### Example using curl
```bash
curl -sS -XPOST 'localhost:80/v1/jobs/<jobid>/_stop?force=true' | jq 
{
  "status": "terminated"
}
```

godber:
We've decided against implementing this, we will add a separate command in the future to accommodate it, not only that, we intend to re-implement this functionality as a separate Teraslice HTTP API endpoint as described here https://github.com/terascope/teraslice/issues/3519.  So instead of `/JOBID/_stop?force` it will be a completely different action on a job.

PR #3517, opened 2024-01-05
Add force stop option to jobs stop and tjm stop cli commands

busma13:
- Add the `force` option to `Teraslice-cli tjm stop`
- Add the `force` option to `Teraslice-cli jobs stop`
  - Add a check to `jobs stop` that disallows the use of `force` with the `all` positional
- Convert query parameters on the `_stop` endpoint to the proper types
- Allow end execution status to be in a terminal state besides stopped when force stopping an execution
  - Allows for force stopping a job in the failed state
  - Allows for a running job to be forced stopped, resulting in a terminated state
  - These changes lead me to believe that creating a new endpoint for force stopping might be preferable

Fixes: #3501

### DIFF
--- a/packages/teraslice-cli/src/cmds/jobs/stop.ts
+++ b/packages/teraslice-cli/src/cmds/jobs/stop.ts
@@ -28,6 +28,9 @@ export = {
         return yargs;
     },
     async handler(argv: any) {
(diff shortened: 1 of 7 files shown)


godber:
Closing, reasoning here: https://github.com/terascope/teraslice/issues/3501#issuecomment-1884014467

Issue #3518, opened 2024-01-09
Use a Multi stage Dockerfile to eliminate the need for our own base image


godber:
I think we should at least try to rejigger our Teraslice `Dockerfile` to try and eliminate the need for maintaining our own base image.  I think the approach would be to use a Multistage Dockerfile (https://docs.docker.com/build/building/multi-stage/) whose first stage creates a temporary "dev image" (not pushed anywhere) that has all the `node-gyp` C dependencies and builds Teraslice.  Then the second stage will build a smaller final image from the output of the dev image.

Roughly speaking:

* The image should be `FROM` one of these: https://hub.docker.com/_/node
  * Note we can start with specific versions for now, e.g. https://hub.docker.com/_/node/tags?page=1&name=18.18.2
  * I would start with `bookworm`, then `bullseye`, then `alpine`
  * The dev image should be `bookworm`, and the final image should be `bookworm-slim` (adjust debian release name as appropriate)
* The first stage should do "mostly this stuff":
  * https://github.com/terascope/base-docker-image/blob/master/Dockerfile
* The second stage should do "mostly this stuff":
 * https://github.com/terascope/teraslice/blob/master/Dockerfile

I think we can still parameterize the Node version the same way we are now.  Our main goal here is to isolate the changes as much as possible in the `Dockerfile` and making sure what we build passes `e2e` tests.

Ref:

When considering naming schemes for our Docker images in the issue below, I thought its probably worth trying out:

https://github.com/terascope/debian-base/issues/1#issuecomment-1883904854

godber:
We attempted to roll out this change in the PR above in release:

https://github.com/terascope/teraslice/releases/tag/v0.91.0

But the resulting image showed significantly higher memory usage in large scale Kafka jobs so we rolled it back in this release:

https://github.com/terascope/teraslice/releases/tag/v0.92.0

For now this effort is on hold.

PR #3520, opened 2024-01-10
Create a multistage docker image from a debian-bookworm base

busma13:
- Rewrite docker file:
  - Discontinue use of `base-docker-image`
  - Create a multi-stage build:
    - base stage:
      - build image from `debian-bookworm`
      - combine functionality from the `base-docker-image` with the yarn build step to create `/app` directory
    - second stage:
      - build image from `debian-bookworm-slim`
      - copy `/app` artifact from base stage, minimizing dependencies
- Copy the following scripts from `base-docker-image` repo to `teraslice/scripts`:
  - `docker-pkg-fix.sh` (this doesn't seem to be used anywhere)
  - `wait-for-it.sh`

ref: #3518

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,46 @@
 # NODE_VERSION is set by default in the config.ts, the following value will only
 # be used if you build images by default with docker build
 ARG NODE_VERSION=18.18.2
-FROM terascope/node-base:${NODE_VERSION}
+FROM node:${NODE_VERSION}-bookworm as base
+
+ENV NPM_CONFIG_LOGLEVEL error
+ENV WITH_SASL 0
+
+RUN node --version
+RUN yarn --version
+RUN npm --version
+
+RUN mkdir -p /app/source
+
+# Install bunyan
+RUN yarn global add \
+    --ignore-optional \
+    --no-progress \
+    --no-emoji \
+    --no-cache \
+    bunyan
+
+# Install any built-in connectors in /app/
+# use npm because there isn't a package.json
+WORKDIR /app
+
+RUN npm init --yes &> /dev/null \
+    && npm install \
+    --build \
+    --no-package-lock \
+    --no-optional \
+    'terafoundation_kafka_connector@~0.11.1' \
+    && npm cache clean --force
+
+WORKDIR /app/source
+
+# verify node-rdkafka is installed right
+RUN node --print --eval "require('node-rdkafka')"
+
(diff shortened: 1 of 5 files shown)


godber:
That seemed too easy, what did you do wrong ;)?

Dockerfile
```diff
@@ -1,7 +1,46 @@
...
+
+WORKDIR /app/source
+
+# verify node-rdkafka is installed right
+RUN node --print --eval "require('node-rdkafka')"
+
+COPY scripts/docker-pkg-fix.js /usr/local/bin/docker-pkg-fix
+COPY scripts/wait-for-it.sh /usr/local/bin/wait-for-it
+
+ENV NODE_OPTIONS "--max-old-space-size=2048"
```
busma13, line 43: @godber This option needs to be in the final docker image, right? This looks to be related to garbage collection. It was in the `base-docker-image`, but with our current structure it will not be inherited.

godber: Yes

Dockerfile
```diff
@@ -24,19 +59,38 @@ RUN yarn --prod=false --frozen-lockfile \
...
 VOLUME /app/config /app/logs /app/assets
 ENV TERAFOUNDATION_CONFIG /app/config/teraslice.yaml
 
+# Use tini to handle sigterm and zombie processes
+ENTRYPOINT ["/usr/bin/tini", "--"]
+
 CMD ["node", "service.js"]
+
+RUN apt-get update && \
+     apt-get install -y libcurl4 tini
```
godber, line 83:
```suggestion
     apt-get install -y libcurl4 tini && \
       apt-get autoremove -y && \
       apt-get clean -y && \
       rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
```

You forget to do the cleanup after the `apt update/install`.  So you're leaving at least 20MB in the image:

```
du -sh /var/lib/apt/lists/ /tmp/ /var/tmp/
19M	/var/lib/apt/lists/
2.2M	/tmp/
4.0K	/var/tmp/
```

scripts/docker-pkg-fix.js
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env node
+
+'use strict';
+
```
godber, line 4:
```suggestion

// We are unsure what this does and it can probably be deleted, looks like something left
// from the migration from Node 10 to Node 12

```

godber (changes requested):
See inline requests

godber (approved):

PR #3524, opened 2024-01-12
Fix Docker tag logic to prevent undefined in dev tag

busma13:
- Pass in the node version to `getDevDockerImage` - this fixes the undefined, but results in the node version being added to the tag twice.
- Automatically add node suffix to all tags.
- Create `removeNodeSuffixFromTag` function. This turns something like `v0.91.0-nodev18.18.2` into `v0.91.0`.
- Add checks for `options.nodeSuffix` and removes the tag if `false`.

I chose to add the node suffix by default to make the transition simple when we decide that having a tag without a node suffix will never be necessary. At that point we can update `formatDailyTag()` to take `nodeVersion` as a parameter and update its test.

```
with node suffix:
daily         terascope/teraslice:daily-2024.01.12-fe31c87bf9f-nodev16.20.2
dev           terascope/teraslice:dev-local-nodev18.18.2
latest        terascope/teraslice:latest-nodev16.20.2
prerelease    terascope/teraslice:v0.91.1-rc.0-nodev16.20.2
tag           terascope/teraslice:v0.91.0-nodev18.18.2

without node suffix:
daily         terascope/teraslice:daily-2024.01.12-fe31c87bf9f
dev           terascope/teraslice:dev-local
latest        terascope/teraslice:latest
prerelease    terascope/teraslice:v0.91.1-rc.0
tag           terascope/teraslice:v0.91.0
```

### DIFF
--- a/packages/scripts/src/helpers/publish/utils.ts
+++ b/packages/scripts/src/helpers/publish/utils.ts
@@ -107,3 +107,7 @@ export async function buildDevDockerImage(
     signale.success(`built docker image ${devImage}, took ${toHumanTime(Date.now() - startTime)}`);
     return devImage;
 }
+
+export function removeNodeSuffixFromTag(tag: string) {
+    return tag.split('-').filter((part) => !part.startsWith('nodev')).join('-');
+}
--- a/packages/scripts/test/publish-spec.ts
+++ b/packages/scripts/test/publish-spec.ts
@@ -1,11 +1,25 @@
 import 'jest-extended';
-import { formatDailyTag } from '../src/helpers/publish/utils';
+import { formatDailyTag, removeNodeSuffixFromTag } from '../src/helpers/publish/utils';
 
 describe('Publish', () => {
     describe('->formatDailyTag', () => {
         it('should return a correctly formatted tag', async () => {
             const tag = await formatDailyTag();
             expect(tag).toMatch(/^daily-\d{4}.\d{2}.\d{2}-[A-Za-z0-9]{5,12}$/);
+        });
+    });
+
+    describe('->removeNodeSuffixFromTag', () => {
+        it('should return a tag without the node version', async () => {
+            const tagBefore = 'v0.91.0-nodev18.18.2';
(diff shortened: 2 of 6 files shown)


packages/scripts/src/helpers/publish/utils.ts
```diff
@@ -107,3 +107,7 @@ export async function buildDevDockerImage(
     signale.success(`built docker image ${devImage}, took ${toHumanTime(Date.now() - startTime)}`);
     return devImage;
 }
+
+export function removeNodeSuffixFromTag(tag: string) {
+    return tag.split('-').filter((part) => !part.startsWith('nodev')).join('-');
+}
```
godber, line 113: This is the kind of thing that might benefit from having a test.

I guess this splits on `-`, filters out the thing we don't want, then joins with `-` again.  Which seems pretty robust.

godber (approved):

packages/scripts/src/helpers/publish/utils.ts
busma13: I added a few tests.

packages/scripts/test/publish-spec.ts
```diff
@@ -8,4 +8,18 @@ describe('Publish', () => {
...
+
+    describe('->removeNodeSuffixFromTag', () => {
+        it('should return a tag without the node version', async () => {
+            const tagBefore = 'v0.91.0-nodev18.18.2';
+            const tagAfter = await removeNodeSuffixFromTag(tagBefore);
+            expect(tagAfter).toEqual('v0.91.0');
+        });
+
+        it('should not modify a tag without -nodev', async () => {
+            const tagBefore = 'v0.91.0-v18.18.2';
```
godber, line 20: Is this an actual case that would be encountered?  `v0.91.0-v18.18.2`

If not what are more realistic cases that could be tested.

busma13: I could test the daily tags: `daily-2024.01.12-fe31c87bf9f` and `daily-2024.01.12-fe31c87bf9f-nodev16.20.2`
This would test a tag with a hyphen in it.

PR #3533, opened 2024-01-30
E2e typescript esm conversion

jsnoble:
- convert e2e to typescript and to esm module output


jsnoble:
The test framework uses the `teraslice-client` package which is still using its own internal types instead of the ones used in the @types library, need to do a seperate PR to fix that and to then cleanup and types upstream in the tests

packages/scripts/src/helpers/test-runner/index.ts
```diff
@@ -163,18 +162,8 @@ async function runTestSuite(
             tracker.ended += pkgs.length;
             tracker.addError(err.message);
 
-            const teardownPkgs = pkgs.map((pkg) => ({
-                name: pkg.name,
-                dir: pkg.dir,
-                suite: pkg.terascope.testSuite
-            }));
```
godber, line 170: The changes here related to teardownPkgs and cleanup, what has changed made this change possible?  I don't really know what the teardown/cleanup process was doing, but please briefly describe why this is OK.

godber (commented):
@jsnoble I am not requesting any changes, just an explanation about the cleanup and teardown changes.  Feel free to just reply on the PR.

packages/scripts/src/helpers/test-runner/index.ts
godber: We were having scenarios where our test teardown code was called twice during test shutdown.  This cleanup is handled in the `jest` shutdown code instead of in the test runner here.

godber (approved):

Issue #3534, opened 2024-02-01
Add a proper check to ts-scripts that ensures Kafka is running


sotojn:
As of right now when you run `ts-scripts test` with kafka enabled, it will not verify that Kafka is actually running before executing tests.

Right now the checkKafka function doesn't do anything

https://github.com/terascope/teraslice/blob/c1358e8770561b97fc08bbf95b6e0ee9f0d4c428/packages/scripts/src/helpers/test-runner/services.ts#L712-L715

We should implement a check by importing a Kafka client that can connect to the brokers to verify that kafka is up and running.

PR #3535, opened 2024-02-01
Add check to ensure Kafka is running in scripts

sotojn:

- Adds a method in `ts-scripts` tests that checks to see if Kafka is running properly
- Bumps `ts-scripts` from `v0.68.0` to `v0.68.1`

Ref to issue #3534


packages/scripts/src/helpers/test-runner/services.ts
```diff
@@ -710,8 +711,45 @@ async function checkRabbitMQ(options: TestOptions, startTime: number): Promise<v
...
+            maxRetryTime: retryTime,
+            factor: 0,
+            retries: retryCount
+        }
+    });
+    const producer = kafka.producer();
     const took = ts.toHumanTime(Date.now() - startTime);
-    signale.success(`kafka@${options.kafkaVersion} *might* be running at ${config.KAFKA_BROKER}, took ${took}`);
+    try {
+        await producer.connect();
```
godber, line 743: Your choice to use the `producer.connect()` method as your test is probably a good one.  I would have probably started with listing topics with the admin client:

https://kafka.js.org/docs/admin#a-name-list-topics-a-list-topics

But if the main usecase for kafka in the tests require being able to write to it, then connect is probably the right thing to do.  I only mention it just in case this goes wrong for some reason.

godber (approved):

PR #3536, opened 2024-02-08
Add node version to spaces e2e image

busma13:
- Updates `getDevDockerImage()` to add node suffix to the image tag passed in by spaces. This allows the right image to be used in e2e tests. The suffix will be removed in the publish stage of spaces, where it is not needed.
- Add a check in e2e tests to ensure the teraslice node version matches the desired version.

### DIFF
--- a/packages/scripts/src/helpers/misc.ts
+++ b/packages/scripts/src/helpers/misc.ts
@@ -129,7 +129,7 @@ export function getServicesForSuite(suite: string): Service[] {
 }
 
 export function getDevDockerImage(nodeVersion?: string): string {
-    if (DEV_DOCKER_IMAGE) return DEV_DOCKER_IMAGE; // an override so shouldn't change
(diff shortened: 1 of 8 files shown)


packages/scripts/src/helpers/misc.ts
```diff
@@ -129,7 +129,7 @@ export function getServicesForSuite(suite: string): Service[] {
 }
 
 export function getDevDockerImage(nodeVersion?: string): string {
-    if (DEV_DOCKER_IMAGE) return DEV_DOCKER_IMAGE; // an override so shouldn't change
+    if (DEV_DOCKER_IMAGE) return `${DEV_DOCKER_IMAGE}-nodev${nodeVersion}`;
```
godber, line 132: Since we're changing ts-scripts, don't we need to bump the version here?

godber (changes requested):

packages/scripts/src/helpers/misc.ts
busma13: Fixed

godber (approved):

Issue #3544, opened 2024-02-15
isbinaryfile dependency update breaking yarn add teraslice-cli in assets CI workflow


busma13:
`isbinaryfile` updated node engine to >20.0.0 in version `5.0.1`. They then downgrade node version to >18.0.0 in version `5.0.2`. Version 5.0.0 is the last version that supports node 16. 
- In `teraslice-cli`, `isBinaryFile` is required by yeoman-environment (version ^4.0.10) and also by its dependency mem-fs-editor (version ^5.0.0).
- `teraslice-cli` is used by the assets CI to build the assets (`yarn global add teraslice-cli`)
- failed CI example: https://github.com/terascope/standard-assets/actions/runs/7907928098/job/21632825647
- We have tried adding `"isbinaryfile": "5.0.0"` to the resolutions in the root directory as well as in `teraslice-cli`. This resolves the error when installing dependencies in the monorepo, but we still get a node engine error when installing the package using `yarn global`.

godber:
It's worth pointing out that we're just about ready to cut away from Node 16, we have on internal project that hasn't shipped all the way to prod on the Node 18 build, but does work and has been working in test environments.  Hypothetically we could drop 16 support and move on to adding 20 and this problem would go away.  The risk being that we find some reason we can't get that internal project shipped on 18 and need some test coverage on Node 16 still.

godber:
The upstream package rolled back to node 16 here: https://github.com/gjtorikian/isBinaryFile/issues/70

If he rolled back to 16 our problem would go away, but I almost feel bad asking it.

We've tried about 4 different ways of trying to constrain this dependency to and older version without any luck.

godber:
The only additional progress I've made on this is that it's pretty clear that yarn only supports resolutions at the top level `package.json` in the case of workspaces/monorepos.

busma13:
Pulling `teraslice-cli` files out of the monorepo and into their own project, then running `yarn install` results in the correct versions of `isbinaryfile`. I can also run `yarn global add file:<path to standalone teraslice-cli>` and I get the same `yarn.lock` file with the correct versions. It is only when the package is pulled directly from npm that the yarn.lock is built with an `isbinaryfile` with version `5.0.2`.

busma13:
I glossed over this in the beginning, but asset tests pass for MacOS: https://github.com/terascope/standard-assets/actions/runs/7907928098/job/21632825426

godber:
[...]

PR #3546, opened 2024-02-16
downgrade yeoman deps yo dep

sotojn:

- Downgrades the following dependencies in `teraslice-cli` to prevent `isbinaryfile` dependency from going over version `5.0.1`. This is because `isbinaryfile` now discontinues node 16 support.
  - **yeoman-environment** from `^3.15.1` to `3.2.0`
  - **yeoman-generator** from `^5.8.0` to `5.2.0`
  - **yo** from `^4.3.1` to `4.0.0`
 - Adds the following `peerDependencies` inside `teraslice-cli` package.json
   - `isbinaryfile`
   - `mem-fs-editor`
 
refs to issue #3544

### DIFF
--- a/packages/teraslice-cli/package.json
+++ b/packages/teraslice-cli/package.json
@@ -1,7 +1,7 @@
 {
     "name": "teraslice-cli",
     "displayName": "Teraslice CLI",
-    "version": "0.61.4",
+    "version": "0.61.5",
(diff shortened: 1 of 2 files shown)


godber:
Ultimately we have decided to get rid of Node 16, so we won't be merging this PR.  Thanks for exploring this solution Joseph.

PR #3551, opened 2024-02-21
Update workflow - drop 16, set default to 18.19.1, add 20.11.1

busma13:
- Remove node 16 from test matrices
- Add node 20.11.1 to test matrices
- Update tests to use node 18.19.1 (was 18.18.2)
- Remove checks on timezone tests that were needed for node 16
- Disable asset upload stream test because nock hangs in node 20.11.1
- Disable lint rules for disabled tests, duplicate enum values


packages/teraslice-client-js/test/assets-spec.ts
```diff
@@ -64,7 +64,10 @@ describe('Teraslice Assets', () => {
             });
         });
 
-        describe('when called with a stream', () => {
+        // This test hangs in node version 20.10.0 and above.
+        // See issue: https://github.com/nock/nock/issues/2595
+        // teraslice/e2e/test/cases/assets/simple-spec.ts covers this test case
+        xdescribe('when called with a stream', () => {
             const testFilePath = path.join(__dirname, 'fixtures', 'test.txt');
             const contents = fs.readFileSync(testFilePath, 'utf-8');
```
godber, line 72: The linter is complaining about this disabled test, can you convince it not to complain?

godber (changes requested):
Fix new linter error.

packages/teraslice-client-js/test/assets-spec.ts
busma13: Done

godber (approved):

PR #3552, opened 2024-02-21
Add node20 to asset bundle-target choices array, remove node14.

busma13:
Add `node20` to asset `bundle-target` choices array, remove `node14` from array. 
Make `node18` default `bundleTarget`

### DIFF
--- a/packages/teraslice-cli/src/helpers/yargs-options.ts
+++ b/packages/teraslice-cli/src/helpers/yargs-options.ts
@@ -66,11 +66,11 @@ export default class Options {
         'bundle-target': () => ({
             describe: 'Specify which version of node is used as the target for esbuild.'
                     + ' Defaults to current major node version',
-            default: `node${process.version.split('.', 1)[0].substr(1)}`,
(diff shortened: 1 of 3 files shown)


packages/teraslice-cli/src/helpers/yargs-options.ts
```diff
@@ -66,11 +66,11 @@ export default class Options {
         'bundle-target': () => ({
             describe: 'Specify which version of node is used as the target for esbuild.'
                     + ' Defaults to current major node version',
-            default: `node${process.version.split('.', 1)[0].substr(1)}`,
+            default: `node${process.version.split('.', 1)[0].substring(1)}`,
```
godber, line 69: Aha, deprecated ... good find:

* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring

godber (approved):

PR #3558, opened 2024-02-24
fix Error.captureStackTrace if not defined

### LINKED ISSUE #3557: Error.captureStackTrace is not a function in non V8 browsers like safari

lesleydreyer:
Browsers like Chrome running off v8 can captureStackTrace, but browsers like Safari running off javascriptcore don't seem to have this function defined

Closes #3557


jsnoble (approved):
LGTM

PR #3560, opened 2024-02-29
convert teraslice-cli to esm modules

jsnoble:


ciorg:
I get this error `TypeError: TerasliceClient is not a constructor` when running the command trying to run any cli command, for example - `earl controllers list CLUSTER`

ciorg (approved):
LGTM

Issue #3561, opened 2024-02-29
Provide alternative to shared file system cache from Terasilce asset loading - S3


godber:
Right now, workers rely on the master to have unzipped an asset into a shared network drive.  Without this shared disk cache, each worker must retrieve and unzip it's own asset to local disk.  This causes excessive load on the ES state cluster used by the teraslice master.

There might be multiple options to achieving this.  I will list options below.

* S3 based asset storage model
  * Assets should be stored in S3, and workers should retrieve those zipfiles directly from S3.
  * The terafoundation should contain
    *  `asset_storage_connector` - optional, but required if using S3 asset storage, where the operators specify which S3 connector stores the assets
    * `asset_storage_bucket` - optional, default `ts-assets-<TERASLICE NAME>/`
* In master memory asset cache
  * If the workers ask the master for the assets it could cache those in memory rather than reading them out of ES each time.
* S3 based cache model
  * This would replace the unzipped asset file system cache with S3.

Note: I use "workers" above to mean both execution controllers and workers.

@kstaken @jsnoble

busma13:
In terafoundation `connector` seems to refer to the service and `connection` refers to the name of an endpoint for a service. I think it will be clearer to use `asset-storage-connection` instead of `asset-storage-connector`.

busma13:
While setting up error handling for the new `S3Store` class I came across this possible issue: #3565

busma13:
@kstaken suggested that we make this feature more generic, allowing for assets to be stored in other locations in the future. With that in mind, it would make sense to specify the connector and the connection. We were assuming S3 as the connector if there is an `asset-storage-connection` field. `asset-storage-connector` will default to ES, will now work for S3, and could be expanded to use some other storage.

sotojn:
When doing a get request to the `txt/assets` api endpoint, we are thinking about adding an extra column called `external_storage` that indicates if there is an associated s3 object with the es assets metadata. 

Before:
```
name           version  id                                        _created                  description                     node_version  platform  arch
-------------  -------  ----------------------------------------  ------------------------  ------------------------------  ------------  --------  ----
standard       0.22.3   85e2f713c615c74b70f4ebe12a1735e619832e52  2024-03-11T16:11:29.153Z  Teraslice standard processor a  18                          
kafka          3.5.2    4fac0e2bfefd18cb1915fc3aae8cc729b16c4533  2024-03-11T16:11:29.003Z  Kafka reader and writer suppor  18                          
[...]

PR #3563, opened 2024-03-04
Add support for storing assets in S3

busma13:
- Add `S3Store` class to teraslice which will store a zipped asset external to ES. Asset metadata is still stored in ES.
- Adds the following fields to `terafoundation`:
  - `asset_storage_connection_type`
  - `asset_storage_connection`
  - `asset_storage_bucket`
- Add check for `asset_storage_connection_type` in `AssetsStorage` class. If value is `S3` create an `s3Backend` to store assets externally.
- Update `AssetsService` class:
  - Add an `external_storage` column to the get `/txt/assets` and get `/assets` endpoints to show if the asset is `available` or `missing`.
- Add new tests to `simple-spec.ts` in e2e that ensure assets are stored in S3.
- Add `s3_store-spec.ts`, `assets_storage-spec.ts`, and `assets-spec.ts` to teraslice test suite.
- Add new job to github `test` workflow that runs e2e tests with external asset storage.

Ref: #3561


packages/teraslice/src/lib/storage/backends/s3_store.ts
```diff
@@ -0,0 +1,227 @@
...
+        const {
+            context,
+            terafoundation,
+            connection: connector,
+            bucket,
+            logger
+
+        } = backendConfig;
+
+        this.bucket = bucket || 'tera-assets';
```
godber, line 47: This should default to include the teraslice instance `name` (hit the root URL of any teraslice), so like `${name}-assets`.  This will facilitate multiple asset buckets in one S3 instance more easily.

packages/teraslice/src/lib/storage/backends/s3_store.ts
```diff
@@ -0,0 +1,227 @@
...
+    */
+    async verifyClient() {
+        if (this.isShuttingDown) return false;
+        const command = new ListObjectsV2Command({
+            Bucket: this.bucket,
+            MaxKeys: 0
+        });
+        try {
+            await this.api.send(command);
+            this.logger.trace('Client verification successful');
```
godber, line 203: I would:

* make this `debug`.
* include detailed connection information in a log like this, including the URL or something, so the operator doesn't have to look that up.

packages/teraslice/src/lib/storage/backends/s3_store.ts
```diff
@@ -0,0 +1,227 @@
...
+            return true;
+        } catch (err) {
+            this.logger.trace('Client verification failed: ', err);
```
godber, line 206: Make `debug`

packages/teraslice/src/lib/storage/backends/s3_store.ts
busma13: It used to, but teraslice instance names tend to have underscores, which are illegal characters for S3 bucket names. I could strip out illegal characters from the teraslice name?

godber: Sure, convert `_` to `-`.  We may have a utility for this ... or ... at least I feel like I had to do this for some k8s stuff.

packages/terafoundation/src/schema.ts
```diff
@@ -1,70 +1,121 @@
...
+        },
+        workers: {
+            doc: 'Number of workers per server',
+            default: workerCount
+        },
+        asset_storage_connector: {
+            doc: 'Name of the connector used to store assets',
+            default: 'elasticsearch-next',
+            format(connectorName: any): void {
+                const connectorList = Object.keys(sysconfig.terafoundation.connectors);
```
godber, line 77: As written, this will make `kafka` a legitimate value here.

packages/teraslice/src/lib/cluster/services/assets.ts
```diff
@@ -200,11 +284,19 @@ export class AssetsService {
...
             ) as Record<string, any>;
 
-            return results.hits.hits.map((asset: any) => {
+            const mappedRecords = results.hits.hits.map((asset: any) => {
                 const record = asset._source;
                 record.id = asset._id;
                 return record;
             });
+
+            if (this.context.sysconfig.terafoundation.asset_storage_connection) {
```
busma13, line 293: change from connection to connector === 's3'

packages/teraslice/src/lib/storage/assets.ts
```diff
@@ -67,12 +69,28 @@ export class AssetsStorage {
         this.logger = logger;
         // TODO: verify this behavior of string vs string[] and undefined
         this.assetsPath = config.assets_directory as string;
-        this.backend = new TerasliceElasticsearchStorage(backendConfig);
+        this.esBackend = new TerasliceElasticsearchStorage(esBackendConfig);
+
+        if (context.sysconfig.terafoundation.asset_storage_connector === 's3') {
+            if (context.sysconfig.terafoundation.asset_storage_connection) {
```
busma13, line 75: think about this

packages/teraslice/src/lib/storage/assets.ts
```diff
@@ -91,8 +109,12 @@ export class AssetsStorage {
 
     private async _assetExistsInES(id: string): Promise<boolean> {
```
busma13, line 110: change name

packages/teraslice/src/lib/storage/assets.ts
```diff
@@ -111,40 +133,55 @@ export class AssetsStorage {
 
         return readable && exists;
     }
-    // data is the buffer form of asset, stored in filesystem
-    // esData is the base64 version which is stored in elasticsearch
+
```
busma13, line 136:
```suggestion
    // data is the buffer form of asset, stored in filesystem
    // esData is the base64 version which is stored in elasticsearch
```

packages/teraslice/src/lib/storage/assets.ts
```diff
@@ -111,40 +133,55 @@ export class AssetsStorage {
...
         const metaData = await saveAsset(
             this.logger,
             this.assetsPath,
             id,
             data,
-            _metaIsUnique(this.backend)
+            _metaIsUnique(this.esBackend)
         );
 
+        const blobContents = emptyBlob ? '' : esData;
```
busma13, line 165: Try to eliminate the blob key altogether

packages/teraslice/src/lib/storage/assets.ts
```diff
@@ -111,40 +133,55 @@ export class AssetsStorage {
 
         return readable && exists;
     }
-    // data is the buffer form of asset, stored in filesystem
-    // esData is the base64 version which is stored in elasticsearch
+
     private async _saveAndUpload({
         id, data, esData, blocking
```
busma13, line 138: create separate fn that doesn't take esData blob

packages/teraslice/src/lib/storage/backends/s3_store.ts
```diff
@@ -0,0 +1,227 @@
...
+import {
+    BucketAlreadyExists,
+    BucketAlreadyOwnedByYou,
+    CreateBucketCommand,
+    DeleteObjectCommand,
+    GetObjectCommand,
+    ListObjectsV2Command,
+    NoSuchKey,
+    PutObjectCommand
+} from '@aws-sdk/client-s3';
```
godber, line 14: Rather than importing directly from `@aws-sdk/client-s3`, we should probably be extending the `file-asset-apis` to include the things you import above.  I think this is an example of how the package already does so.

https://github.com/terascope/file-assets/blob/master/packages/file-asset-apis/src/s3/client-types/client-params.ts

docs/configuration/overview.md
```diff
@@ -183,3 +187,48 @@ terafoundation:
                 node:
                     - YOUR_ELASTICSEARCH_IP:9200"
 ```
+
+## Configuration Asset Storage
+
+By default asset bundles are stored in Elasticsearch when uploaded. To store assets external to Elasticsearch you must define the `asset_storage_connection_type` within `terafoundation`. If using a connection besides `default`, specify it with the `asset_storage_connection` field.
```
sotojn, line 193: I find the second sentence a little hard to read. 

> To store assets external to Elasticsearch you must define the `asset_storage_connection_type` within `terafoundation`.

Something like this sound a little better when I read it out loud.

Defining the `asset_storage_connection_type` within `terafoundation` will enable the ability to store assets outside of elasticsearch.

packages/teraslice/src/lib/storage/assets.ts
```diff
@@ -179,36 +255,49 @@ export class AssetsStorage {
         };
     }
 
+    // this will work, but wont return the blob containing the asset. It doesn't seem needed.
+    // if blob is needed update to get the asset from s3, then convert it and put in the blob field?
```
busma13, line 259: remove comments

godber:
Is assetSearch called once on job startup or once per worker?

https://github.com/terascope/teraslice/blob/dad488eedd7230805abcb11289463e196834341f/packages/teraslice/src/lib/cluster/services/assets.ts#L219

packages/teraslice/src/lib/storage/assets.ts
```diff
@@ -124,27 +163,54 @@ export class AssetsStorage {
...
             _created: new Date().toISOString()
         }, metaData);
 
-        if (blocking) {
-            const elapsed = Date.now() - startTime;
-            const remaining = responseTimeout - elapsed;
-            await this.backend.indexWithId(id, assetRecord, undefined, remaining);
-        } else {
-            await this.backend.indexWithId(id, assetRecord, undefined, responseTimeout);
+        await this._saveToEs(id, assetRecord, blocking, responseTimeout, startTime);
```
godber, line 174: Technically this would benefit from retries I think but lets leave it as is since the original code didn't either.

sotojn:
> Is assetSearch called once on job startup or once per worker?
> 
> https://github.com/terascope/teraslice/blob/dad488eedd7230805abcb11289463e196834341f/packages/teraslice/src/lib/cluster/services/assets.ts#L219

`assetsearch` is called whenever you hit the `v1/assets` or the `txt/assets`  endpoints. The only other code I see using those endpoints are inisde the `teraslice-client-js`

Also the only thing I see that utilizes the `teraslice-client-js` package is `teraslice-cli`.

packages/teraslice/src/lib/storage/backends/s3_store.ts
```diff
@@ -0,0 +1,258 @@
...
+
+export interface TerasliceS3StorageConfig {
+    context: Context;
+    terafoundation: TerafoundationConfig;
+    connection: string;
+    bucket?: string;
+    logger?: Logger
+}
+
+export class S3Store {
```
godber, line 28: I think we have an issue with the degree to which this class is implemented in a generic manner.  It's not a problem now, but I think `elasticsearch_store` is implemented in a generic manner, whereas `s3_store` has a few minor things in it that make it suitable only for asset storage.  Of course this is the only way we are currently using it.  But I think something architectural is implied by implementing a "store" in this "backend" directory that we might be violating.

I'm mostly talking out loud here.  No changes are requested regarding this.

godber: Also, I think the only real strike against being general is the assumption that all object keys end in `.zip`.

godber (commented):
Lets go through these in the morning.

packages/teraslice/src/lib/storage/assets.ts
busma13: this changed so terafoundation validation makes a default value

godber (approved):

PR #3564, opened 2024-03-05
update graphql to v16

lesleydreyer:
- [x] Upgrade graphql from 14 to 16 

NOTE:  they now have comments and descriptions - comments do not print in schemas anymore as they are more for development comments so switched to descriptions (https://github.com/graphql/graphql-js/pull/2900)


jsnoble (approved):
LGTM

Issue #3567, opened 2024-03-13
e2e tests shared asset_directory can cause test failure.


busma13:
Adding functionality to store assets in an external storage like S3 instead of in Elasticsearch (see: #3563) has surfaced a race condition in e2e tests. 
- `simple-spec` saves assets to the `teraslice.assets_directory` cache.
- `recovery-spec` gets the array of asset IDs needed to create a `jobSpec` from the list of IDs in the `teraslice.assets_directory` cache.
- Since jest runs the two spec files in parallel, `simple-spec` can be adding an asset while `recovery-spec` is reading all the assets in the directory.
- The additional time required to save an asset in S3 slows the post('/assets') route enough to make obvious that an asset can be added to the local cache in `teraslice.assets_directory`, but not yet be stored in Elasticsearch.
- If `recovery-spec` gets an ID that has just been added to the cache, it will add it to `jobSpec.assets` and then run validation on `jobSpec`. If the validation check for the presence of the asset happens before it has been saved to ES by `simple-spec` the test will fail with an error: `Failure to get assets, caused by TSError: No assets found for "2909ec5fd38466cf6276cc14ede25096f1f34ee9"`

busma13:
There are a few ways that we may be able to resolve this issue:
- Use jests `test-sequencer` to ensure that `recovery-spec` runs first.
- Change how the `jobSpec` in `recovery-spec` gets the list of IDs for assets it needs.
- Add a function to the `teraslice-harness` that will create a `jobSpec` that does not need to be modified by the test?

busma13:
I decided the best approach was to add a function to the `teraslice-harness` that returns an array of IDs belonging to the "base assets" that the e2e tests autoload. `recover-spec` calls this function and assigns the return value to `jobSpec.assets`, ensuring that job validation only checks the assets that were autoloaded.

PR #3568, opened 2024-03-14
Fix recovery-spec test to use only the autoloaded assets in the jobSpec

busma13:
- export the array of e2e autoloaded assets
- Use the array of autoloaded assets in `TerasliceHarness.getBaseAssetIds()` to get the IDs of autoloaded assets
- Use `TerasliceHarness.getBaseAssetIds()` in `recovery-spec` to set the `jobSpec.assets` array to only the autoloaded assets.

These changes prevent the `recovery-spec` from trying to use assets placed in the `teraslice.assets_directory` by other tests running in parallel, that have not finished uploading to ES.
Ref: #3567

### DIFF
--- a/e2e/test/cases/data/recovery-spec.ts
+++ b/e2e/test/cases/data/recovery-spec.ts
@@ -47,9 +47,7 @@ describe('recovery', () => {
         }
         jobSpec.name = 'test recovery job';
 
-        const files = await fse.readdir(ASSETS_PATH);
-        jobSpec.assets = files.filter((asset) => asset.length === 40);
-
+        jobSpec.assets = await terasliceHarness.getBaseAssetIds();
         specIndex = terasliceHarness.newSpecIndex('test-recovery-job');
 
         if (!jobSpec.operations) {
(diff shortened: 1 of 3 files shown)


busma13:
This solution works, but I want to look into why we can't just use the array of asset names for `jobSpec.assets`. The initializeTestExecution calls JobValidator.validateConfig(). The JobValidator uses the OperationLoader which expects asset IDs, not assets names. It seems like jobConfig and ExecutionConfig types might be getting muddled.

busma13:
I dug into this and here's what I found out:
- `OperationLoader` uses the asset name it is given as part of the path to the asset.
- `JobValidator` is used exclusively by the `JobsService` except for when it is used by tests. `JobsService._ensureAssets()` converts all `jobSpec.assets` strings to their IDs using `spawnAssetLoader`, so validation always passes because the assets are in a subdirectory named as the asset ID.
- `recovery-spec` is the only test that uses assets from the `teraslice.assets_directory`. So asset names must be converted to asset IDs so that `OperationLoader` has the proper path to the assets.
- Any other assets used in tests are in a directory named as the asset name, so the `OperationLoader` can use the asset name in the asset path to find the information it is looking for.

I think the changes I have made in the PR solve the problem sufficiently.

godber (approved):

Issue #3523, opened 2024-01-12
Unexpected error when starting a job that completes before the `running` status check


ciorg:
When starting a job with the CLI if a job completes before the code verifies that it is running it throws the error:

```
Job cannot reach the target status, "running", because it is in the terminal state, "completed"
```

This only happens on very short jobs, but it can be confusing.  I think the issue is somewhere in here:
https://github.com/terascope/teraslice/blob/9a44c8eae4d1d41ceb96aab80478c197b9e0d09f/packages/teraslice-cli/src/helpers/jobs.ts#L435

PR #3570, opened 2024-03-20
fix error if job finishes before status check

ciorg:
* no longer errors if job finishes before status check
* updated language for job that doesn't have a stopped status on restarting
* minor consistency fixes in message properties


jeffmontagna (approved):
Changes look good.

godber:
https://github.com/terascope/teraslice/issues/3523

PR #3571, opened 2024-03-21
convert packages to esm, migrate esm unit tests to their own workflow

jsnoble:
- [x]  Convert teraslice to esm
- [x] Convert terafoundation to esm
- [x] Convert job-components to esm
- [x] Convert teraslice-messaging to esm
- [x] migrate types from terafoundation
- [x]  migrate types from job-components
- [x]  basic job running
- [x] switch benchmark files for the libraries that have them so they still work
- [x] update from master all major changes
- [x] add hidden backwards compatiblity for getClientAsync in apis.foundation
- [x] verify kafka asset backwards compatibility


packages/job-components/src/operations/core/api-core.ts
```diff
@@ -1,27 +1,27 @@
-import Core from './core';
+import Core from './core.js';
 import {
     ExecutionConfig,
     WorkerOperationLifeCycle,
     SlicerOperationLifeCycle,
-    WorkerContext,
+    Context,
```
godber, line 6: @jsnoble can you describe what motivated this change of `WorkerContext` to `Context`?  Was it just a type consolidation situation?

jsnoble:
Breaking

- got rid of foundation.api.getConnection (use apis.foundation.getClient)
- some foundation apis are now forced to be async, created process and validation are now async
- got rid of legacy processor loading
- got rid of op-test-harness (use modern teraslice-test-harness)

packages/job-components/src/operations/core/api-core.ts
jsnoble: its mainly a type consolidation, as well as the fact that is a little poorly named. If there is a WorkerContext, is there a ExecutionContext? But that has a totally different meaning. The enviroments are the same, and its name does not lean toward a worker or a execution controller or otherwise

packages/teraslice/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "teraslice",
     "displayName": "Teraslice",
-    "version": "1.8.0",
+    "version": "1.9.0",
```
godber, line 4: The teraslice version should be `2.0.0`.  Same for the top level `package.json`.

packages/teraslice/worker-service.js
```diff
@@ -69,16 +67,22 @@ class Service {
...
-    .then(() => cmd.initialize())
-    .then(() => cmd.run())
-    .then(() => cmd.shutdown())
-    .catch((err) => cmd.shutdown(err));
+    try {
+        await cmd.initialize();
+        await cmd.run();
+        await cmd.shutdown();
+    } catch (err) {
+        cmd.shutdown(err);
```
godber, line 84: What if the error is thrown in shutdown on line 82?

packages/types/src/teraslice.ts
```diff
@@ -283,8 +292,15 @@ export interface ValidatedJobConfig {
     volumes?: Volume[];
     /** This will only be available in the context of k8s */
     kubernetes_image?: string;
+    /** This will only be available in the context of k8s */
+    prom_metrics_enabled?: boolean;
+    /** This will only be available in the context of k8s */
+    prom_metrics_port?: number;
+    /** This will only be available in the context of k8s */
+    prom_metrics_add_default?: boolean;
```
godber, line 300: Were these deleted from elsewhere?  Or do they appear due to a type consolidation?

godber (approved):
I think we can probably merge this, most of my comments are food for thought or dumb questions.

PR #3572, opened 2024-03-21
WIP: add date math to xlucene date range queries

lesleydreyer:
- [x] Add date math support to getValidDate so document-matcher, data-entity, and whatever uses it can use date-math
- [x] Coerce ranges in xlucene parser so date math is coerced to a date


jsnoble (approved):
LGTM

Issue #3573, opened 2024-03-21
Add minio as an optional service in k8s-env


sotojn:
A new feature has been implemented to enable the use of an S3 bucket for storing asset zip files. However, we currently lack a method for testing this new feature locally within Kubernetes.

I propose we can accomplish this by doing the following changes:

- Develop a YAML resource that defines the configuration for an S3 service within the Kubernetes cluster.
- Update the existing k8s-env code to incorporate support for the new YAML resource.
  - Then implement the enabling/disabling of the S3 service resource when running `yarn k8s-env` CLI command

PR #3574, opened 2024-03-21
Add minio service to k8s-env 

sotojn:

- Adds a working `minio` to local k8s environment
  - Running the `yarn k8s` command now has the added option of using minio as a service
- Adds `--asset-storage` option flag for k8s command that allows the user to chose which asset storage backend to use
- Implements error handling for conflicting k8s options
- Bumps **@terascope/scripts** from `v0.73.0` to `v0.74.0`

Ref to issue #3573


packages/teraslice/test/services/assets-spec.ts
```diff
@@ -43,11 +44,12 @@ describe('Assets Service', () => {
...
     context.sysconfig.teraslice.api_response_timeout = 30000;
-    /// Setting port for the asset service
-    process.env.port = '55678';
-    const service = new AssetsService(context);
+    let service: AssetsService;
 
     beforeAll(async () => {
+        /// Setting port for the asset service
+        process.env.port = String(await findPort());
+        service = new AssetsService(context);
```
godber, line 52: Why are these changes here?

sotojn: This was something I threw in to potentially resolve that test from failing randomly in CI. I should just take this out, write an issue and make a separate PR for this.

godber: Let's pull this out since it's unrelated and I think we want to talk more about it.  At least I don't want to evaluate it in the context of this pr.

godber (approved):

Issue #3578, opened 2024-03-28
Fix asset_storage_connection default connection validation error


godber:
When testing the Teraslice `v1.1.1` release we encountered an issue when starting a Teraslice master without `asset_storage_connection` or `asset_storage_connection_type` set.  It threw this error.

```
/app/source/packages/terafoundation/dist/src/validate-configs.js:56
        throw new utils_1.TSError(err, { reason: 'Error validating configuration' });
              ^

TSError: Error validating configuration, caused by Error: asset_storage_connection: default not found in terafoundation.connectors.elasticsearch-next: value was "default"
    at validateConfig (/app/source/packages/terafoundation/dist/src/validate-configs.js:56:15)
    at validateConfigs (/app/source/packages/terafoundation/dist/src/validate-configs.js:92:29)
    at new ClusterContext (/app/source/packages/terafoundation/dist/src/cluster-context.js:45:58)
    at file:///app/source/packages/teraslice/service.js:45:5
    at ModuleJob.run (node:internal/modules/esm/module_job:195:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:336:24)
    at async importModuleDynamicallyWrapper (node:internal/vm/module:429:15) {
  fatalError: false,
  statusCode: 500,
  context: {},
  code: 'INTERNAL_SERVER_ERROR'
}

Node.js v18.19.1
```

PR #3579, opened 2024-03-29
Asset storage connection validation bug fix

sotojn:

- Fixes issue where `terafoundation` validation fails if `elasticsearch-next` is the connection_type without a default connection
- Added test that checks said scenario 
- Removed `elasticsearch`  from the list of valid asset storage connection types

Ref to issue #3578

### DIFF
--- a/packages/terafoundation/src/schema.ts
+++ b/packages/terafoundation/src/schema.ts
@@ -100,9 +100,9 @@ export function foundationSchema(sysconfig: SysConfig<any>): convict.Schema<any>
                 /// Check to make sure the asset_storage_connection exists inside the connector
                 /// Exclude elasticsearch as this connection type does not utilize this value
                 if (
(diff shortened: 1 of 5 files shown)


packages/terafoundation/src/schema.ts
busma13, line 1: The array on line 78 no longer needs `elasticsearch`

packages/terafoundation/test/validate-configs-spec.ts
```diff
@@ -389,4 +389,51 @@ describe('Validate Configs', () => {
...
+        const cluster = {
+            isMaster: true,
+        };
+        const config = {
+            config_schema() {
+                return {};
+            }
+        };
+
+        const validedConfig = validateConfigs(cluster as any, config as any, configFile as any);
```
busma13, line 413: change name to validated

jsnoble (approved):
LGTM

PR #3580, opened 2024-04-01
fix range coercion for infinity

lesleydreyer:
Introduced range coercion for date math support in a previous version but it was failing to resolve variables if the node value is  "*"


jsnoble (approved):
LGTM

PR #3583, opened 2024-04-02
Consider node version when calling deleteOlderAssets() in e2e

busma13:
When e2e tests download the latest version of default assets to autoload, the `deleteOlderAssets` function would run. The node version was not being checked, resulting in all duplicate assets with the same version being erased, regardless of node version.

This function will now delete all autoload files that do not match the teraslice node version before checking for older asset versions.

Ref: #3566

### DIFF
--- a/e2e/test/download-assets.ts
+++ b/e2e/test/download-assets.ts
@@ -50,7 +50,8 @@ export const defaultAssetBundles = [
 ];
 
 function assetFileInfo(assetName: string): AssetInfo {
-    const [name, version, , assetNodeVersion] = assetName.split('-');
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const [name, version, _, assetNodeVersion] = assetName.split('-');
     return {
         name,
         version: semver.coerce(version),
--- a/e2e/test/download-assets.ts
+++ b/e2e/test/download-assets.ts
@@ -6,6 +6,16 @@ import { downloadRelease } from '@terascope/fetch-github-release';
 import signale from './signale.js';
 import { AUTOLOAD_PATH } from './config.js';
 
+type AssetInfo = {
+    name: string;
+    version: semver.SemVer | null;
+    newerVersion?: semver.SemVer | null;
+    repo: string;
+    bundle: boolean;
+    fileName: string;
+    assetNodeVersion: string;
+}
+
 /**
  * This will get the correct teraslice node version so
  * we can download the correct asset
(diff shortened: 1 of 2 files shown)


e2e/test/download-assets.ts
```diff
@@ -39,24 +49,33 @@ export const defaultAssetBundles = [
     }
 ];
 
-function assetFileInfo(assetName: string) {
-    const [name, version] = assetName.split('-');
+function assetFileInfo(assetName: string): AssetInfo {
+    const [name, version, , assetNodeVersion] = assetName.split('-');
```
sotojn, line 53: Extra comment here that needs to be removed

e2e/test/download-assets.ts
```diff
@@ -39,24 +49,33 @@ export const defaultAssetBundles = [
...
 }
 
-function getOlderAsset(assets: any[], assetName: string) {
+function getOlderAssets(assets: AssetInfo[], assetName: string): AssetInfo[] {
     const { name, version } = assetFileInfo(assetName);
 
-    return assets.find((a) => {
-        if (a.name !== name) return false;
-        // @ts-expect-error TODO: fix this
-        return semver.gt(version, a.version) && a.bundle;
```
sotojn, line 59: I'm curious about the `a.bundle` thing. Is your argument for removing this because we now exclusively have bundled assets and it makes the assumption they're all bundled?

e2e/test/download-assets.ts
busma13: It's actually meant to be there. I am using array destructuring to assign the value at index 3 of an array like this, `[standard, v0.23.1, node, 18, bundle.zip]`, to `assetNodeVersion`. Index 2 contains the string `node` which is not used.

busma13: Would assigning index 2 to an underscore make it more clear? I've heard it is a common practice in some codebased to use an underscore for an unused variable.

`const [name, version, _, assetNodeVersion] = assetName.split('-');`

e2e/test/download-assets.ts
busma13: That is my assumption. In theory someone could create an asset that isn't bundled and add it to their e2e autoload directory. In that case it would be removed if it had the same name and older version number than another autoload asset.

e2e/test/download-assets.ts
godber: Using a `_` is pretty common in Python, I would have probably understood it, but might have asked you about it regardless.  The intent is MORE obvious when you have multiple `_` ... otherwise someone might think you're just being lazy about variable names (though they don't get used later).

godber: I'd do the `_` thing if your editor lets you.

e2e/test/download-assets.ts
godber: If we are or already have broken non-bundled assets I think that implies we ought to do a lot of cleanup of the other cases where we don't use them ... like in some examples or simple asset tests.  Maybe this doesn't really imply all that much though.

e2e/test/download-assets.ts
busma13: eslint doesn't like the unused variable.
I would have to add an ignore like this:
```
// eslint-disable-next-line @typescript-eslint/no-unused-vars
const [name, version, _, assetNodeVersion] = assetName.split('-');
```

e2e/test/download-assets.ts
```diff
@@ -39,24 +49,33 @@ export const defaultAssetBundles = [
...
+    // If we have no version we can't compare to other versions
+    if (version === null) {
+        return [];
+    }
+    return assets.filter((other) => {
+        if (other.name !== name) return false;
+
+        // Asset file name should always contain node version. If it doesn't, lets delete it.
```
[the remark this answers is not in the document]
godber: The only thing this function (and whole file) can download is listed in `defaultAssetBundles` right?  So I think you're asking if this is true for `teraslice/elasticsearch-assets`, `teraslice/kafka-assets`, and `teraslice/standard-assets`.  Right now you are only going to get assets that have versions, right?  This is safe at the moment, but would have to change if we stopped including node versions in those bundled assets.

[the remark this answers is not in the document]
godber: That's a use case imposed by you ... you have to decide.  I have no idea what crazy stuff you might put in there.

godber (approved):

Issue #3471, opened 2023-11-21
JSON  API _inactive fails with invalid job config


jeffmontagna:
JSON API: `POST /v1/jobs/{jobId}/_inactive`

When `_inactive` is posted to a job the action will fail because there is a check that the job config is valid.  The post will fail with a `500` and the error `Validation failed for job config`.  In some cases connectors used by a job may have been removed from the cluster configuration, and it become impossible to set a job to inactive.  `_inactive` should probably bypass the job validation or allow for a bypass as option.

godber:
Assigning back to Jeff so he can validate this when we get this rolled out somewhere.

jeffmontagna:
The fix for setting a job inactive works without an error.  
- Tested on a teraslice cluster running `v1.3.0`.  
- Removed kafka connector `kafka_test1` from terafoundation and restarted teraslice master.
- Setting job to inactive does not error with `POST /v1/jobs/{jobId}/_inactive`
```bash
curl -sS -X POST ts-test/v1/jobs/1c5056b2-9264-4767-8005-c8b9811233c3/_inactive | jq
{
  "name": "datagen-store-kafka",
  "lifecycle": "once",
  "workers": 2,
  "slicers": 1,
  "cpu": 2,
  "memory": 4294967296,
  "assets": [
    "standard:0.22.5",
    "kafka:3.5.2"
  ],
  "operations": [
    {
      "_op": "data_generator",
      "size": 6000000,
      "set_id": "base64url",
      "stress_test": false
    },
    {
      "_op": "set_key"
    },
    {
      "_op": "kafka_sender",
      "topic": "topic-tst-1",
      "size": 15000,
      "connection": "kafka_test1"
    }
  ],
  "_created": "2024-05-03T22:19:55.565Z",
  "job_id": "1c5056b2-9264-4767-8005-c8b9811233c3",
  "_context": "job",
  "_updated": "2024-05-03T22:36:35.131Z",
  "active": false
}
```
- Setting job to active with `POST /v1/jobs/{jobId}/_active` does error, which is expected as the connector is not in the terafoundation.
```bash
curl -sS -X POST ts-test/v1/jobs/1c5056b2-9264-4767-8005-c8b9811233c3/_active | jq 
{
  "error": 500,
[...]

PR #3592, opened 2024-04-12
add bypass option to skip validation for _inactive jobs

busma13:
- When marking a job as `_inactive` using the JSON API, job validation will be skipped. An old job may not have the needed connectors, assets, etc to pass validation, but we still want to be able to mark it as inactive.
- This will only skip validation when switching active from true or undefined to false.
- Bump teraslice from **1.2.0** to **1.2.1**

Ref: #3471

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "teraslice-workspace",
     "displayName": "Teraslice",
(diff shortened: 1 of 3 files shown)


godber (approved):
This looks right.

Issue #3595, opened 2024-04-18
Execution controller OOMs when large asset is used in job


godber:
We have been doing further testing of the S3 backed asset store and we recently tested with an internal asset that was 60MB zipped.  Unzipped the asset had the following composition:

```
.
├── [ 192]  __static_assets
│   ├── [ 22M]  data1.json.gz
│   ├── [7.1K]  data2.txt
│   ├── [1.2K]  data3.json
│   └── [ 37M]  data4.json.gz
├── [ 200]  asset.json
└── [7.5M]  index.js
```

It should be sufficient to create a mock asset with roughly the same characteristics and get a job to start up with this asset.  The execution controller should then OOM when run in k8s using the default memory limit of `512MB`.  We tested increasing the memory limit (to 6GB) and the execution controller did not OOM.

Here is some of the log output:

```
[2024-04-17T22:41:06.009Z] DEBUG: teraslice/18 on ts-exc-datagen-100m-noop-tmp1-3d18bacc-9e6d-7fsdc: getting record with id: 46f47558baf4e3f0e8f736ad5c91827a53cc4b4b from s3 minio_test1 connection, ts-assets-teraslice-tmp1 bucket. (assignment=execution_controller, module=assets_storage, worker_id=97W8Ruer, ex_id=77c23ff5-409e-4b45-a264-c089fd90b3e1, job_id=3d18bacc-9e6d-4651-96bf-5fffec667073)
[2024-04-17T22:41:06.533Z]  INFO: teraslice/18 on ts-exc-datagen-100m-noop-tmp1-3d18bacc-9e6d-7fsdc: loading assets: a5b3d9e48bce3b5f997ba7c21cb3d47945e231a2 (assignment=execution_controller, module=asset_loader, worker_id=97W8Ruer, ex_id=77c23ff5-409e-4b45-a264-c089fd90b3e1, job_id=3d18bacc-9e6d-4651-96bf-5fffec667073)
[...]

PR #3598, opened 2024-04-23
Large asset Ex OOM fix when in s3 asset mode

sotojn:

- Improves the s3 backend get() requests to grab assets in a more memory efficient way
  - This resolves and issue where pulling and decompressing large assets from s3 would cause and OOM on the execution controller on job startup 
- Add error message when asset loader would close with an error that advises what to do in the case of an OOM issue 

Ref to issue #3595

### DIFF
--- a/e2e/test/cases/assets/simple-spec.ts
+++ b/e2e/test/cases/assets/simple-spec.ts
@@ -1,5 +1,9 @@
 import 'jest-extended';
-import { createReadStream } from 'node:fs';
+import fs from 'node:fs';
+import os from 'os';
+import path from 'path';
+import decompress from 'decompress';
+import archiver from 'archiver';
 import {
     createS3Client,
     getS3Object,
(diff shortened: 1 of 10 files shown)


godber:
Using this branch with ES backed asset storage, I started the cluster with this command:

```
yarn k8s:minio
```

I uploaded the asset fine:

```
earl assets deploy local -f autoload/common_processors-v0.13.1-node-18-bundle.zip
Asset posted to local: eabe46a623bc55886e1e81f3eefe74754a903fd1
```

When trying to run in ES mode I get the following error when the job is registered:

```
earl tjm register local examples/jobs/data_generator.json
Error Failure to get assets, caused by TSError: index_not_found_exception
    at _errorHandlerFn (/app/source/packages/elasticsearch-api/index.js:840:21)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
Caused by: ResponseError: index_not_found_exception
    at IncomingMessage.<anonymous> (/app/source/node_modules/elasticsearch6/lib/Transport.js:310:25)
    at IncomingMessage.emit (node:events:529:35)
    at endReadableNT (node:internal/streams/readable:1400:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
If running out of memory, try consider increasing the memory allocation for the process by adding/modifying the "memory_execution_controller" or "resources_limits_memory" (for workers) field in the job file. registering job Data Generator on http://localhost:5678
```

godber (approved):

PR #3607, opened 2024-05-08
Refactor PromMetrics to use Collect functions

busma13:
- Refactor `addMetric` into individual functions for counter, gauge, and histogram.
- Add `collect()` parameter to all `add<metric>` functions.
- Update types: 
    - import `prom-client` types into `types` package
    - remove duplicate types from `terafoundation` and `job-components` and import from `types`. This required some refactoring to not use context as a parameter, as the `context` type differs between `terafoundation` and `job-components`.
 - Move `info` metrics of slice, worker, and master into the `setPromMetrics()` functions.
 - Update PromClient docs and all relevant tests


godber (approved):
I think this looks good

godber:
I would like to merge this but I would like @jsnoble to double check some of the lower level type changes.  I think he's seen these recently but I didn't want to merge without an OK from him.

godber (approved):

PR #3613, opened 2024-05-14
Prom client in mock prom metrics

busma13:
- add `teraslice_` prefix to `promMetrics.init()` for master so all metric names match what was in `teraslice-exporter` package.
- update `mockPrommetrics` to use `prom-client`. This allows `collect()` functions to be used in `TestContext`.

### DIFF
--- a/packages/teraslice/src/lib/cluster/cluster_master.ts
+++ b/packages/teraslice/src/lib/cluster/cluster_master.ts
@@ -150,7 +150,8 @@ export class ClusterMaster {
                     tf_prom_metrics_enabled: terafoundation.prom_metrics_enabled,
                     tf_prom_metrics_port: terafoundation.prom_metrics_port,
                     logger: this.logger,
-                    assignment: 'cluster_master'
+                    assignment: 'master',
+                    prefix: 'teraslice_'
                 });
 
                 await this.setupPromMetrics();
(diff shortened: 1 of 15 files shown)


packages/teraslice/src/lib/workers/execution-controller/index.ts
```diff
@@ -1113,9 +1113,7 @@ export class ExecutionController {
      * @link https://terascope.github.io/teraslice/docs/development/k8s#prometheus-metrics-api
      */
     async setupPromMetrics() {
-        this.logger.info(`adding ${this.context.assignment} prom metrics...`);
```
godber, line 1116: You don't need to add it back right now but I think logging, at the `info` level, that prom metrics are enabled (once at "init" or "setup") makes sense.

Generally speaking a log for every optional feature that is "enabled" is useful, if it occurs ONCE per run.

packages/teraslice/src/lib/cluster/cluster_master.ts
```diff
@@ -150,7 +150,8 @@ export class ClusterMaster {
                     tf_prom_metrics_enabled: terafoundation.prom_metrics_enabled,
                     tf_prom_metrics_port: terafoundation.prom_metrics_port,
                     logger: this.logger,
-                    assignment: 'cluster_master'
+                    assignment: 'master',
+                    prefix: 'teraslice_'
```
godber, line 154: These are the "API" changes that are relevant to document in the CHANGELOG.

godber (approved):

packages/teraslice/src/lib/workers/execution-controller/index.ts
busma13: I must have deleted that accidentally. I'll add it back in.

Issue #3622, opened 2024-05-21
Node 22 'Import assertions' no longer allowed


sotojn:
When trying to build a docker image using node 22.2.0, I come across this error:

```
MacBook-Pro teraslice % docker compose up -d --build
[+] Building 53.6s (18/28)                                                                                                                                                                                                                       docker:desktop-linux
 => [teraslice-worker internal] load build definition from Dockerfile                                                                                                                                                                                            0.0s
 => => transferring dockerfile: 1.16kB                                                                                                                                                                                                                           0.0s
 => [teraslice-master internal] load build definition from Dockerfile                                                                                                                                                                                            0.0s
 => => transferring dockerfile: 1.16kB                                                                                                                                                                                                                           0.0s
 => [teraslice-worker internal] load metadata for docker.io/terascope/node-base:22.2.0                                                                                                                                                                           0.7s
 => [teraslice-master internal] load .dockerignore                                                                                                                                                                                                               0.0s
 => => transferring context: 336B                                                                                                                                                                                                                                0.0s
 => [teraslice-worker internal] load .dockerignore                                                                                                                                                                                                               0.0s
 => => transferring context: 336B                                                                                                                                                                                                                                0.0s
 => [teraslice-master internal] load build context                                                                                                                                                                                                               0.1s
 => => transferring context: 104.34kB                                                                                                                                                                                                                            0.1s
 => [teraslice-worker  1/11] FROM docker.io/terascope/node-base:22.2.0@sha256:071e1ca28a232d6c18cff1157df5b164f2705f95624ad6df7db5d48e5b645120                                                                                                                   0.0s
 => [teraslice-worker internal] load build context                                                                                                                                                                                                               0.1s
 => => transferring context: 104.34kB                                                                                                                                                                                                                            0.1s
 => CACHED [teraslice-master  2/11] COPY package.json yarn.lock tsconfig.json .yarnrc /app/source/                                                                                                                                                               0.0s
 => CACHED [teraslice-master  3/11] COPY .yarnclean.ci /app/source/.yarnclean                                                                                                                                                                                    0.0s
 => CACHED [teraslice-master  4/11] COPY .yarn /app/source/.yarn                                                                                                                                                                                                 0.0s
 => [teraslice-master  5/11] COPY packages /app/source/packages                                                                                                                                                                                                  0.2s
 => [teraslice-master  6/11] COPY scripts /app/source/scripts                                                                                                                                                                                                    0.0s
[...]

Issue #3605, opened 2024-05-06
Adding node 22 to CI test suite


sotojn:
We need to add node 22 to the list of node versions teraslice tests. There are several separate repos on top of teraslice that need to be changed in order to properly include a new node version. So far we don't have documentation on this topic so we should also include that in the development docs while we add node 22.

NOTE: It's required to update the base image [here](https://github.com/terascope/base-docker-image/issues/24) before moving on with doing the following steps.

We need to update the arrays in the following files:
https://github.com/terascope/teraslice/blob/879af4247784a39449e536ca3a79165a3c902469/.github/workflows/publish-tag.yml#L45
https://github.com/terascope/teraslice/blob/879af4247784a39449e536ca3a79165a3c902469/.github/workflows/test.yml#L35
https://github.com/terascope/teraslice/blob/879af4247784a39449e536ca3a79165a3c902469/.github/workflows/test.yml#L58
https://github.com/terascope/teraslice/blob/879af4247784a39449e536ca3a79165a3c902469/.github/workflows/test.yml#L86
https://github.com/terascope/teraslice/blob/879af4247784a39449e536ca3a79165a3c902469/.github/workflows/test.yml#L120C33-L120C40
https://github.com/terascope/teraslice/blob/879af4247784a39449e536ca3a79165a3c902469/.github/workflows/test.yml#L154
https://github.com/terascope/teraslice/blob/879af4247784a39449e536ca3a79165a3c902469/.github/workflows/test.yml#L188
https://github.com/terascope/teraslice/blob/879af4247784a39449e536ca3a79165a3c902469/.github/workflows/test.yml#L226

Adding `22.1.0` to all of these will include node 22 in all of the CI tests and also build and push a node 22 teraslice image to docker hub.

godber:
Note, these failures are not due to a missing image, but because running yarn fails
https://github.com/terascope/teraslice/actions/runs/8975998521/job/24655637262

godber:
@sotojn added some docs here: 

https://github.com/terascope/teraslice/blob/master/docs/development/node.md

But they don't show up in the site under the development page:

https://terascope.github.io/teraslice/docs/development/overview

There's probably an index somewhere or some frontmatter we need to add to get it to show up.

godber:
Oh, it shows up as it's own page here:

https://terascope.github.io/teraslice/docs/development/node

Maybe the use of `sidebar_label: Node` somehow breaks it out of the Development tab its supposed to be under.

PR #3628, opened 2024-05-23
Add node 22.2.0

sotojn:
# New Changes
- Adds node `22.2.0` support 
- Removes experimental `json` imports with standard typescript imports
  - This removes the need to use `assert` to grab objects out of `.json` files 
  - #3622
- Adds **nan** `v2.19.0` to resolutions for node 22 compatibility
- Updates docs to have a new `.md` file that gives detailed instructions on how to introduce new node versions of node
- **Removes** outdated `generator-teraslice` package
- Release: (minor) **teraslice** `v1.8.0`

Ref to issue #3605


Dockerfile
```diff
@@ -14,6 +14,9 @@ COPY packages /app/source/packages
 COPY scripts /app/source/scripts
 COPY types /app/source/types
 
+# Check to see if distutils is installed because python 3.12 removed it
+RUN python -c "import distutils" || apk update && apk add py3-setuptools
```
godber, line 18:
```suggestion
RUN python -c "import distutils" || apk --no-cache add py3-setuptools
```

If this works I'd try that, otherwise I think we end up adding package caches laying around in the image.

sotojn:
Should this be a `minor` or a `patch` bump for teraslice? I'm thinking minor because I consider node 22 support as a new feature, but it overall does not contribute much to the api at all. So there is an argument for a patch.

godber:
I would say bump the minor.  Adding support for Node 22 is the primary goal of this PR, you've made a few small changes related to that.  Since you're adding features rather than fixing bugs, you bump minor.

godber:
I've cancelled the current workflow on this and will retry after the images being built here:

https://github.com/terascope/base-docker-image/actions/runs/9278045777

are completed.

godber (approved):

Issue #3596, opened 2024-04-18
Failure initializing assets_service does not cause a fatal fault


briend:
If you use s3 for asset storage (`asset_storage_connection_type: s3`) and misconfigure the s3 connector with an incorrect `certLocation` or `caCertificate`, the teraslice master container does not exit with an error (but also does not function, obviously).  Normally I'd expect a crash-looping Pod.

example master logs for invalid `certLocation`:

```
[2024-04-17T00:22:56.578Z]  INFO: teraslice/7 on teraslice-tmp1-master-5cf46cd4bf-5wrf6: Service starting (assignment=node_master)
[2024-04-17T00:22:56.580Z] DEBUG: teraslice/7 on teraslice-tmp1-master-5cf46cd4bf-5wrf6: messaging service configuration for assignment node_master (assignment=node_master, module=node_master)
[2024-04-17T00:22:56.580Z]  INFO: teraslice/7 on teraslice-tmp1-master-5cf46cd4bf-5wrf6: node teraslice-tmp1-master-5cf46cd4bf-5wrf6 is attempting to connect to cluster_master: http://teraslice-tmp1:5678 (assignment=node_master, module=node_master)
[2024-04-17T00:22:56.593Z] DEBUG: teraslice/7 on teraslice-tmp1-master-5cf46cd4bf-5wrf6: client network connection is online (assignment=node_master, module=node_master)
[2024-04-17T00:22:56.593Z] DEBUG: teraslice/7 on teraslice-tmp1-master-5cf46cd4bf-5wrf6: node teraslice-tmp1-master-5cf46cd4bf-5wrf6 is creating the cluster_master (assignment=node_master, module=node_master)
[2024-04-17T00:22:56.594Z]  INFO: teraslice/7 on teraslice-tmp1-master-5cf46cd4bf-5wrf6: Starting 1 cluster_master (assignment=node_master)
[2024-04-17T00:22:56.599Z] DEBUG: teraslice/7 on teraslice-tmp1-master-5cf46cd4bf-5wrf6: node teraslice-tmp1-master-5cf46cd4bf-5wrf6 is creating assets endpoint on port 45679 (assignment=node_master, module=node_master)
[2024-04-17T00:22:56.599Z]  INFO: teraslice/7 on teraslice-tmp1-master-5cf46cd4bf-5wrf6: Starting 1 assets_service (assignment=node_master)
[...]

PR #3636, opened 2024-05-30
Add node_master child process shutdown logic

sotojn:

- Fixes issue where the `node_master` process  doesn't shutdown in the event that the `cluster_master` or the `asset_service` child process exits with an error 
  - This will result in the master pod being left in the `running` state instead of `failed` with a `restartCrashLoop`
  - #3596

### DIFF
--- a/packages/teraslice/src/lib/cluster/node_master.ts
+++ b/packages/teraslice/src/lib/cluster/node_master.ts
@@ -2,6 +2,7 @@ import ms from 'ms';
 import _ from 'lodash';
 import { Mutex } from 'async-mutex';
 import { getFullErrorStack } from '@terascope/utils';
+import { Terafoundation } from '@terascope/types';
 import { makeLogger } from '../workers/helpers/terafoundation.js';
 import { Messaging } from './services/cluster/backends/native/messaging.js';
 import { spawnAssetLoader } from '../workers/assets/spawn.js';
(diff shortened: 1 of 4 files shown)


packages/teraslice/src/lib/workers/helpers/worker-shutdown.ts
```diff
@@ -99,7 +99,12 @@ export function shutdownHandler(
 
     async function exit(event: string, err?: Error) {
         if (api.exiting) return;
-
+        /// Potential logic for cluster_master and asset_service
+        if (err) {
+            if (err.name.includes('Error')) {
+                setStatusCode(1);
```
busma13, line 105: I I don't understand why this is necessary. Don't the three functions that call `exit` and pass in an error already set the status code to 1?

busma13: The exit function is also called from `cluster-service.js` and `worker-service.js` so this is necessary.

jsnoble (approved):
LGTM

packages/teraslice/src/lib/cluster/node_master.ts
```diff
@@ -396,19 +397,31 @@ export async function nodeMaster(context: ClusterMasterContext) {
     if (context.sysconfig.teraslice.master) {
         logger.debug(`node ${context.sysconfig._nodeName} is creating the cluster_master`);
```
godber, line 398: I worry about adding any additional opportunities for the master to exit when it shouldn't, but I believe the code here is only called on cluster master start up, this log message reinforces that belief.

godber: Peter points out that while this code only runs at startup, it is creating an event handler that will persist and will execute at any time based on the events coming from subprocess behavior.

Thinking through this from the kubernetes perspective.  There are three processes:

* Node Master - started by k8s
* Cluster Master - started by Node Master
* Asset Service - started by Node Master

If either the Cluster Master or Asset Service exit, the Node Master should exit because it won't work without those subprocesses.  That being said, there are probably two sets of errors, those that are recoverable and those that are not.  So in the case where a recoverable error is encountered, it should restart the failed process.  I don't think we really know what these are at this point.  But if we have problems with the Cluster Master failing, we can start looking at handling individual errors here.

godber (approved):

godber:
I'm going to merge this despite the risk it adds to teraslice master restarts.  We should make note of this possibility in the release notes.

Issue #2673, opened 2021-04-28
Job status is running when k8s node running execution controller pod dies


godber:
When the kubernetes node running an execution controller for a job dies, that job is left in the running state, calling stop on the job doesn't work (the execution controller pod gets stuck in terminating).  Force deleting the execution controller pod will remove the pod and then the job will go to the failed or stopped state.  Or something close to that.  We need to add some sort of execution controller watchdog capability to the k8s cluster master.  This was present in native clustering.

godber:
You might be able to reproduce as follows:

> One specific example that is easily reproducible is use the es reader to read from an index that doesn't exist. The job will get stuck in initializing.

godber:
The example above is a little weak since there are scenarios where a missing index isn't necessarily an error condition.

Despite that. what we are looking for here is ANY scenario where the job tries to start and the execution controllers and workers experience a terminal error and the job doesn't get put into the `failed` state.  I think we have a number of conditions where this is possible.

godber:
Maybe we can write an asset that deliberately generates errors in different parts of the slicer, or fetcher or wherever.

sotojn:
I decided to try testing out using a job that reads into a non existent index and writes it into a new index:

```json
{
    "name": "failing-job",
    "lifecycle": "persistent",
    "workers": 2,
    "assets": [
        "elasticsearch"
    ],
    "operations": [
        {
            "_op": "elasticsearch_reader",
            "index": "joseph_index",
            "date_field_name": "created"
        },
        {
            "_op": "elasticsearch_bulk",
            "size": 10000,
            "index": "joseph_index2"
        }
    ]
}
``` 

Here is the results after running the job for a couple seconds:

Command: `earl ex list localhost `

```
 name         lifecycle   slicers  workers  _status  ex_id                                 job_id                                _created                  _updated                 
 failing-job  persistent  1        2        failed   de64394f-b32d-4680-90a0-5b7ae6e7b386  e2566fb1-963b-45a8-801e-07aeec18736d  2024-06-03T15:42:54.724Z  2024-06-03T15:43:03.718Z 
```

Command: `kubectl get all -n ts-dev1 `

```
NAME                                                    READY   STATUS      RESTARTS   AGE
pod/teraslice-master-67f49bfd75-s92bd                   1/1     Running     0          2d17h
pod/ts-exc-failing-job-e2566fb1-963b-2z7bb              0/1     Completed   0          56m
[...]

PR #3641, opened 2024-06-05
Add resource cleanup on ex error scenario

sotojn:

- Adds resource cleanup on ex error scenario
  - This adds a functionality to the shutdown function on the `ex controller` that checks to see if shutdown was caused by a specific error. Then proceeds to tell the `cluster master` to cleanup the resources.
  - #2673

### DIFF
--- a/packages/teraslice-messaging/src/cluster-master/server.ts
+++ b/packages/teraslice-messaging/src/cluster-master/server.ts
@@ -97,7 +97,6 @@ export class Server extends core.Server {
             });
         });
 
-
         this.handleResponse(socket, 'cluster:analytics', (msg: core.Message) => {
             const data = msg.payload as i.ExecutionAnalyticsMessage;
             if (!this.clusterAnalytics[data.kind]) {
--- a/packages/teraslice-messaging/src/cluster-master/server.ts
+++ b/packages/teraslice-messaging/src/cluster-master/server.ts
@@ -74,23 +74,9 @@ export class Server extends core.Server {
         });
     }
 
-    onExecutionError(fn: (clientId: string, error?: core.ResponseError) => void): void {
-        this.on('execution:error', (msg) => {
-            fn(msg.scope, msg.error);
-        });
-    }
-
     private onConnection(exId: string, socket: SocketIO.Socket) {
         this.handleResponse(socket, 'execution:finished', (msg: core.Message) => {
             this.emit('execution:finished', {
-                scope: exId,
-                payload: {},
-                error: msg.payload.error,
-            });
-        });
-
-        this.handleResponse(socket, 'execution:error', (msg: core.Message) => {
-            this.emit('execution:error', {
                 scope: exId,
                 payload: {},
                 error: msg.payload.error,
--- a/packages/teraslice-messaging/src/cluster-master/server.ts
+++ b/packages/teraslice-messaging/src/cluster-master/server.ts
@@ -74,6 +74,12 @@ export class Server extends core.Server {
         });
     }
 
+    onExecutionError(fn: (clientId: string, error?: core.ResponseError) => void): void {
+        this.on('execution:error', (msg) => {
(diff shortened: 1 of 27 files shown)


packages/teraslice-messaging/src/cluster-master/server.ts
```diff
@@ -74,6 +74,12 @@ export class Server extends core.Server {
         });
     }
 
+    onExecutionError(fn: (clientId: string, error?: core.ResponseError) => void): void {
```
godber, line 77: During review we concluded that onExecutionError could be eliminated and we can use onExecutionFinished with an error parameter instead.

packages/teraslice/src/lib/workers/execution-controller/index.ts
```diff
@@ -402,7 +402,19 @@ export class ExecutionController {
...
+        if (eventType === 'error' && shutdownError) {
+            /// Add errors to this list as needed. Errors not in this list won't cleanup resources
+            const errorList = [
+                'index specified in reader does not exist'
+            ];
+            /// Tell cluster_master that shutdown is due to a specific error
+            /// Cleans up kubernetes resources. For native, kills processes
+            if (errorList.includes(shutdownError.message)) {
+                this.logger.error('sent request to cluster_master to cleanup job resources.');
+                this.client.sendExecutionError(shutdownError.message);
```
godber, line 415: We need an await here.

godber (changes requested):

packages/teraslice/src/lib/workers/worker/index.ts
```diff
@@ -232,7 +232,7 @@ export class Worker {
         this.slicesProcessed += 1;
     }
 
-    async shutdown(block?: boolean, event?: string, shutdownError?: Error) {
+    async shutdown(event?: string, shutdownError?: Error, block?: boolean) {
```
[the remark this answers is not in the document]
sotojn: This is because of a lint error where default params must be at the end.

packages/teraslice/src/lib/cluster/services/execution.ts
```diff
@@ -255,6 +255,10 @@ export class ExecutionService {
         }
     }
 
+    private async _DeleteFailedEx(exId: string, err?: Error) {
+        this._finishExecution(exId, err);
+    }
+
```
busma13, line 261: This function is no longer used.

packages/teraslice/src/lib/workers/execution-controller/index.ts
```diff
@@ -403,7 +403,19 @@ export class ExecutionController {
...
+    async shutdown(eventType?: string, shutdownError?: Error, block: boolean = true) {
+        if (eventType === 'error' && shutdownError) {
+            /// Add errors to this list as needed. Errors not in this list won't cleanup resources
+            const errorList = [
+                'index specified in reader does not exist'
+            ];
+            /// Tell cluster_master that shutdown is due to a specific error
+            /// Cleans up kubernetes resources. For native, kills processes
+            if (errorList.includes(shutdownError.message)) {
+                this.logger.error('sent request to cluster_master to cleanup job resources.');
```
busma13, line 415: Should this be an `error` level log? Has the error been logged already? If so maybe this could be a `warn` level log. If not we should tell them what the error was.

busma13 (changes requested):

packages/teraslice/src/lib/workers/execution-controller/index.ts
busma13: The error is logged prior to this log. I changed this to a `warn`

busma13 (approved):

godber (approved):
After the changes this looks good.

godber:
Lets bump the patch level now.  We'll release after this.

godber (approved):
approved version bumps

Issue #3604, opened 2024-05-03
Add external_ports to job automatically when prom_metrics_enabled


busma13:
For Prometheus to collect metrics from a pod the port must be exposed using the `external_ports` field. https://terascope.github.io/teraslice/docs/configuration/clustering-k8s#external-ports

If prom metrics are enabled (either on the job or in terafoundation) the external ports field should automatically be added or modified in the jobConfig to expose the `prom_metrics_port`.

PR #3643, opened 2024-06-06
Add external ports to job automatically when promMetrics are enabled

busma13:
This PR makees the following changes:
- Creates an `addExternalPortsToJobSpec()` function that will add the `external_ports` field to a job if `promMetrics` are enabled and configure it to use the correct port.
- Adds a call to `addExternalPortsToJobSpec()` with the `submit()` and `update()` function within the `JobService` class.

ref: #3604

### DIFF
--- a/packages/teraslice/src/lib/cluster/services/jobs.ts
+++ b/packages/teraslice/src/lib/cluster/services/jobs.ts
@@ -346,9 +346,9 @@ export class JobsService {
      * This ensures that the Prometheus exporter server can be scraped.
      * Check if prom_metrics_enabled is true on jobSpec or teraslice config.
      * If so, add or update external_ports property with correct port.
-     * @param {Partial<JobRecord>} jobSpec
+     * @param {Partial<JobConfig>} jobSpec
     */
-    addExternalPortsToJobSpec(jobSpec: Partial<JobRecord>) {
+    addExternalPortsToJobSpec(jobSpec: Partial<JobConfig>) {
         const {
             prom_metrics_enabled: enabledInTF,
             prom_metrics_port: tfPort
--- a/packages/teraslice/src/lib/cluster/services/jobs.ts
+++ b/packages/teraslice/src/lib/cluster/services/jobs.ts
@@ -79,6 +79,7 @@ export class JobsService {
             });
         }
 
+        this.addExternalPortsToJobSpec(jobSpec);
         const validJob = await this._validateJobSpec(jobSpec);
 
         // We don't create with the fully parsed validJob as it changes the asset names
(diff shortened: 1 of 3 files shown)


packages/teraslice/src/lib/cluster/services/jobs.ts
```diff
@@ -338,4 +340,34 @@ export class JobsService {
...
         return parsedAssetJob;
     }
+
+    /**
+     * Automatically add external_ports to jobSpec if needed.
+     * This ensures that the Prometheus exporter server can be scraped.
+     * Check if prom_metrics_enabled is true on jobSpec or teraslice config.
+     * If so, add or update external_ports property with correct port.
+     * @param {Partial<JobConfig>} jobSpec
+    */
```
godber, line 350: You should describe the precedence behavior in the docs here:

https://terascope.github.io/teraslice/docs/configuration/clustering-k8s#prometheus-metrics

Otherwise this is good.

godber (changes requested):
see comment about docs

packages/teraslice/src/lib/cluster/services/jobs.ts
busma13: `The following properties will override the matching property set within the terafoundation configuration.`

I'm guessing this sentence from the docs is not clear enough?

godber: Actually, I need to think through that logic more carefully overall.

godber: Yeah, I know why this is hard.

Technically, I don't think we have enough properties to control all the conditions an operator/user may want.  We have a single `prom_metrics_enabled` property, set either at the Terafoundation or the job level.  But we may want the ability to trigger Prometheus independently on the master or on the jobs.  That being the case what does the Terafoundation level `prom_metrics_enabled` represent?  Is it the setting for the metrics on the master or is it the default for the jobs?

I hadn't really considered this until I saw your conditional:

```javascript
        if (enabledInJob === true || (enabledInJob === undefined && enabledInTF)) {
```

and wondered what `enabledInJob === false` would imply.

I think the logic you have here is a great compromise.

Here are the conditions we hypothetically want to support:

* master metrics off, job metrics off
  * no settings anywhere
  * good
* master metrics on, job metrics off
  * `enabledInTF: True` and `enabledInJob: False` in all jobs
  * I can see this one potentially being a problem, lets say the metrics are costly somehow but we want them in the master, but not in the jobs, we will be forced to set this for `prom_metrics_enabled: False` in EVERY job to work around this.
* master metrics on, job metrics on
  * `enabledInTF: True`
  * good
* master metrics off, job metrics on
  * `enabledInTF: False` and `enabledInJob: True` in all jobs
  * this is pretty good, because the odds of not wanting master metrics is low

What I'm getting at is our properties were kind of wrong from the start, for completeness they might need to be:

* Terafoundation
  * `prom_metrics_enabled_master`
  * `prom_metrics_enabled_job_default`
* Job
  * `prom_metrics_enabled` 

But really this all seems a little fussy, unless we really need it for that second scenario above.

godber (approved):
Let's just proceed with this.

PR #3652, opened 2024-06-12
e2e TLS minio test coverage

sotojn:

- Adds e2e test coverage for TLS encrypted `minio` services
- Adds new `generate-certs.sh` script that quickly builds new certs for `minio` using mkcert
- Bumps @terascope/scripts to `v.0.78.0`


scripts/generate-cert.sh
```diff
@@ -0,0 +1,105 @@
...
+
+    # remove first arg in arg array
+    shift
+
+    for arg in "$@"; do
+        DNS_NAMES+="$arg "
+    done
+    echo "$DNS_NAMES"
+    if [ -d "$CERT_DIR" ]; then
+        rm -rf $CERT_DIR
```
godber, line 58: I think this is safe enough the way it is currently implemented, but make sure you think long and hard anytime you add an `rm -rf $var` to your scripts.  Make sure there's no way to screw up `$var`.  One thing you can do (it doesn't matter in this case) is `set -u`, to treat unset variables as errors.  I'm not asking for a change here, just warning you.

godber (approved):

Issue #3655, opened 2024-06-14
Track down which dependency is logging errors


busma13:
E2e tests are spitting out thousands of lines of errors. We believe a dependency update from PR #3608 introduced the errors.

Example:
```
  ERROR  failed to read input source map: failed to parse inline source map url
v1ReplicaSetCondition.js.map
Caused by:
    relative URL without a base
    at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/swc-0.275.1/src/lib.rs:405
  ERROR  failed to read input source map: failed to parse inline source map url
v1ReplicaSetList.js.map
Caused by:
    relative URL without a base
    at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/swc-0.275.1/src/lib.rs:405
  ERROR  failed to read input source map: failed to parse inline source map url
v1ReplicaSetSpec.js.map
Caused by:
    relative URL without a base
    at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/swc-0.275
```

busma13:
The errors are coming from `@swc/core`. They start showing up at version `1.4.12`. 

I believe this issue is referring to the problem we are having: https://github.com/swc-project/swc/issues/8789

busma13:
I am going to revert back to the previous version we were using (1.4.0) until there is a fix.

PR #3661, opened 2024-06-17
revert @swc/core to version 1.4.0

busma13:
When upgrading `@swc/core` from v1.4.0 to v1.5.5 there was a change that results in source mappings failing. This would print thousands of lines of error messages in out e2e tests, but everything seems to compile okay. To cut down this noise until there is a fix, this PR reverts the `@swc/core` version back to 1.4.0.

More details and links here: #3655

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
         "nan": "^2.19.0"
     },
     "devDependencies": {
-        "@swc/core": "^1.5.5",
+        "@swc/core": "1.4.0",
         "@swc/jest": "^0.2.36",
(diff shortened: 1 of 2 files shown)


godber:
@jsnoble I don't know what this tool is, perhaps you do, if so, you think its OK to roll this back until they address the noisy output problem?

jsnoble (approved):
LGTM

jsnoble:
swc is a rust typescript compiler that we are using in some cases rather than ts-jest.

PR #3662, opened 2024-06-18
Test harness shutdown fix

jsnoble:
- the executionContext might not be set if shutdown is called before initialize, so we check for its existance


busma13 (approved):
Looks good.

PR #3667, opened 2024-06-21
Add k8s-env --dev option

sotojn:

- Adds `--dev` option to k8s-env command
  - This flag when enabled will add the entire `teraslice` repo as a volume to allow for fast build changes when running in kubernetes. 
- Adds dev compatibility to k8s master backend
- Fixes asset-storage setting in k8s-env
  - This resolves a bug where launching k8s-env with `asset-storage` set to `s3` would not work correctly

### DIFF
--- a/packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetes/k8sResource.ts
+++ b/packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetes/k8sResource.ts
@@ -85,6 +85,9 @@ export class K8sResource {
         this._setTargets();
         this._setResources();
         this._setVolumes();
+        if (process.env.MOUNT_LOCAL_TERASLICE === 'true') {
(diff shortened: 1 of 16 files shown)


busma13 (changes requested):

packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetes/k8sResource.ts
```diff
@@ -108,6 +111,150 @@ export class K8sResource {
         }
     }
 
+    _mountLocalTeraslice(contextType: string) {
```
godber, line 114: The volumemounts you've added here ... this is like the fifth time in this PR you've listed these.  Is there no sensible way to consolidate these?  If you need to add one, you have to make changes in 5 places in three different files.

godber (approved):

PR #3670, opened 2024-06-26
teraslice-cli - fix old asset build

godber:
Removing the `format: esm` argument to esbuild lets old assets build correctly.


jsnoble (approved):
LGTM

Issue #3676, opened 2024-07-05
CI Tests Failing Due to Docker Hub Rate Limiting on Image Pulls


sotojn:
Recently we've had issues with with CI throwing an error on tests that depend on docker to pull images of services the test depends on. The problem is most likely due to the excessive amount of pulling services for each test environment. For example e2e tests will need to pull at minimum 3 images (kafka, elasticsearch, and zookeeper). This is multiplied by the matrix that will test different versions of node and then multiplied again with different versions of elasticsearch. 

Error: 
```
toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```
Link: https://github.com/terascope/teraslice/actions/runs/9784466833/job/27015404404

Quoted from docker itself:
https://www.docker.com/increase-rate-limits/
> The rate limits of 100 container image requests per six hours for anonymous usage, and 200 container image requests per six hours for free Docker accounts are now in effect. Image requests exceeding these limits will be denied until the six hour window elapses.

Will look into this more and add potential solutions soon.

busma13:
List of images that teraslice CI tests will pull:

terascope/node-base:18.19.1
terascope/node-base:20.11.1
terascope/node-base:22.2.0
elasticsearch:6.8.6
elasticsearch:7.9.3
opensearchproject/opensearch:1.3.10
opensearchproject/opensearch:2.8.0
confluentinc/cp-kafka:7.1.9
confluentinc/cp-zookeeper:7.1.9
minio/minio:RELEASE.2020-02-07T23-28-16Z

godber:
This was mostly resolved and we have added a daily cache creation job.  We have also started moving our images into ghcr.

PR #3679, opened 2024-07-08
Docker ratelimit ci

busma13:
- Add `images` cmd to scripts
  - `images list` creates a list of images that CI will need
  - `images save` will use the list to pull and save those images
- Update `test.yml` workflow to check for a docker image cache, create one if needed, then use the cache to load images on all jobs.
- Update `test-runner` to load images from cache if in CI
- Update `scripts` from 0.80.0 to 0.81.0

ref: #3676


packages/scripts/src/helpers/kind.ts
```diff
@@ -83,7 +83,19 @@ export class Kind {
...
+            let subprocess;
+            if (isCI) {
+                // In CI we load images directly from the github docker image cache
+                // Without this we run out of disk space
+                const fileName = `${serviceImage}_${version}`.replace(/[/:]/g, '_');
+                const filePath = path.join(DOCKER_CACHE_PATH, `${fileName}.tar.gz`);
+                if (!fs.existsSync(filePath)) {
+                    throw new Error(`No file found at ${filePath}. Have you restored the cache?`);
+                }
+                subprocess = await execa.command(`kind --name ${this.clusterName} load image-archive <(gunzip -c ${filePath})`);
```
godber, line 95: I don't understand how the `<(gunzip ..)` subprocess would work without the `shell=true` option.

godber (approved):
Since the one issue called out here actually works at the moment, we can fix that in the follow up clean up PR.

PR #3682, opened 2024-07-12
Further improvements to CI test pipeline

busma13:
- Run `verify-build` and `cache-docker-images` jobs in parallel in `test.yml` workflow. E2e tests now await both.
- Remove node 20 from e2e k8s jobs matrices
- Move `saveAndZip` function from `images` to `scripts`
- Add JSDoc style comments to functions
- Add tests for `images` command and `loadOrPullServiceImages` function
- Load kind image at proper time and load service images directly into kind correctly.
- Change error handling on `kindLoadImages` so errors aren't swallowed.

### DIFF
--- a/packages/scripts/src/helpers/kind.ts
+++ b/packages/scripts/src/helpers/kind.ts
@@ -82,23 +82,26 @@ export class Kind {
     async loadServiceImage(
         serviceName: string, serviceImage: string, version: string
     ): Promise<void> {
+        let subprocess;
         try {
-            let subprocess;
             if (isCI) {
                 // In CI we load images directly from the github docker image cache
-                // Without this we run out of disk space
                 const fileName = `${serviceImage}_${version}`.replace(/[/:]/g, '_');
                 const filePath = path.join(DOCKER_CACHE_PATH, `${fileName}.tar.gz`);
+                const tarPath = path.join(DOCKER_CACHE_PATH, `${fileName}.tar`);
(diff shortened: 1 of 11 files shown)


packages/scripts/src/helpers/kind.ts
```diff
@@ -82,23 +82,27 @@ export class Kind {
...
                 const filePath = path.join(DOCKER_CACHE_PATH, `${fileName}.tar.gz`);
+                const tarPath = path.join(DOCKER_CACHE_PATH, `${fileName}.tar`);
                 if (!fs.existsSync(filePath)) {
                     throw new Error(`No file found at ${filePath}. Have you restored the cache?`);
                 }
-                subprocess = await execa.command(`kind --name ${this.clusterName} load image-archive <(gunzip -c ${filePath})`);
+                subprocess = await execa.command(`gunzip -d ${filePath}`);
+                signale.info(`${subprocess.command}: successful`);
+                subprocess = await execa.command(`kind load --name ${this.clusterName} image-archive ${tarPath}`);
+                fs.rmSync(tarPath);
```
godber, line 98: Was the one step process not working?  Is that why you're going to this two step process?

godber: Oh, 

```
subprocess = await execa.command(`kind --name ${this.clusterName} load image-archive <(gunzip -c ${filePath})`);
```

Doesn't have the `shell=true`.

godber (approved):

PR #3684, opened 2024-07-12
Additional esm package changes

jsnoble:
- [x] check benchmarks before and after conversion of libraries like data-mate
- [x] convert all libraries

will refactor in another PR as that could be cascading changes.

- reverted the changes to the `eslint-config` package, the refactor requires a library update that would also force a config change which will have an impact on this codebase as well as our others as well


godber (approved):
If Jared is confident about the jexl changes ... or confident that the tests sufficiently cover those changes, I approve this PR as is.

Issue #3683, opened 2024-07-12
Unpin Node Patch Versions in CI


godber:
In cases where we are using specific patch level node versions we should switch those to major only.

So for instance 18.19.1, should become 18, or more specifically, where we currently build just a `terascope/node-base:18.19.1` image, instead, we would build a `terascope/node-base:18.19.1` image, and add the tags `terascope/node-base:18.19` and `terascope/node-base:18`.

Then we change our downstream processes to use the `terascope/node-base:18`  image.  But we ALWAYs tag all three.  This means that `terascope/node-base:18` gets over written all of the time, but IF theres problem with a minor or patch level build, we can change our downstream projects to  temporarily use the most recent version without the issue.

PR #3687, opened 2024-07-18
Unpin node versions in CI

sotojn:
- Changes node versions across `teraslice` to use major versions of node as a valid node version
  - This is because the env variable `NODE_VERSION` used in e2e is tied to a valid tag in [terascope/node-base](https://hub.docker.com/r/terascope/node-base/tags)
- Added `workflow_dispatch` to the `daily-docker-cache.yml` to allow for manual workflow trigger
- Added `check-latest-node-version.sh` script for devs to find out which node version is currently used in the base image
  - This is a convenience utility script that could be used somewhere else later to pop up a warning.
- Bumped ` @terascope/scripts` from `v0.81.2` to `v0.81.3` 
- Updated docs to explain how tagging works in the base image repo

Ref to issue #3683

### DIFF
--- a/docs/development/node.md
+++ b/docs/development/node.md
@@ -17,7 +17,7 @@ https://github.com/terascope/base-docker-image
 
 The workflow for the base image tags is closely linked to the Node.js version used in the image. Here's a simple breakdown of how it works:
 
-**Major Version Tag:** The base image will either grab the latest available version of a specific major Node.js release from the node alpine image(e.g., Node 18) or it will be pinned to the latest node version thats compatible with the base image. This image is tagged with the major version number (e.g., 18). So in some cases this version will be pinned and not completley up to date with a node release. This tag is always overwritten on release.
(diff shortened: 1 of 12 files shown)


docs/development/node.md
```diff
@@ -13,6 +13,15 @@ Start by modifying the CI YAML files in the base-docker-image repository.
 
 https://github.com/terascope/base-docker-image
 
+### Note on how tags and node versions work in the base image
+
+The workflow for the base image tags is closely linked to the Node.js version used in the image. Here's a simple breakdown of how it works:
+
+**Major Version Tag:** The base image will either grab the latest available version of a specific major Node.js release from the node alpine image(e.g., Node 18) or it will be pinned to the latest node version thats compatible with the base image. This image is tagged with the major version number (e.g., 18). So in some cases this version will be pinned and not completley up to date with a node release. This tag is always overwritten on release.
```
godber, line 20:
```suggestion
**Major Version Tag:** The base image will either grab the latest available version of a specific major Node.js release from the node alpine image(e.g., Node 18) or it will be pinned to the latest node version that is compatible with the base image. This image is tagged with the major version number (e.g., 18). So in some cases this version will be pinned and not completely up to date with a node release. This tag is always overwritten on release.
```

godber (approved):

PR #3689, opened 2024-07-23
make extract schema async

lesleydreyer:
Makes the function that gets the schema async  (i.e. awaiting imports of external plugins to add to the schema)


jsnoble (approved):
LGTM

PR #3690, opened 2024-07-23
Update 'images list' cmd to use config variables, not hardcoded values

busma13:
- Add several new variables to `config.ts`: 
  - KIND_DOCKER_IMAGE
  - KIND_VERSION
  - BASE_DOCKER_IMAGE
  - TEST_NODE_VERSIONS
  - DEFAULT_NODE_VERSION
  - __DEFAULT_ELASTICSEARCH6_VERSION
  - __DEFAULT_ELASTICSEARCH7_VERSION
  - __DEFAULT_OPENSEARCH1_VERSION
  - __DEFAULT_OPENSEARCH2_VERSION
- `images list` command creates the image tags and versions based off the values in `config.ts` instead of hard-coded values.
- Replace other hard-coded values where necessary
- Remove version flags from test-runner and k8s-env commands and get all versions from `config.ts`


packages/scripts/src/helpers/test-runner/utils.ts
```diff
@@ -132,7 +132,7 @@ export function getEnv(options: TestOptions, suite: string): ExecEnv {
             OPENSEARCH_USER: config.OPENSEARCH_USER,
             OPENSEARCH_PASSWORD: config.OPENSEARCH_PASSWORD,
             OPENSEARCH_PORT: config.OPENSEARCH_PORT,
-            OPENSEARCH_VERSION: config.OPENSEARCH_VERSION,
+            OPENSEARCH_VERSION: options.opensearchVersion,
```
busma13, line 135: @jsnoble Can you confirm that the opensearch version should be coming from the test options and not the config.ts file?

godber (approved):
I am going to approve this PR but I REALLY dislike the use of environment variables and would like to push for us to STOP using them except were absolutely necessary (like passing stuff to jest).  The reason I dislike it is that they can be abused and used as global variables.  This PR does not abuse that and all of the uses are confined to config.ts and then properties on config.  This is OK.  But if people start using `process.env` to access these values all over the place we will start shooting ourselves in the foot.

PR #3699, opened 2024-07-24
Migrate teraslice image from docker.hub to ghcr.io registry

sotojn:

- Moves `teraslice` images to be published in the `ghcr.io` registry instead of `docker.hub`
- Teraslice images are now built off the `node-base` image in the `ghcr.io` registry 
- Unpins node 22

### DIFF
--- a/packages/scripts/src/helpers/scripts.ts
+++ b/packages/scripts/src/helpers/scripts.ts
@@ -209,18 +209,15 @@ export async function dockerTag(from: string, to: string): Promise<void> {
 }
 
 export async function getNodeVersionFromImage(image: string): Promise<string> {
-    let nodeVersion:string;
     try {
         const { stdout } = await execa(
             'docker',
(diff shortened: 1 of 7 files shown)


packages/scripts/src/helpers/scripts.ts
```diff
@@ -208,6 +208,21 @@ export async function dockerTag(from: string, to: string): Promise<void> {
...
 }
 
+export async function getNodeVersionFromImage(image: string): Promise<string> {
+    let nodeVersion:string;
+    try {
+        const { stdout } = await execa(
+            'docker',
+            ['run', image, 'node', '-v']
+        );
+        nodeVersion = stdout;
```
busma13, line 218: Maybe we can return stdout directly?

busma13 (changes requested):

busma13 (approved):
LGTM

godber (approved):

PR #3701, opened 2024-07-24
Fix node version check in k8s --dev mode

sotojn:

- Fixes bug when running `ts-scripts k8s --dev` that will fail when env variable `NODE_VERSION` is a major version
  - Before, this check only expected full semver versions
- Added kind cleanup in error scenarios
- Bumps **@terascope/scripts** from `v0.83.0` to `v0.83.1`

### DIFF
--- a/packages/scripts/src/helpers/k8s-env/index.ts
+++ b/packages/scripts/src/helpers/k8s-env/index.ts
@@ -78,11 +78,14 @@ export async function launchK8sEnv(options: K8sEnvOptions) {
             const { stdout } = await execa('docker', ['run', e2eImage, 'node', '-v']);
             imageVersion = stdout;
         } catch (err) {
+            await kind.destroyCluster();
             throw new Error(`Problem running docker command to check node version: ${err}`);
         }
         if (process.version !== imageVersion) {
-            throw new Error(`The node version this process is running on (${process.version}) does not match
+            signale.fatal(`The node version this process is running on (${process.version}) does not match
             the version set in k8s-env image (${imageVersion}). Check your version by running "node -v"`);
+            await kind.destroyCluster();
+            process.exit(1);
         }
         signale.info(`Running yarn setup with node ${process.version}...`);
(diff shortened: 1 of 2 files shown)


packages/scripts/src/helpers/k8s-env/index.ts
```diff
@@ -39,9 +39,17 @@ export async function launchK8sEnv(options: K8sEnvOptions) {
     // If --dev is true, we must run yarn setup before creating resources
     // We need a local node_modules folder built to add it as a volume
     if (options.dev) {
-        if (process.version.substring(1) !== options.nodeVersion) {
+        let imageVersion: string;
+        const baseImage = `terascope/node-base:${config.NODE_VERSION}`;
```
busma13, line 43: i know the teraslice image we build should have the same version as the base image, but should we be checking the version we build further down in this function, just to be safe?

packages/scripts/src/helpers/k8s-env/index.ts
```diff
@@ -87,6 +70,33 @@ export async function launchK8sEnv(options: K8sEnvOptions) {
...
+    // If --dev is true, we must run yarn setup before creating resources
+    // We need a local node_modules folder built to add it as a volume
+    if (options.dev) {
+        let imageVersion:string;
+        try {
+            const { stdout } = await execa('docker', ['run', e2eImage, 'node', '-v']);
+            imageVersion = stdout;
+        } catch (err) {
+            await kind.destroyCluster();
+            throw new Error(`Problem running docker command to check node version: ${err}`);
```
busma13, line 82: Can we follow the same pattern as below and use `signale.fatal` to show this error, then `process.exit(1)`?

sotojn: This is because when I throw an error it shows the trace stack, which in this case is helpful because it's catching the failure of a docker command run

godber (approved):

PR #3702, opened 2024-07-24
Create repo specific images lists

busma13:
- Add ability to get the repo name out of the root `package.json`. This allows other repos that use ts-scripts to get a custom docker image list to use to create a cache specific to that repo. Valid repos are `elasticsearch-assets`, `kafka-asset-bundle`, `file-asssets-bundle`, and `teraslice-workspace`.
- Update tests 
- Bump scripts from version 0.81.4 to 0.82.0

### DIFF
--- a/packages/scripts/src/helpers/images/index.ts
+++ b/packages/scripts/src/helpers/images/index.ts
@@ -3,10 +3,11 @@ import * as config from '../config';
 import { ImagesAction } from './interfaces';
 import signale from '../signale';
 import { dockerPull, saveAndZip } from '../scripts';
+import { getRootInfo } from '../misc';
 
-export async function images(action: ImagesAction, repo: string): Promise<void> {
+export async function images(action: ImagesAction): Promise<void> {
     if (action === ImagesAction.List) {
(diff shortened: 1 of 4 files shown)


packages/scripts/src/helpers/images/index.ts
```diff
@@ -33,6 +46,9 @@ export async function createImageList(): Promise<void> {
                + `${config.ZOOKEEPER_DOCKER_IMAGE}:${config.KAFKA_IMAGE_VERSION}\n`
                + `${config.MINIO_DOCKER_IMAGE}:${config.MINIO_VERSION}\n`
                + `${config.KIND_DOCKER_IMAGE}:${config.KIND_VERSION}`;
+    } else {
+        throw new Error(`Repo ${repo} is not supported.`);
```
godber, line 50: This error is written from the devloper's point of view, not the user or operator.  It should be telling me where that `name` in the package json is not supported.  Mentioning repo is fine but users have no context for understanding it.  Accepting anyway.

godber (approved):

PR #3703, opened 2024-07-25
rework generateTSConfig

lesleydreyer:
The generateTSConfig writes to commonjs.

Ideas were to... 
A) check the package.json and adjust the compilerOptions to module/target/resolution to nodenext/esnext if it's a module
B) remove generateTSConfig completely
C) keep some options 

We looked at the git history and weren't completely clear why generating a tsconfig is needed since we already do what it's doing in the tsconfigs, so decided maybe keeping the references and just allow the original options for the rest.

### DIFF
--- a/packages/scripts/package.json
+++ b/packages/scripts/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@terascope/scripts",
     "displayName": "Scripts",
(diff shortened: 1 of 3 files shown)


jsnoble (approved):
LGTM

PR #3707, opened 2024-07-30
Update workflows and kafka test version

busma13:
- Update default kafka version in `ts-scripts` from 3.1 to 3.5
- Update `daily-docker-cache.yml` workflow to use reusable workflow from `terascope/workflows`
- Update `test.yml` workflow:
  - use reusable workflows from `terascope/workflows`
  - remove rate limit checks for each job
  - pass node versions into each job from top level variables. This uses the `compute` job to put them in scope for the matrix.
  - Bump `ts-scripts` from 0.83.1 to 0.83.2


jsnoble (approved):
LGTM

Issue #3694, opened 2024-07-23
Add earl tjm diff


godber:
It would be excellent if there was an `earl tjm diff file.json` command.  It would show the difference between the registered job on the cluster and the file.

godber:
There's an argument that this could be a `--diff` option on the `view` subcommand.

PR #3710, opened 2024-08-05
Add --diff option teraslice-cli

sotojn:

- Adds `--diff` option to `teraslice cli view` that will compare the local `json` job file with what teraslice has on the state cluster
- Bumps `teraslice-cli` from `v2.1.0` to `v2.2.0`

Ref to issue #3694

### DIFF
--- a/packages/teraslice-cli/src/helpers/jobs.ts
+++ b/packages/teraslice-cli/src/helpers/jobs.ts
@@ -696,27 +696,39 @@ export default class Jobs {
         return localJobConfigs;
     }
 
-    printDiff(diffResult: Diff.Change[]) {
+    printDiff(diffResult: Diff.Change[], showUpdateField: boolean) {
         diffResult.forEach((part) => {
             let color: chalk.Chalk;
             let symbol: string;
+            let pointer: string;
             if (part.added) {
                 color = chalk.green;
                 symbol = '+';
+                pointer = '   <--- local job file value';
             } else if (part.removed) {
                 color = chalk.red;
                 symbol = '-';
+                pointer = '   <--- state cluster value';
             } else {
                 color = chalk.grey;
                 symbol = ' ';
+                pointer = '';
             }
             const lines = part.value.split('\n');
             lines.forEach((line) => {
-                /// These fields aren't in the job file so don't compare in diff
-                if (line.includes('"_created":') || line.includes('"_context":')) {
-                    process.stdout.write(chalk.grey(` ${line}\n`));
-                } else if (line.length !== 0) {
-                    process.stdout.write(color(`${symbol} ${line}\n`));
+                /// Don't print blank lines
+                if (line.length !== 0) {
+                    /// These fields aren't in the job file so don't compare in diff
+                    if (!line.includes('"_created":') && !line.includes('"_context":')) {
+                        /// Check to see if we want to display _updated field
+                        if (line.includes('"_updated":')) {
+                            if (showUpdateField) {
+                                process.stdout.write(color(`${symbol} ${line}${pointer}\n`));
+                            }
+                        } else {
+                            process.stdout.write(color(`${symbol} ${line}${pointer}\n`));
+                        }
+                    }
                 }
             });
         });
@@ -728,7 +740,18 @@ export default class Jobs {
             this.config.args.jobFile
         );
         const diffObject = diff.diffJson(job.config, localJobConfigs[job.id]);
-        this.printDiff(diffObject);
+
+        /// "_update" fields on the job file are always off by a couple milliseconds
+        /// We only want to display a diff of this field if it's greater than a minute
+        let showUpdateField = false;
+        const jobConfigUpdateTime = new Date(job.config._updated).getTime();
+        const localConfigUpdateTime = new Date(localJobConfigs[job.id]._updated).getTime();
+        const timeDiff = Math.abs(localConfigUpdateTime - jobConfigUpdateTime);
+        if (timeDiff > (1000 * 60)) {
+            showUpdateField = true;
+        }
+
+        this.printDiff(diffObject, showUpdateField);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


packages/teraslice-cli/src/helpers/jobs.ts
```diff
@@ -669,6 +672,65 @@ export default class Jobs {
...
+        return finalJobConfig;
+    }
+
+    getLocalJSONConfigs(srcDir: string, files: string[]) {
+        const localJobConfigs = {};
+        for (const file of files) {
+            const filePath = path.join(srcDir, file);
+            const jobConfig: JobConfigFile = JSON.parse(fs.readFileSync(filePath, { encoding: 'utf-8' }));
+            const formatedJobConfig = this.formatJobConfig(jobConfig);
+            localJobConfigs[formatedJobConfig.job_id as string] = formatedJobConfig;
```
busma13, line 694:
```suggestion
            const formattedJobConfig = this.formatJobConfig(jobConfig);
            localJobConfigs[formatedJobConfig.job_id as string] = formattedJobConfig;
```

packages/teraslice-cli/src/helpers/yargs-options.ts
```diff
@@ -228,6 +228,11 @@ export default class Options {
             describe: 'Silence non-error logging.',
             type: 'boolean'
         }),
+        diff: () => ({
+            describe: 'Shows diff on a job in a cluster and a job file',
```
busma13, line 232: Maybe `Shows diff between a job...`?

busma13 (changes requested):

busma13 (approved):
Looks good

sotojn:
Here is an example on how this option looks. In the json file I change the amount of workers from 1 to 2 and run:

`earl tjm view test-job.json --diff`

Output:
```diff
  {
   "_context": "job",
   "_created": "2024-08-01T22:34:50.219Z",
-   "_updated": "2024-08-06T16:03:58.734Z",
+   "_updated": "2024-08-06T16:03:58.777Z",
    "assets": [
      "standard",
      "file"
    ],
    "job_id": "48e59745-ed75-4060-998e-002262abefc3",
    "lifecycle": "once",
    "name": "data-to-s3",
    "operations": [
      {
        "_op": "data_generator",
        "size": 100000
      },
      {
        "_op": "s3_exporter",
        "format": "ldjson",
        "path": "josephorigin"
      }
    ],
-   "workers": 1
+   "workers": 2
  }
```
```
--------
job: data-to-s3
id: 48e59745-ed75-4060-998e-002262abefc3
cluster: http://localhost:5678
--------
```
It does appear that updated seems slightly off all the time and maybe we shouldn't compare it?

busma13:
> It does appear that updated seems slightly off all the time and maybe we shouldn't compare it?

I think it could be helpful to know that `_update` will be different between the two records.

sotojn:
> > It does appear that updated seems slightly off all the time and maybe we shouldn't compare it?
> 
> I think it could be helpful to know that `_update` will be different between the two records.

Should this be some sort of warning like, "_update field is subject to being slightly different from job file, to whats on the job record in the state cluster". Or should I just ignore this comparison entirely?

busma13:
My feeling is that it's fine how it is now and a warning isn't necessary. But maybe it would be better to not compare `_update` if only the milliseconds are different.

ciorg:
If there is no difference could it out put something like - no difference between the configs instead of printing the whole config, not a big deal it's a pain to do so?   

Also, what color applies to what config? Green is the local file and red is the cluster config?  Maybe something to let people know what applies to what.

So far as the `update` difference, which for some reason is always different than the cluster value, at least on all the jobs I've looked at so far.   ( I'm not sure how the update is calculated, but it seems like that value should be the same for both?  Or maybe it shouldn't  even have millisecond precision?  But that is a different issue)   

It is interesting to know when the job was last updated, but if the diff is 100 ms it doesn't really tell me anything useful.  So maybe if there's a way to hide that diff unless it's at least a few min difference that would be better.

sotojn:
I've updated the behavior to include what is being compared to what. I also added logic around the "_updated" field to only display a diff if the time difference is greater than a minute. 

Results if "_updated field is less than 1 minute":
Command: `earl tjm view  data-to-s3.json  --diff`
```diff
  {
    "assets": [
      "standard",
      "file"
    ],
    "job_id": "48e59745-ed75-4060-998e-002262abefc3",
    "lifecycle": "once",
    "name": "data-to-s3",
    "operations": [
      {
        "_op": "data_generator",
        "size": 100000
      },
      {
        "_op": "s3_exporter",
        "format": "ldjson",
        "path": "josephorigin"
      }
    ],
-   "workers": 1   <--- state cluster value
+   "workers": 2   <--- local job file value
  }
```
```
--------
job: data-to-s3
id: 48e59745-ed75-4060-998e-002262abefc3
cluster: http://domain123
--------
```

Results if "_updated field is greater than 1 minute":
Command: `earl tjm view  data-to-s3.json  --diff`
```diff
  {
-   "_updated": "2024-08-06T16:03:58.734Z",   <--- state cluster value
+   "_updated": "2024-07-06T16:03:58.777Z",   <--- local job file value
    "assets": [
      "standard",
      "file"
    ],
    "job_id": "48e59745-ed75-4060-998e-002262abefc3",
    "lifecycle": "once",
    "name": "data-to-s3",
    "operations": [
      {
        "_op": "data_generator",
        "size": 100000
      },
      {
        "_op": "s3_exporter",
        "format": "ldjson",
        "path": "josephorigin"
      }
    ],
-   "workers": 1   <--- state cluster value
+   "workers": 2   <--- local job file value
  }
```
```
--------
job: data-to-s3
id: 48e59745-ed75-4060-998e-002262abefc3
cluster: http://domain123
--------
```

ciorg (approved):
LGTM

busma13 (approved):
Looks good.

PR #3712, opened 2024-08-06
remove docker images after saving zip file

busma13:
- Add `dockerImageRm` function to helper scripts
- Call `dockerImageRm` within the `saveAndZip` function
- Bump scripts from 0.83.2 to 0.83.3

The CI runner was running out of storage space while attempting to save the images to cache.

### DIFF
--- a/packages/scripts/src/helpers/images/index.ts
+++ b/packages/scripts/src/helpers/images/index.ts
@@ -2,7 +2,7 @@ import fse from 'fs-extra';
 import * as config from '../config';
 import { ImagesAction } from './interfaces';
 import signale from '../signale';
-import { dockerPull, saveAndZip } from '../scripts';
+import { dockerImageRm, dockerPull, saveAndZip } from '../scripts';
 import { getRootInfo } from '../misc';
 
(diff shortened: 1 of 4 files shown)


sotojn (approved):
Looks good

packages/scripts/src/helpers/images/index.ts
```diff
@@ -81,8 +82,11 @@ export async function saveImages(): Promise<void> {
                     saveAndZip(imagesArray[i], config.DOCKER_CACHE_PATH),
                     saveAndZip(imagesArray[i + 1], config.DOCKER_CACHE_PATH)
                 ]);
+                await dockerImageRm(imagesArray[i]);
+                await dockerImageRm(imagesArray[i + 1]);
```
jsnoble, line 86: these should be run in parallell, not one after the other, wrap this in a promise.all

jsnoble: better yet, it should just cleanup after its downloaded and be within the save and zip, it should be encapsulated

jsnoble (approved):
LGTM

Issue #3720, opened 2024-08-14
Add readiness probe to execution pods 


sotojn:
When creating a new execution controller job resource in kubernetes, we check to see if the execution controller pod is operational by looking at the pod status. If the pod status is in a `running` state, we proceed to create the `worker` deployment. 

The issue is that checking to see if the pod is in a `running` state is not a good indication that the execution service is ready to have workers connect and start distributing slices.

Adding a `readiness probe` to the container and checking to see if the readiness probe currently is returning `true` would solve this.

I'm looking to see if there is already a built in health check on the `execution controller` to see what we can use to verify it's `ready`.

PR #3721, opened 2024-08-16
Add readiness probe to k8s job in KubernetesV2 backend

sotojn:

- Adds readiness probe to `execution controller` job (_**Only applies to k8s V2 backend**_)
  - Worker deployments will now wait for the ex readiness probe to come back as `ready` before creation.
- Adds `/ready` endpoint to execution controller server 
  - Will return with status code `200` when initializing is complete 
  - Will return with status code `503` when initializing either fails or hasn't yet completed 
- Bumps **@terascope/teraslice-messaging** from `v1.2.0` to `v1.3.0`

Ref to issue #3720

### DIFF
--- a/packages/teraslice-messaging/src/execution-controller/server.ts
+++ b/packages/teraslice-messaging/src/execution-controller/server.ts
@@ -1,5 +1,4 @@
 import { isNumber, get, Queue } from '@terascope/utils';
-import http from 'node:http';
 import { SliceCompletePayload, EnqueuedWorker, Slice } from '@terascope/types';
 import * as core from '../messenger/index.js';
 import * as i from './interfaces.js';
@@ -13,7 +12,8 @@ export class Server extends core.Server {
 
     constructor(opts: i.ServerOptions) {
         const {
-            port, actionTimeout, networkLatencyBuffer, workerDisconnectTimeout, logger
+            port, actionTimeout, networkLatencyBuffer,
+            workerDisconnectTimeout, logger, requestListener
         } = opts;
 
         if (!isNumber(workerDisconnectTimeout)) {
@@ -23,6 +23,7 @@ export class Server extends core.Server {
         super({
             port,
             actionTimeout,
+            requestListener,
             networkLatencyBuffer,
             clientDisconnectTimeout: workerDisconnectTimeout,
             serverName: 'ExecutionController',
@@ -32,9 +33,6 @@ export class Server extends core.Server {
         this.queue = new Queue();
         this._activeWorkers = {};
         this.executionReady = false;
-
-        this.httpServer.removeListener('request', core.defaultRequestListener);
-        this.httpServer.on('request', this.requestListener);
     }
 
     async start(): Promise<void> {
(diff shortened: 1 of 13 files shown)


packages/teraslice-messaging/src/execution-controller/server.ts
```diff
@@ -21,14 +23,16 @@ export class Server extends core.Server {
...
             networkLatencyBuffer,
             clientDisconnectTimeout: workerDisconnectTimeout,
             serverName: 'ExecutionController',
-            logger,
+            logger
         });
 
         this.queue = new Queue();
         this._activeWorkers = {};
+        this.executionReady = false;
```
godber, line 35: This is a change that will impact all backend types and could be problematic.  I find it strange that this wasn't set before, how is this used in other `native` and `kubernetes` backend types?

packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetes/index.ts
```diff
@@ -162,6 +163,14 @@ export class KubernetesClusterBackend {
         // @ts-expect-error
         execution.k8sUid = jobs.items[0].metadata.uid;
 
+        /// Wait for ex readiness probe to return 'Ready'
```
godber, line 166: Why are we changing the `kubernetes` backend?

busma13 (approved):
This looks good

godber (changes requested):
Don't change the original `kubernetes` backend.

packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetesV2/jobs/execution_controller.hbs
```diff
@@ -30,6 +30,17 @@
...
                     "name": "{{name}}",
                     "ports": [{ "containerPort": 45680 }],
+                    "readinessProbe": {
+                        "httpGet": {
+                            "path": "/health",
+                            "port": 45680
+                        },
+                        "initialDelaySeconds": 5,
+                        "periodSeconds": 10,
+                        "timeoutSeconds": 5,
```
godber, line 40:
```suggestion
                        "initialDelaySeconds": 3,
                        "periodSeconds": 10,
                        "timeoutSeconds": 1,
```

I don't think we should extend `timeoutSeconds` from the default.  I've shortened `initialDelaySeconds`, if that doesn't work for some reason, you can put it back.  We can leave `periodSeconds` as it is.

godber: Was there any reason you made `timeoutSeconds` longer?  Did you experiment with a shorter `initialDelaySeconds` at all?

sotojn: No I have not. It was more of ensuring that the probe doesn't fail after the retry amount exceeded. But maybe I could have increased the `failureThreshold` instead of adding to the delay. Giving the probe an opportunity to return a Ready status a lot sooner.

godber (approved):

PR #3722, opened 2024-08-19
Include fix deps script in npm package

busma13:
This PR adds `scripts/fix-deps.js` to the `files` array within `xlucene-parser/package.json` so that the script is included in the NPM package.
Bumps xlucene-parser from 1.0.0 to 1.0.1


jsnoble (approved):
LGTM

PR #3726, opened 2024-08-22
Remove KAFKA_VERSION in e2e package json scripts

sotojn:

- Removes `KAFKA_VERSION` in all of the `package.json` scripts in e2e
  - This fixes issues with caching and allows all test using `kafka` to run on the same default version `3.5`


busma13 (approved):
LGTM

Issue #3725, opened 2024-08-21
Earl assets build on standard assets fails


busma13:
After updating terascope packages within standard assets to ESM versions I get the following error:
`Bundled asset failed to require: TypeError: Invalid property descriptor. Cannot both specify accessors and a value or writable attribute, #<Object>`

It looks like within `packages/data-mate/src/index.ts` the cloning of `FieldTransform` and `RecordTransform` fail.

Steps to reproduce: 
- pull master branch of standard-assets (currently 1.0.0)
- run `yarn upgrade-interactive --latest` and update all teraslice libraries to their ESM versions
- run `yarn setup`
- run `teraslice-cli assets build`

busma13:
`cloneDeep` calls this `object` function:
```
object(input: any): any {
        const descriptors = Object.getOwnPropertyDescriptors(input);
        // eslint-disable-next-line guard-for-in
        for (const key in descriptors) {
            descriptors[key].value = cloneDeep(descriptors[key].value);
        }
        return Object.create(
            Object.getPrototypeOf(input),
            descriptors
        );
    }
```

`input.key` before and after the recursive deepClone call within the for loop:
```
@@@@ before:  {
  get: [Function: decodeBase64],
  set: undefined,
  enumerable: true,
  configurable: false
}
@@@@ after:  {
  get: [Function: decodeBase64],
  set: undefined,
  enumerable: true,
  configurable: false,
  value: undefined
}
```

busma13:
ESBuild creates this export function that is directly defining a getter:
```
var __export = (target, all) => {
  for (var name in all)
    __defProp(target, name, { get: all[name], enumerable: true });
};
```

busma13:
The simplest way to fix this was to not use `cloneDeep` and instead copy over all properties of the imported objects.

godber:
The MR above resolved this problem.

PR #3727, opened 2024-08-22
Fix data mate import copy

busma13:
- Modify `packages/data-mate/src/index.ts` to copy each property from the `FTransform` and `RTransform` imports to a new object instead of using `cloneDeep`. This avoids a type error that occurs when `teraslice-cli assets build` tries to build an ESM asset that imports `data-mate` 
- Bump data-mate from 1.0.2 to 1.0.3, elasticsearch-store from 1.0.2 to 1.0.3, terafoundation from 1.2.2 to 1.2.3, ts-transforms from 1.0.2 to 1.0.3 

See here for details: #3725

### DIFF
--- a/packages/data-mate/src/index.ts
+++ b/packages/data-mate/src/index.ts
@@ -11,7 +11,7 @@ import { AggregationFrame } from './aggregation-frame/AggregationFrame.js';
 import { FieldTransformInterface, RecordTransformInterface } from './interfaces.js';
 
 // import are immutable, so we rename and clone to alter methods
-const tmpFieldTransform = { };
+const tmpFieldTransform = {};
 for (const key of Object.keys(FTransform)) {
     tmpFieldTransform[key] = FTransform[key];
 }
--- a/packages/data-mate/src/index.ts
+++ b/packages/data-mate/src/index.ts
@@ -12,20 +12,22 @@ import { AggregationFrame } from './aggregation-frame/AggregationFrame.js';
 import { FieldTransformInterface, RecordTransformInterface } from './interfaces.js';
 
 // import are immutable, so we rename and clone to alter methods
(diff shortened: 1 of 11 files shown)


packages/data-mate/src/index.ts
```diff
@@ -9,16 +8,24 @@ import {
...
 import { AggregationFrame } from './aggregation-frame/AggregationFrame.js';
+import { FieldTransformInterface, RecordTransformInterface } from './interfaces.js';
 
 // import are immutable, so we rename and clone to alter methods
-const FieldTransform = cloneDeep(FTransform);
-const RecordTransform = cloneDeep(RTransform);
-
-FieldTransform.repository.extract = extractConfig;
-FieldTransform.extract = extract;
+const tmpFieldTransform = {};
```
jsnoble, line 14: You are creating an empty object and trying to force a type on it, hence all the type additions which would be burdensome to mantain or know to add to it when we add another function.  

This should be a simple
 ```
const FieldTransforms = {
     ...FTransfom
}
```
This copies the types over for you and still maintains the types

File: packages/data-mate/src/interfaces.ts [hunk 2]
```diff
@@ -35,3 +36,55 @@ export interface ExtractFieldConfig {
     start?: any;
     end?: any;
 }
+
+export interface FieldTransformInterface {
+    decodeBase64: typeof FieldTransform.decodeBase64;
+    decodeHex: typeof FieldTransform.decodeHex;
+    decodeURL: typeof FieldTransform.decodeURL;
+    encodeBase64: typeof FieldTransform.encodeBase64;
+    encodeHex: typeof FieldTransform.encodeHex;
```
jsnoble, line 45: this should not be needed

jsnoble (changes requested):

jsnoble (approved):
LGTM

PR #3735, opened 2024-08-28
Add yarn to run jest command

busma13:
- Changes the `runJest` script to call `yarn` with `jest` as the first argument rather than calling `jest` on its own. This allows the test runner to run tests in a package using `yarn3 PnP` and ESM modules without getting module not found errors.
- Bump scripts from 1.0.0 to 1.0.1

### DIFF
--- a/packages/scripts/package.json
+++ b/packages/scripts/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@terascope/scripts",
     "displayName": "Scripts",
-    "version": "1.0.0",
+    "version": "1.0.1",
     "description": "A collection of terascope monorepo scripts",
     "homepage": "https://github.com/terascope/teraslice/tree/master/packages/scripts#readme",
     "bugs": {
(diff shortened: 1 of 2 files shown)


godber:
Has this solution been tested and verified to work in the other use cases?  Here's what I can think of:

* Teraslice
* Asset
* Spaces - (directly addressed by this change)

busma13:
It has been tested with teraslice, spaces, and elasticsearch-assets. I'm about to test it on the other assets, but I see no reason that it wouldn't work with any of them.

busma13:
I tested all assets from the root directory and also from within the packages. The only one that didn't work is `standard-asset-apis`, which does not use `ts-scripts`, it calls `jest` directly. Modifying the command there to use `ts-scripts` might fix the issues the tests are having currently.

busma13:
Yes, using ts-scripts in `standard-asset-apis` fixes the failures with the current setup

godber:
> Modifying the command there to use ts-scripts might fix the issues the tests are having currently.

Sounds like a future todo

sotojn (approved):
I've tested this in spaces, es-assets, and file-assets. They all work and looks good.

PR #3741, opened 2024-09-06
[e2e] Prevent the dockerUp function from failing by retrying when node version check returns undefined

busma13:
- Wrap node version check in a pRetry and throw if undefined. Occasionally e2e tests will fail because using docker compose to exec into the teraslice master and checking the node version will return undefined. It is most likely a timing issue, so we will now attempt this 3 times and then throw if we still get undefined.
- Bump scripts from 0.7.0 to 0.7.1

### DIFF
--- a/e2e/package.json
+++ b/e2e/package.json
@@ -1,7 +1,7 @@
 {
     "name": "e2e",
     "displayName": "E2E Tests",
-    "version": "0.7.0",
+    "version": "0.7.1",
     "private": true,
     "description": "Teraslice integration test suite",
     "keywords": [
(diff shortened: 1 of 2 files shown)


sotojn:
Are you saying this happens because `runCmd()` executes the command and the command fails, leaving `undefined` as the result?

sotojn (approved):
Overall this is a good fix in the event docker fails to execute at that moment.

busma13:
> Are you saying this happens because `runCmd()` executes the command and the command fails, leaving `undefined` as the result?

I have not been able to reproduce locally. All I know is that `compose.runCmd` does not error and receives `undefined` as stdout to `docker compose exec teraslice-master node --version`. My guess is that `node --version` fails for some unknown reason.

Issue #3750, opened 2024-09-12
Force stop command deletes all jobs if the kubernetes job no longer exists


busma13:
When force deleting a job, there is a possibility that the kubernetes job resource has been deleted, while there are other resources (pods, services, deployments) that have been orphaned. In this case `K8s._deleteObjByExId()` will receive `undefined` as the `name` of the job resource. The `K8s.delete()` function uses `name` to select which jobs to delete. If passed `undefined` or an empty string it will return ALL job resources and then delete them. 

`K8s.delete()` needs to ensure that `name` is defined and not an empty string.
`K8s._deleteObjByExId()` must ensure there is a resource to delete before calling `delete()`.

This bug is only in kubernetes clustering, as kubernetesV2 receives a typed object when looking up the job resource, and therefore already checks whether name is defined.

godber:
I was just thinking about this and I am pretty sure the steps to reproduce are easily achieved:

1. start two jobs, `job1` with a long `delay` and `job2` that will be accidentally stopped
2. call `/_stop` on `job1`
3. immediately call `/_stop?force` on `job1`
4. check state of jobs, `job1` and `job2` should be gone

The first call to stop deletes the all of the resources related to that `exId`, but leaves the pods in terminating due to their grace period.  Then the second call to `_stop` uses `force`, has no Kubernetes Job for the execution controller, leading to the `undefined` being sent to the k8s API delete calls.

busma13:
While resolving this issue I ran into some other problems:
- The force option was only deleting jobs and pods. It now deletes services, deployments, and replicasets. This will let us force delete any of these resources if K8s has a strange failure that leaves them up.
- Force stopping after a regular stop is not possible in K8s. K8s marks resources for deletion, adds the `metadata.deletionGracePeriodSeconds` and `metadata.deletionTimestamp` fields, and will not send a SIGKILL until the grace period has concluded. `metadata.deletionGracePeriodSeconds` cannot be modified or removed, so a second call to delete with `gracePeriodSeconds = 0` (equivalent of kubectl delete --force) will not be able to update the grace period and a SIGKILL will not be sent immediately. For this reason the order that we force delete resources is important. We must delete child resources before parents or they will be deleted in the background with the normal grace period. It may also be possible to change the `propogationPolicy` from `background` to `orphan` or `foreground`, but that would require other changes.

busma13:
Useful links:
- https://kubernetes.io/docs/reference/kubectl/generated/kubectl_delete/
[...]

PR #3751, opened 2024-09-12
[teraslice] fix multi-job deletion bug in force job stop

busma13:
- Ensure that `K8s._deleteObjByExId()` contains an object in `objList.items[]` before calling `K8s.deleted()`.
- Throw if `K8s.delete()` receives a `name` argument that is undefined or an empty string.
- Refactor `K8s._deleteObjByExId()` to delete multiple objects and not handle `force logic`
- Refactor `k8s.deleteExecution() to handle `force` logic by calling `K8s._deleteObjByExId()` on multiple resources.
- bump teraslice from version 2.3.1 to 2.3.2
- pin the kind image used by k8s to: `kindest/node:v1.28.12@sha256:fa0e48b1e83bb8688a5724aa7eebffbd6337abd7909ad089a2700bf08c30c6ea`. Note that the version in the image tag refers to the kubernetes server version that will be used. We should keep this in sync with prod.

ref: #3750

### DIFF
--- a/packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetesV2/k8s.ts
+++ b/packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetesV2/k8s.ts
@@ -15,7 +15,7 @@ interface KubeConfigOptions {
 
 type K8sObjectList =
     k8s.V1DeploymentList | k8s.V1ServiceList
-    | k8s.V1JobList | k8s.V1PodList;
+    | k8s.V1JobList | k8s.V1PodList | k8s.V1ReplicaSetList;
 
 export class K8s {
     logger: Logger;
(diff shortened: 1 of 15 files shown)


packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetesV2/k8s.ts
```diff
@@ -443,21 +450,27 @@ export class K8s {
         }
 
         await this._deleteObjByExId(exId, 'execution_controller', 'jobs', force);
+
+        if (force) {
+            await this._deleteObjByExId(exId, 'execution_controller', 'services', force);
+            await this._deleteObjByExId(exId, 'execution_controller', 'pods', force);
+            await this._deleteObjByExId(exId, 'worker', 'deployments', force);
+            await this._deleteObjByExId(exId, 'worker', 'pods', force);
```
busma13, line 458: I believe replicaset is the only resource type that jobs create that isn't on this list, maybe we should add it.

godber (approved):

Issue #3738, opened 2024-09-04
Fix bug: k8s --dev option prevents proper ex and worker pod shutdown


sotojn:
I've noticed when running in the local k8s envirnment when running `--dev` the execution controller and worker do not run their proper shutdown code when a `SIGTERM` is called.

For example i'll start a job then call `stop` on the job. I should see this in the logs of the execution controller:

```
[2024-09-04T16:55:17.255Z]  INFO: teraslice/10 on ts-exc-data-to-es-10b57068-276a-67g7p: execution_controller received process:SIGTERM, exiting in 30s... (assignment=execution_controller, module=execution_controller:shutdown_handler, worker_id=Z2XGrEF_, ex_id=10acf2e3-e852-4a29-b050-e7e416dd99e3, job_id=10b57068-276a-4d65-bb51-86ff829341a8)
[2024-09-04T16:55:17.357Z] DEBUG: teraslice/10 on ts-exc-data-to-es-10b57068-276a-67g7p: execution shutdown was called for ex 10acf2e3-e852-4a29-b050-e7e416dd99e3 (assignment=execution_controller, module=execution_controller, worker_id=Z2XGrEF_, ex_id=10acf2e3-e852-4a29-b050-e7e416dd99e3, job_id=10b57068-276a-4d65-bb51-86ff829341a8)
[2024-09-04T16:55:17.357Z] DEBUG: teraslice/10 on ts-exc-data-to-es-10b57068-276a-67g7p: stopping scheduler... (assignment=execution_controller, module=execution_scheduler, worker_id=Z2XGrEF_, ex_id=10acf2e3-e852-4a29-b050-e7e416dd99e3, job_id=10b57068-276a-4d65-bb51-86ff829341a8)
[2024-09-04T16:55:17.357Z] DEBUG: teraslice/10 on ts-exc-data-to-es-10b57068-276a-67g7p: execution 10acf2e3-e852-4a29-b050-e7e416dd99e3 is finished scheduling, 4 remaining slices in the queue (assignment=execution_controller, module=execution_scheduler, worker_id=Z2XGrEF_, ex_id=10acf2e3-e852-4a29-b050-e7e416dd99e3, job_id=10b57068-276a-4d65-bb51-86ff829341a8)
[...]

PR #3758, opened 2024-09-18
[teraslice] Fix teraslice k8s --dev mode process exit bug 

sotojn:

## Bug fixes
- Fixes bug when running `ts-scripts k8s` where the master, execution contoller, and workers don't properly run their respective exit processes.

Ref to issue #3738


busma13 (approved):
Looks good. Do we need to bump teraslice, or sneak this in with another release?

godber (approved):

PR #3763, opened 2024-09-20
[teraslice] Dependency updates 2024-09-19

busma13:
- New dependency:
  - ajv-formats@^3.0.1 - ajv no longer includes formats by default

- `dependabot.yml` ignore field updated to remove dependencies that were newer than the version being ignored and dependencies that have been removed. 

- Dependency updates:

| Package | From | To |
| ------------- | -------- | ------------- |
| @kubernetes/client-node | 0.21.0 | 0.22.0 |
| @types/diff | 5.2.1 | 5.2.2 |
| @types/jest | 29.5.12 | 29.5.13 |
| @types/uuid | 9.0.8 | 10.0.0 |
| @types/validator | 13.11.10 | 13.12.2 |
| ajv | 6.12.6 | 8.17.1 |
| auto-bind | 4.0.0 | 5.0.1 |
| chalk | 4.1.2 | 5.3.0 |
| debug | 4.3.6 | 4.3.7 |
| diff | 5.2.0 | 7.0.0 |
| express | 4.20.0 | 4.21.0 |
| geo-tz | 7.0.7 | 8.1.1 |
| ip-cidr | 3.1.0 | 4.0.2 |
| is-cidr | 4.0.2 | 5.1.0 |
| latlon-geohash | 1.1.0 | 2.0.0 |
| nanoid | 3.3.4 | 5.0.7 |
| p-event | 4.2.0 | 6.0.1 |
| p-map | 4.0.0 | 7.0.2 |
| package-json | 7.0.0 | 10.0.1 |
| pkg-up |3.1.0 | package-up@5.0.0 (name change) |
| pretty-bites | 5.6.0 | 6.1.1 |
| sort-package-json | 1.57.0 | 2.10.1 |
| uuid | 9.0.1 | 10.0.0 |


sotojn:
This looks good, do we update all of our packages by a patch?

sotojn (approved):

godber (approved):

packages/teraslice-messaging/src/utils/index.ts
```diff
@@ -1,9 +1,12 @@
 import os from 'node:os';
 import url from 'node:url';
-import { nanoid } from 'nanoid/async';
+import { nanoid } from 'nanoid';
 
 export function newMsgId(): Promise<string> {
-    return nanoid(10);
+    return new Promise((resolve) => {
```
jsnoble, line 6: there is no need to do this, you could just add `async` to the function, did they remove the async call?

packages/scripts/src/helpers/misc.ts
```diff
@@ -1,5 +1,5 @@
 import path from 'node:path';
-import pkgUp from 'pkg-up';
+import { packageUpSync } from 'package-up';
```
jsnoble, line 2: is this a different node-module?

jsnoble: nvm, saw the comment

packages/teraslice-messaging/src/utils/index.ts
busma13: In version 5.0 the starting using the Web Crypto API which has no async support. I am assuming we were using the async version so the event loop didn't get blocked. So I wrapped the call in a promise.
```

From the changelog:
## 5.0
* Moved Node.js version to Web Crypto API.
* Removed async API since Web Crypto API has only sync version.
```

packages/scripts/src/helpers/misc.ts
busma13: Same package, new npm name and github repo name.  https://github.com/sindresorhus/pkg-up now redirects to https://github.com/sindresorhus/package-up, so I'm positive it's the same package.

busma13: image

godber (approved):

PR #3767, opened 2024-09-23
[scripts, teraslice-cli, ts-transforms] Update execa from 5.1.0 to 9.4.0

busma13:
- Update execa from 5.1.0 to 9.4.0
- Update import statements and commands
- Add type check to exec() and runCli() commands to ensure string type is returned


packages/scripts/src/helpers/scripts.ts
```diff
@@ -68,6 +68,10 @@ export async function exec(opts: ExecOpts, log = true): Promise<string> {
         _opts.env = env;
         const subprocess = _exec(_opts);
         const { stdout } = await subprocess;
+
+        if (typeof stdout !== 'string') {
+            throw new Error('exec() requires ExecOpts that result in a stdout string. See the execa docs for details.');
+        }
```
busma13, line 74: Depending on the ExecOpts passed into exec stdout could be any of 'string | string[] | Uint8Array | unknown[]'. We don't use any options that change the stdout return type and always expect a string, so I think this shold be fine. Though the error message could be better

packages/scripts/src/helpers/scripts.ts
```diff
@@ -231,11 +235,11 @@ export async function getContainerInfo(name: string): Promise<any> {
 }
 
 export async function dockerNetworkExists(name: string): Promise<boolean> {
-    const subprocess = await execa.command(
+    const subprocess = await execaCommand(
         `docker network ls --format='{{json .Name}}' | grep '"${name}"'`,
         { reject: false }
     );
-    return subprocess.exitCode > 0;
+    return subprocess.exitCode ? subprocess.exitCode > 0 : false;
```
godber, line 242: I find this extra confusing, were you forced to do this?  What extra condition does it catch?

godber (approved):
I approve this, but I'm confused by your ternary.

packages/scripts/src/helpers/scripts.ts
busma13: The types changed so that exit code can be undefined "when the subprocess could not be spawned or was terminated by a signal."

If it exists it will be 'truthy' and we check if exitcode > 0, just like before, otherwise the command failed and we can't assume the docker network exists.

Issue #3765, opened 2024-09-20
is-ip dependency update issue


busma13:
Currently we are on `is-ip@3.1.0`. We were held back because of ESM. Current is 5.0.1. 

Updating to this version breaks tests that check to see if `null` is an IP addresses. Expected result is false, but a change in 5.0.1 checks the length of the parameter we pass in, resulting in an error.

~~I feel like the library should catch a null value and return false. If others agree I will open an issue.~~ Otherwise we might need to add null checks around the `isIP` function.
EDIT: After walking through some code with Joseph, I think we need to check types whenever we call this function. The `logicBuilder` function gets too complicated for typescript to realize the types don't match.

Updating to 5.0.0 should be possible.

```
 FAIL   data-mate  packages/data-mate/test/matcher/document-matcher-spec.ts
  Document-Matcher
    ✓ should not mutate the original data (7 ms)
    normal mode
      when testing ip queries
        can do ip type annotations with crazy data
          ✕ should be able to match on query ipField:"192.198.0.0/24" (27 ms)
    filterNil mode
      when testing ip queries
        can do ip type annotations with crazy data
          ✕ should be able to match on query ipField:"192.198.0.0/24" (4 ms)

  ● Document-Matcher › normal mode › when testing ip queries › can do ip type annotations with crazy data › should be able to match on query ipField:"192.198.0.0/24"

    TypeError: Cannot read properties of null (reading 'length')

      84 |             return checkCidr(ip, range);
      85 |         }
    > 86 |         if (isIP(ip)) return range.contains(ip);
         |             ^
      87 |         return false;
      88 |     };
      89 | }

      at length (node_modules/is-ip/index.js:12:13)
      at isIP (packages/data-mate/src/document-matcher/logic-builder/ip.ts:86:13)
      at DocumentMatcher.cb [as filterFn] (packages/data-mate/src/document-matcher/logic-builder/index.ts:47:16)
      at DocumentMatcher.filterFn [as match] (packages/data-mate/src/document-matcher/index.ts:23:21)
      at match (packages/data-mate/test/matcher/document-matcher-spec.ts:21:80)
          at Array.map (<anonymous>)
      at Object.map (packages/data-mate/test/matcher/document-matcher-spec.ts:21:46)

  ● Document-Matcher › filterNil mode › when testing ip queries › can do ip type annotations with crazy data › should be able to match on query ipField:"192.198.0.0/24"

    TypeError: Cannot read properties of null (reading 'length')

[...]

PR #3769, opened 2024-09-24
[data-mate, ts-transforms, utils] bump is-ip fr0m 3.1.0 to 5.0.1

busma13:
- bumps is-ip from 3.1.0 to 5.0.1
- Update imports for ESM
ref: #3765


packages/data-mate/src/document-matcher/logic-builder/ip.ts
```diff
@@ -89,6 +89,7 @@ function checkCidr(ip: string, range: any) {
 
 function pRangeTerm(range: any) {
     return function checkIP(ip: string) {
+        if (typeof ip !== 'string') return false;
```
busma13, line 92: @jsnoble What would you do in this case? The isIP function on line 96 changed to where it checks string.length on the input parameter. The way we have things typed it is possible to pass in null. This used to return false, now it throws.

jsnoble (approved):
LGTM

PR #3776, opened 2024-09-27
[teraslice, teraslice-client-js, terafoundation, scripts] update got from 11.8.3 to 13.0.0

busma13:
- update got from 11.8.3 to 13.0.0.
- remove `gotESM` module workaround from teraslice `api.ts`
- replace use of the removed `query` option field with `searchParams`.
- since `query` is no longer an option `RequestOptions` and `SearchOptions` types were identical. They are just a partial list of all got options. I chose to keep `RequestOptions` because that is more general. Requests that make a search still use a parameter named `searchOptions`, but it is now of type `RequestOptions`.
- `timeout` and `reply` options now take objects.
- add path to got types in `tsconfig.js`. Without this types are taken from `@types/got` which is a dependency of `kubernetes-client`. We can remove this once we go to `kubernetesV2`.

Note: got 14 requires node 20+, so we can only update to 13 at this time.


jsnoble (approved):
LGTM

PR #3780, opened 2024-10-01
[data-mate, ts-transforms, utils] update awesome-phonenumber from 2.70.0 to 7.1.0

busma13:
- update `awesome-phonenumber` from 2.70.0 to 7.1.0
- import `parsePhoneNumber` instead of default import. This is now a function that returns an object, not a class with functions to call. `valid() and getNumber() are replaced with accessing the data on the return object.


jsnoble (approved):
LGTM

Issue #3304, opened 2022-09-22
Update Blocked ESM Dependencies where possible


jsnoble:
There are packages that are being published in pure ESM module format (import and export syntax). The problem is that our compiler is not set to handle these types of packages. Most compilers like babel and ours convert code into CJS format (require statements) . 

We can change our compiler to export ESM modules instead but it will most likely require us to convert all our js files to use the ESM module format

Things to investigate:
-  How will this affect loading and executing assets?
- Are there ways to use old CJS formatted code despite the switch?
- Take in consideration on our private assets and how feasible it is to convert those
- how well can jest be used, it read somewhere it is having some issues

Overall, code that is already written in typescript should be able to easily convert over with a few setting changes to the build logic.

From what I found we need to use typescript v4.7 and higher for this, currently we are using 4.6

jsnoble:
List of packages converted to ESM with no CJS exports:

- got v12
- nanoid v4
- is-ip v5
- is-cidr v5
- auto-bind v5
- chalk v5
- execa v6
- pkg-up v4
- package-json v8
- p-map v5
- p-event v5
- pretty-bytes v6
- ip-bigint v5 (starts at v4)
- globby
- ajv?
- latlon-geohash

busma13:
sort-package-json starting from v2

godber:
@busma13 you could be referencing this issue in your PRs.

busma13:
I will from now on.

godber:
I think enough of these are updated now that this is good.  We're far less blocked on these and we're regularly updated.  The few long running exceptions like socket.io have their own issues or will get some.

PR #3781, opened 2024-10-01
[data-mate, utils] update ip-bigint from 3.0.3 to 8.2.0

busma13:
- update ip-bigint from 3.0.3 to 8.2.0
- import names updated: parse -> parseIP, stringify -> stringifyIp
- `parseIp` no longer validates an IP with the `isIP` library, so we add a check before calling the function.
- `stringifyIp` no longer validates that a bigint is greater than 0n or that it's not bigger than the largest IP address of its version, so a check was added for those.
- remove unneeded types files

ref: #3304


jsnoble (approved):
LGTM

PR #3783, opened 2024-10-01
[data-mate, utils, xlucene-parser] update date-fns and date-fns-tz

busma13:
- update date-fns from 2.30.0 to 4.1.0
- update date-fns-tz from 1.3.7 to 3.2.0
- switch to named imports
- `parse()` now only accepts strings, so `_parseDate` will convert inputs of type number to strings.


jsnoble (approved):
LGTM

PR #3785, opened 2024-10-03
[scripts] Update env variables for Rabbitmq

sotojn:

## Bug Fixes
- Fixes environment variables for Rabbitmq service that prevents a health check from completing

## Chores
- Updates **@terascope/scripts** from `v1.3.1` to `v1.3.2`


busma13 (approved):
LGTM

Issue #3772, opened 2024-09-25
Update typescript version - 5.5 or 5.6


busma13:
Currently we are using typescript 5.2.2. The latest version is 5.6.2.

It would be nice to get completely caught up.
We could then update typedoc which is blocked due to typescript version (#3768)

busma13:
In 5.5 the `suppressImplicitAnyIndexErrors` rule will be removed. That means we have to properly type any `object[key]` style call. We can turn off the rule in each package's tsconfig and then fix the errors in that package only, making the PRs smaller. Keeping that rule around until it can be removed from the root tsconfig will prevent any new code from introducing more errors of this type.

godber:
Is there an MR for this?

busma13:
Not yet.

PR #3789, opened 2024-10-07
[job-components, scripts, terafoundation, teraslice-cli, teraslice-client-js, teraslice-messaging, utils, xlucene-parser] fix implicit any index typescript errors

busma13:
In typescript 5.5 the `suppressImplicitAnyIndexErrors` rule will be removed. This PR fixes all ignored errors of this type in the following packages in preparation for the move to version 5.5 or above:
- job-components
- scripts
- terafoundation
- teraslice-cli
- teraslice-client-js
- teraslice-messaging
- utils
- xlucene-parser

`suppressImplicitAnyIndexErrors` is set to true in the tsconfig of all updated packages so we don't introduce any more of these errors before the rule is removed at the root level.

Todo: bump packages

Packages that still have these errors:
- data-mate - 1
- elasticsearch-store - 22
- teraslice - 124
- teraslice-state-storage - 13
- ts-transforms - 20

ref: #3772


packages/job-components/src/job-validator.ts
```diff
@@ -31,7 +31,7 @@ export class JobValidator {
         // top level job validation occurs, but not operations
         const jobConfig = validateJobConfig(this.schema, cloneDeep(jobSpec));
         const assetIds = jobConfig.assets || [];
-        const apis = {};
+        const apis: Record<string, any> = {};
```
jsnoble, line 34: would this not be `Record<string, OpAPI>` ? I believe we have a type for this and its not just any

packages/teraslice-cli/src/helpers/display.ts
```diff
@@ -5,6 +5,7 @@ import CliTable from 'cli-table3';
 import easyTable from 'easy-table';
 import prompts from 'prompts';
 import { toTitleCase } from '@terascope/utils';
+import { Action, Tense, UpdateActions } from '../interfaces';
```
jsnoble, line 8: this happens to work as it references a type folder which doesnt really comiple to anything, but you need a postfix here `../interfaces.js`

packages/teraslice-cli/src/helpers/jobs.ts
```diff
@@ -1,6 +1,6 @@
 import fs from 'fs-extra';
 import {
-    has, toString, pDelay, pMap, pRetry,
+    has, toString, pDelay, pMap, pRetry, getKeys,
```
jsnoble, line 3: This is more of a stylistic thing, we try to keep imports from going to wide, either put some of these on a new line or make all of them a new line, as long as that is not excessive long either

packages/teraslice-cli/src/helpers/jobs.ts
```diff
@@ -725,8 +727,8 @@ export default class Jobs {
     }
 
     formatJobConfig(jobConfig: JobConfigFile) {
-        const finalJobConfig: Partial<Teraslice.JobConfig> = {};
-        Object.keys(jobConfig).forEach((key) => {
+        const finalJobConfig: Record<string, any> = {};
```
jsnoble, line 730: Is there a reason why we are removing the type for Teraslice.JobConfig and making it more loose and uninformative to change it to any?

jsnoble: does Object.keys not take any types to its function? why was that lost in favor of getKeys? Doesn't that use types as well to infer what it has or expects?

packages/teraslice-cli/src/helpers/jobs.ts
```diff
@@ -738,12 +740,12 @@ export default class Jobs {
     }
 
     getLocalJSONConfigs(srcDir: string, files: string[]) {
-        const localJobConfigs = {};
+        const localJobConfigs: Record<string, any> = {};
```
jsnoble, line 743: should this not be `Record<string, Record<string, any>>`

packages/teraslice-messaging/src/cluster-master/server.ts
```diff
@@ -85,21 +86,22 @@ export class Server extends core.Server {
 
         this.handleResponse(socket, 'cluster:analytics', (msg: core.Message) => {
             const data = msg.payload as i.ExecutionAnalyticsMessage;
-            if (!this.clusterAnalytics[data.kind]) {
+            if (!(data.kind in this.clusterAnalytics)) {
                 return;
             }
+            const current = this.clusterAnalytics[data.kind as keyof i.ClusterAnalytics];
```
jsnoble, line 92: this does not necessarily need to be done now but shouldn't data.kind type be used in the type of ClusterAnalytics so it does not have to be forced and overridden here?

packages/utils/src/geo.ts
```diff
@@ -582,15 +582,28 @@ const esTypeMap = {
...
  */
 export function toGeoJSON(input: unknown): GeoShape | undefined {
     if (isGeoJSON(input)) {
         const { type: inputType } = input;
-        const type = esTypeMap[inputType] ? esTypeMap[inputType] : inputType;
-        return { ...input, type };
+        const type = _isEsTypeMapKey(inputType) ? esTypeMap[inputType] : inputType;
+
+        if (type === GeoShapeType.Point) {
+            return { ...input, type } as GeoShapePoint;
```
jsnoble, line 601: is this ok as is? In the function output type its GeoShape, but you are giving specific types as a return, this is most likely fine, but have you check the other spots using this? Is there any usage in spaces that might have a complication? It might be fine

packages/job-components/src/job-validator.ts
busma13: Looking closer, `apis` holds the constructors that create the OpAPIs. So `Record<string, OperationAPIConstructor>`

packages/utils/src/objects.ts
```diff
@@ -33,6 +33,15 @@ export function getFirstKey<T extends object>(input: T): (keyof T) | undefined {
     return Object.keys(input)[0] as keyof T;
 }
 
+/**
+ * Get the typed keys of an object. Same as Object.keys() but you should get
+ * proper typing if "O" is a typed object. Do not use with any object of
+ * type "object" or with "{}" as it will return type "never[]".
+ */
+export function getKeys<O extends object>(obj: O): Array<keyof O> {
```
jsnoble, line 41: this is incorrect, as it is it will accept `getKeys([1, 2, 3])` and it will not warn or throw. Many things in javascript are objects, arrays, functions, null and some other weird things

packages/utils/test/data-entity-spec.ts
```diff
@@ -7,6 +7,7 @@ import {
 import {
     parseJSON, cloneDeep, fastCloneDeep,
     firstToLower,
+    getKeys,
```
jsnoble, line 10: this does not need to be on its own line

packages/job-components/tsconfig.json
```diff
@@ -3,6 +3,7 @@
     "compilerOptions": {
         "outDir": "dist",
         "rootDir": ".",
+        "suppressImplicitAnyIndexErrors": false,
```
jsnoble, line 6: this should not be spread out everywhere, this should live in the top level tsconfig

packages/teraslice-cli/src/helpers/jobs.ts
busma13: If we type `finalJobConfig` as a partial then `undefined` is a valid key type. `finalJobConfig[key as keyof Partial<Teraslice.JobConfig>]` throws an error because none of the keys are 'assignable to undefined'. I could return `finalJobConfig as Partial<Teraslice.JobConfig>` so the function returns the type we expect. Maybe there is another way to do this besided loosening the types?

Object.keys returns a string[]. To preserve the types I saw people adding an overload to the `ObjectConstructor` interface. Or they were making there own function like `getKeys` instead.

packages/utils/src/geo.ts
busma13: Oh, i can just do `return {...input, type } as GeoShape`

packages/utils/src/objects.ts
busma13: I don't think I'm following. `Object.keys` accepts arrays, functions, etc without throwing. Wouldn't we want to follow suit here?

jsnoble (approved):
LGTM

packages/utils/src/objects.ts
```diff
@@ -33,6 +33,13 @@ export function getFirstKey<T extends object>(input: T): (keyof T) | undefined {
     return Object.keys(input)[0] as keyof T;
 }
 
+/**
+ * Verify that k is a key of object O
+*/
+export function isKey<T extends object>(O: T, k: PropertyKey): k is keyof T {
+    return k in O;
+}
+
```
busma13, line 42: I removed the `getKeys` function used in earlier commits in favor of `isKey`. `getKeys` used type assertion, so could be unsafe. `isKey` just returns a boolean and narrows the type for us.

PR #3791, opened 2024-10-10
Asset upload response

jsnoble:
- make the response consistent with our other api responses
- `_id` => `asset_id`
- this change is backwards compatible but `_id` is marked as deprecated.

### DIFF
--- a/e2e/test/cases/assets/simple-spec.ts
+++ b/e2e/test/cases/assets/simple-spec.ts
@@ -37,11 +37,15 @@ describe('assets', () => {
 
         // save the asset ID that was submitted to terslice
         const assetId = result.asset_id;
+        // TODO: remove this in teraslice V3
+        const oldAssetId = result._id;
+
         const response = await terasliceHarness.teraslice.assets.remove(assetId);
 
         // ensure the deleted asset's ID matches that of
         // the saved asset
         expect(assetId).toEqual(response.asset_id);
+        expect(oldAssetId).toEqual(response.asset_id);
     });
 
     // Test a bad asset
@@ -82,6 +86,8 @@ describe('assets', () => {
             blocking: true
         });
         const assetId = assetResponse.asset_id;
+        // TODO: remove this in teraslice V3
+        const oldAssetId = assetResponse._id;
 
         const ex = await terasliceHarness.submitAndStart(jobSpec);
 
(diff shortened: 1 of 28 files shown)


packages/teraslice-cli/src/cmds/assets/deploy.ts
```diff
@@ -198,7 +198,7 @@ export default {
                 const resp = await terasliceClient.assets.upload(assetZip, {
                     blocking: cliConfig.args.blocking
                 });
-                reply.green(`Asset posted to ${cliConfig.args.clusterAlias}: ${resp._id}`);
+                reply.green(`Asset posted to ${cliConfig.args.clusterAlias}: ${resp.asset_id}`);
```
godber, line 201: This doesn't leave much room for backwards compatibility.  Basically this version of Teraslice CLI will stop working with all versions of Teraslice that are old.  Can't we throw a ternary in there and do `asset_id` or `_id`?  Or something along those lines?

e2e/test/cases/assets/simple-spec.ts
```diff
@@ -36,12 +36,12 @@ describe('assets', () => {
...
 
         // save the asset ID that was submitted to terslice
-        const assetId = result._id;
+        const assetId = result.asset_id;
         const response = await terasliceHarness.teraslice.assets.remove(assetId);
 
         // ensure the deleted asset's ID matches that of
         // the saved asset
-        expect(assetId).toEqual(response._id);
+        expect(assetId).toEqual(response.asset_id);
```
godber, line 44: We should probably continue to expect both, and file an issue to remove `._id` in `v3.0.0`.  And add a comment here to remove the test for `._id` in 3.0.0 (linking to the issue above or vice versa)

packages/types/src/teraslice.ts
```diff
@@ -19,8 +19,13 @@ export interface AssetStatusResponse {
 }
 
 export type AssetIDResponse = {
-    _id: string;
+    asset_id: string;
+    /**
+        @deprecated Use asset_id instead
```
godber, line 24: The issue to remove this in 3.0.0 should link to this spot too.

godber (changes requested):
This fails linting ... you love linting.

There are a few minor things to address here.  Let me know if you have any concerns about them.

godber (approved):
Thanks for the fixes, looks great Jared!

Issue #3770, opened 2024-09-25
Expose execution status as virtual property on job resource - teraslice api


godber:
Would it make sense to expose the `ex._status` property as `.status` or perhaps `.exStatus` on the Teraslice job API resources?

I would expect this to be implemented by making a request to see if a current ex exists for the job, if so, return it's status, if not, return ... 🤷 

This would simplify a lot of client side logic.  I am not sure what the downsides would be.  The extra DB requests of course.  Anything else?

See also #535 , this is effectively a duplicate of that.

busma13:
In teraslice-cli we give the job a registered status if there isn't an execution
```
export interface JobMetadata {
    id: string;
    api: Job;
    config: Teraslice.JobConfig;
    status: Teraslice.ExecutionStatus | RegisteredStatus;
}

export type RegisteredStatus = keyof typeof RegisteredStatusEnum;

export enum RegisteredStatusEnum {
    no_execution = 'no_execution'
}
```

godber:
I talked with Kimbro about this a little bit today and he suggested if we chose to go this route we should probably make a distinction between this "virtual property" (if there's a word for this someone let me know) and the other properties.  That way if we start extending the use of this technique, they will all be grouped together.

So maybe something like

```json
"_ex": {
  "status": "running"
}
```

godber:
This PR:

https://github.com/terascope/teraslice/pull/3792

only added the functionality to `/jobs/<JOB_ID>`.  We want it added to `/jobs` too ...

PR #3792, opened 2024-10-10
Add execution info query to job api

sotojn:

# teraslice
## New Features
- Adds `ex` query parameter to the `v1/jobs/:job_id` api endpoint
  - It's now possible to to get a job with specific information about the current execution in one request
## Documentation
- Adds docs on how to use new `ex` query parameter in v1 jobs api endpoint

### DIFF
--- a/docs/management-apis/endpoints-json.md
+++ b/docs/management-apis/endpoints-json.md
@@ -290,6 +290,15 @@ $ curl 'localhost:5678/v1/jobs'
 
 Returns the job that matches given job id.
 
+**Query Options:**
+
+- `ex: string = [execution controller field options]`
+
+You can pass in a query using `ex` which takes field options for what you want returned in the `ex` object. This gives information on the current execution accociated with the specified job. If no execution is present, it will return the `ex` field with an empty object. If no fields are passed in, it will return all fields. Fields MUST be separated with commas. Example: `localhost:5678/v1/jobs/<job_id>?ex=_status,assets`
+
+Look at the returned object in [GET v1/ex](#get-v1ex) for valid field names.
+
+
 **Usage:**
 
 ```sh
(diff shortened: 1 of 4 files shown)


docs/management-apis/endpoints-json.md
```diff
@@ -952,3 +985,5 @@ $ curl 'localhost:5678/v1/ex/1cb20d4c-520a-44fe-a802-313f41dd5b05/controller'
     }
 ]
 ```
+
+## test /stuff
```
sotojn, line 989: We need to remove this.
```suggestion

```

packages/teraslice/src/lib/cluster/services/api.ts
```diff
@@ -286,10 +286,13 @@ export class ApiService {
         });
 
         v1routes.get('/jobs/:jobId', (req, res) => {
+            const { ex } = req.query;
             const { jobId } = req.params;
             // @ts-expect-error
             const requestHandler = handleTerasliceRequest(req as TerasliceRequest, res, 'Could not retrieve job');
-            requestHandler(async () => this.jobsStorage.get(jobId));
+            typeof ex === 'string'
```
godber, line 293: What happens if we specify `ex=dingo`?

godber: Can/should we reject the query if the field is not valid?

godber (commented):
Consider the invalid field case

packages/teraslice/src/lib/cluster/services/api.ts
sotojn: I've tested this yesterday and if you do `ex=dingo` it will return an empty execution object.
```
ex: {}
``` 
If you were to include a mix between valid and invalid fields such as `ex=_status,dingo,assets` the result would pull up all valid fields and ignore invalid fields.
```
ex: {
    _status: "running",
    assets: [
        abc123blahimahash
    ]
}
```

sotojn: This behavior is carried over on how elasticsearch requests handle invalid fields, which in this case just ignores them.

godber: Ok, that sounds appropriate.

docs/management-apis/endpoints-json.md
godber, line 989:
```suggestion
```

Not sure why this isn't deleted.  Lets try again

godber (changes requested):
you've left garbage around in here.

godber (approved):

godber:
This PR was for issue: https://github.com/terascope/teraslice/issues/3770

PR #3797, opened 2024-10-17
Ip duplicate code

### LINKED ISSUE #3752: duplicate code in data-mate
`data-mate/src/document-matcher/logic-builder/ip.ts` and `data-mate/src/data-frame/search/ip-utils.ts` are nearly the exact same with barely and difference besides some some string validation. Need to investigate why there is a duplicate and the consequences the code changes have made to each other and if there are behaviour differences between the two of them.

jsnoble:
- fix duplication of ip code between document matcher and dataframes
- differentiate between throwing and non throwing code
- resolves #3752


ciorg (approved):
LGTM

PR #3803, opened 2024-10-22
Utils ip dependency fix

ciorg:
* replaced `is-ip` dependency with `@chainsafe/is-ip` in utils package
* removed all dependencies on `is-ip`  and replaced those with the utils ip functions
* bumped versions for all affected packages


jsnoble (approved):
LGTM

Issue #3798, opened 2024-10-17
Remove Node 20 from test/build arrays


godber:
Let's remove Node 20 from all of our test arrays to reduce test time.

sotojn:
Should we continue building node 20 versions of the base image as newer versions of node 20 comes out? I feel like we should discontinue it in the base image as well. Also will this apply to all assets.

PR #3804, opened 2024-10-22
[CI] Remove node 20 from CI tests

sotojn:

- Removes all CI testing for node 20

Ref to issue #3798

### DIFF
--- a/scripts/check-latest-node-version.sh
+++ b/scripts/check-latest-node-version.sh
@@ -12,7 +12,7 @@ check_docker() {
 # Function to prompt the user for a node tag. Over time we will need to update
 # this as we discontinue and add major node versions.
 prompt_tag() {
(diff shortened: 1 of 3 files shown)


busma13 (approved):
LGTM

scripts/check-latest-node-version.sh
```diff
@@ -16,7 +16,7 @@ prompt_tag() {
     read -p "Enter the tag: " tag
 
     case $tag in
-        18|20|22)
+        18|22)
```
godber, line 19: This file has other occurrences of node 20 in it.  Please fix all of them.  Mostly found in user prompts like line 15 `[18, 20, 22]`.

godber (changes requested):
One small thing to fix.

godber (approved):

PR #3808, opened 2024-10-24
[teraslice, elasticsearch-store] opensearch2 mappings missing dynamics: false field

busma13:
- updates `ensureNoTypeInMapping()` to copy `dynamic` property when removing type from mapping.
- creates `method-helpers-spec.ts` and adds unit tests for `ensureNoTypeInMapping()`.
- creates an e2e `storage-spec.ts` to confirm that the jobs state storage index mapping is created with `dynamic` set to `false`.
- bump teraslice from version 2.6.2 to 2.6.3
- bump elasticsearch-api from 1.3.2 to 1.3.3
- bump elasticsearch-store from 1.3.2 to 1.3.3
- bump terafoundation from 1.5.3 to 1.5.4

Ref: #3809


packages/elasticsearch-store/src/elasticsearch-client/method-helpers/helper-utils.ts
```diff
@@ -34,9 +34,11 @@ export function ensureNoTypeInMapping(mappings: Record<string, any> | undefined)
         for (const [k, v] of Object.entries(mappings)) {
```
godber, line 34: Can this function be implemented more generically?  That is, for any set of incoming property names?  Or is there some reason that we need to explicitly list properties.  Are we deliberately trying to get rid of some incoming properties?  If so, presumably that set of incoming properties would be better known and hence more easily filtered out?

busma13: That was my first thought, but that `type` key could be anything. Beyond the teraslice state store mappings I think we would need a list of every ES6 style mapping in prod that still uses a `type`.

busma13: Maybe it would be safe to copy everything from the second level in, except `_all`, as that will throw an error in ES8 or OS2?
```
{
        job: { // this level can be anything, so remove unless it's a known key
            _all: { // remove this property
                enabled: false
            },
            _meta: { // keep everything else at this level
                foo: 'bar'
            },
            dynamic: false,
            properties: {
                active: {
                    type: 'boolean'
                },
                job_id: {
                    type: 'keyword'
                },
                _context: {
                    type: 'keyword'
                },
                _created: {
                    type: 'date'
                },
                _updated: {
                    type: 'date'
                },
                _deleted: {
                    type: 'boolean'
                },
                _deleted_on: {
                    type: 'date'
                }
            }
        }
    }
```
Or maybe we look for `properties` and keep everything on that same level unless it is known to be invalid?

packages/elasticsearch-store/test/method-helpers-spec.ts
```diff
@@ -0,0 +1,134 @@
...
+                }
+            };
+
+            it('should remove myType wrapper object from mappingsWithType', () => {
+                const newMappings = ensureNoTypeInMapping(mappingsWithType);
+                expect(newMappings.myType).toBe(undefined);
+            });
+
+            it('should remove _all from mappingsWithType', () => {
+                const newMappings = ensureNoTypeInMapping(mappingsWithType);
```
busma13, line 41: I don't need to call this repeatedly.

e2e/test/cases/cluster/storage-spec.ts
```diff
@@ -0,0 +1,29 @@
...
+    beforeAll(async () => {
+        terasliceHarness = new TerasliceHarness();
+        await terasliceHarness.init();
+        await terasliceHarness.resetState();
+    });
+
+    it('should have a jobs index with dynamic mapping false', async () => {
+        const mapping = await terasliceHarness.client.indices.getMapping({ index: '*__jobs' });
+        const indexName = Object.keys(mapping)[0];
+        const searchVersion = (await terasliceHarness.client.info()).version.number;
```
jsnoble, line 16: `(await terasliceHarness.client.info()).version.number;` is not very readable and its uneeded. The client already has that information on it so there is no need to make the call either through `terasliceHarness.distributionMeta` though the types wont like that or through `terasliceHarness.__meta`

e2e/test/config.ts
```diff
@@ -111,5 +111,6 @@ export {
     MINIO_ACCESS_KEY,
     MINIO_SECRET_KEY,
     ENCRYPT_MINIO,
-    ROOT_CERT_PATH
+    ROOT_CERT_PATH,
+    TEST_OPENSEARCH
```
jsnoble, line 115: you can also pull the version from the enviroment not only just TEST_OPENSEARCH

jsnoble (approved):
LGTM besides some simple test changes

e2e/test/cases/cluster/storage-spec.ts
busma13: I didn't know that. Thanks

e2e/test/config.ts
busma13: For my use case I would need to expose both elasticsearch and opensearch versions, so I'll stick will getting the version from client.__meta.

godber (approved):

PR #3812, opened 2024-10-30
[teraslice] add typescript types to KubernetesV2 backend

busma13:
- Add types to functions and classes in the `kubernetesV2` backend
- Make `K8sResource` an abstract class and add `K8sJobResource`, `K8sServiceResource`, and `K8sDeploymentResource` classes. This makes creating resources from templates cleaner.
- Extend `K8sResource` types as `TSResource` types with all required fields defined.
- Add types to tests
- bump teraslice from 2.6.3 to 2.7.0


packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetesV2/k8s.ts
```diff
@@ -536,19 +518,28 @@ export class K8s {
...
 
         // the selector provided to list above should always result in a single
         // deployment in the response.
-        // TODO: test for more than 1 and error
+        if (listResponse.items.length === 0) {
+            const msg = `Teraslice deployment matching the following selector was not found: ${selector} (retriable)`;
+            this.logger.warn(msg);
+            throw new TSError(msg, { retryable: true });
+        } else if (listResponse.items.length > 1) {
+            throw new TSError(`Unexpected number of Teraslice deployments matching the following selector: ${selector}`, {
```
godber, line 539: So what is the consequence of this being thrown?

godber (approved):

packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetesV2/k8s.ts
busma13: This function can only be reached from a post to `/jobs/:jobId/_workers` or `/ex/:exId/_workers`. There are no catches along that route, so it should return the error back to the user.

packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetesV2/index.ts
```diff
@@ -42,8 +42,7 @@ export class KubernetesClusterBackendV2 {
             this.logger,
             null,
             kubernetesNamespace,
-            // @ts-expect-error
-            context.sysconfig.teraslice.kubernetes_api_poll_delay,
+            context.sysconfig.teraslice.kubernetes_api_poll_delay || 1000,
```
jsnoble, line 45: Is this needed? Usually default values and settings belong in the schema, it worries me a little bit that we are not relying on it. If its conditional on other settings we could rely on post validation to set it or something

packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetesV2/k8sResource.ts
```diff
@@ -160,42 +150,29 @@ export class K8sResource {
             dockerImage,
             execution: safeEncode(this.execution),
             exId: this.execution.ex_id,
-            // @ts-expect-error TODO: not sure where these come from
-            exName: this.execution.k8sName,
-            // @ts-expect-error TODO: not sure where these come from
-            exUid: this.execution.k8sUid,
+            exName: this.exName,
+            exUid: this.exUid,
```
jsnoble, line 154: what is exUid? we have a jobId, and other spots we have exId, is that somehow different from exUId or is it a k8 specific thing?

packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetesV2/index.ts
busma13: This is one of those cases where the type can be undefined, but terafoundation validation will fill in a default. Maybe it is better to throw an error if this value is undefined? Or is there some better way to handle this that I'm missing?

packages/teraslice/test/lib/cluster/services/cluster/backends/kubernetes/v2/k8s-v2-spec.ts
```diff
@@ -70,7 +70,7 @@ describe('k8s', () => {
                 .query({ labelSelector: 'app=teraslice' })
                 .reply(200, { kind: 'PodList' });
 
-            const pods = await k8s.list('app=teraslice', 'pods');
+            const pods = await k8s.list('app=teraslice', 'pod');
```
jsnoble, line 73: I noticed this, why the move from a plural to a singular thing? Currently in teraslice we have a naming system that when we interact with services or list thing they are plural, only really singular if its a specific instance of something

packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetesV2/k8sResource.ts
busma13: This is the UID assigned to the execution controller pod by kubernetes. It is used as the `owner reference` by workers to tell the kubernetes system what resource is in control of their pod.

packages/teraslice/test/lib/cluster/services/cluster/backends/kubernetes/v2/k8sResource-v2-spec.ts
```diff
@@ -37,48 +38,65 @@ describe('k8sResource', () => {
...
+            expect(configVolumeMount)
                 .toEqual(yaml.load(`
                     mountPath: /app/config
                     name: config`));
         });
 
         it('has valid resource object when terasliceConfig has kubernetes_image_pull_secret.', () => {
             terasliceConfig.kubernetes_image_pull_secret = 'teraslice-image-pull-secret';
-            const kr = new K8sResource('deployments', 'worker', terasliceConfig, execution, logger);
+            const kr = new K8sResource('deployment', 'worker', terasliceConfig, execution, logger, 'example-job-abcd', 'UID1');
```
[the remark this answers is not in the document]
busma13: `K8sResource returns a single type with K8sResource.resource being a union of 3 types. I think I could make 3 distinct K8sResource types that each return a specific K8sResource.resource. I'm not sure why I missed that.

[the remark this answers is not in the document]
busma13: Oh wait, K8sResource is a class. I would need to create 3 distinct classes. That would create some duplicate code, but it would also reduce a lot of the checks we make to figure out which resource we are trying to build.

packages/teraslice/test/lib/cluster/services/cluster/backends/kubernetes/v2/k8s-v2-spec.ts
busma13: I decided to go to plural.

packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetesV2/k8s.ts
```diff
@@ -3,19 +3,20 @@ import {
...
+import {
+    DeleteApiResponse, DeleteParams, DeleteResponseBody,
+    KubeConfigOptions, K8sResource, ListParams,
+    NodeType, PatchApiResponse, ResourceApiResponse,
+    ResourceListApiResponse, ResourceType, ScaleOp,
+    TSDeployment, TSDeploymentList, TSJob,
+    TSJobList, TSPod, TSPodList,
+    TSReplicaSet, TSReplicaSetList, TSResource,
+    TSResourceList, TSService, TSServiceList
+} from './interfaces.js';
```
busma13, line 19: Would it be better to do `import * as i from './interfaces.js'` since we are importing so many types?

packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetesV2/utils.ts
```diff
@@ -3,17 +3,36 @@ import path from 'node:path';
...
 import barbe from 'barbe';
 import { isTest } from '@terascope/utils';
+import * as k8s from '@kubernetes/client-node';
+import {
+    K8sConfig, NodeType, K8sResource,
+    TSDeployment, TSJob, TSPod, TSService,
+    TSReplicaSet, TSResource, K8sResourceList,
+    TSDeploymentList, TSJobList, TSPodList,
+    TSReplicaSetList, TSResourceList, TSServiceList
+} from './interfaces.js';
```
busma13, line 13: `import * as i` here too?

packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetesV2/utils.ts
```diff
@@ -57,3 +71,162 @@ export function getRetryConfig() {
...
+    return resource instanceof k8s.V1Pod || resource.kind === 'Pod';
+}
+
+export function isReplicaSet(resource: K8sResource): resource is k8s.V1ReplicaSet {
+    return resource instanceof k8s.V1ReplicaSet || resource.kind === 'ReplicaSet';
+}
+
+export function isService(resource: K8sResource): resource is k8s.V1Service {
+    return resource instanceof k8s.V1Service || resource.kind === 'Service';
+}
```
busma13, line 93: Break these into 2 functions? `instanceof` only works to narrow which `K8sResource` we have.
Proving a manifest, which is just an object, is a `K8sResource` using `instanceof` does not work.
`resource.kind` does not get returned when calling `k8s.list()`, so we need to check with
`instanceof` in that case.

packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetesV2/index.ts
busma13: I made this property non-optional on the teraslice Config type since it has a default in the schema.

packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetesV2/k8s.ts
jsnoble: yes, when it becomes tedious like this it would be better to refactor it like that

packages/teraslice/test/lib/cluster/services/cluster/backends/kubernetes/v2/k8s-v2-spec.ts
```diff
@@ -2,6 +2,10 @@
 // output for nock
 //   env DEBUG='nock*' make test
```
jsnoble, line 3: this should not be left in here

packages/job-components/src/test-helpers.ts
```diff
@@ -174,6 +174,7 @@ export class TestContext implements i.Context {
                     analytics: 'yearly',
                     state: 'montly',
                 },
+                kubernetes_api_poll_delay: 1000,
```
jsnoble, line 177: Is this the only kubernetes config or are there other ones as well?

packages/types/src/teraslice.ts
```diff
@@ -480,7 +480,7 @@ export interface Config {
     execution_controller_targets?: ExecutionControllerTargets[];
     hostname: string;
     index_rollover_frequency: IndexRolloverFrequency;
-    kubernetes_api_poll_delay?: number | 1000;
+    kubernetes_api_poll_delay: number | 1000;
```
jsnoble, line 483: so there seems to be a mistake, there should be no numbers at all in `interface Config`, also, the other kubernetes config values, are they really optional? Or do they have a real default setting that should be manifested in the types?

PR #3813, opened 2024-10-31
[terasclice] Ensure asset uploads are zip files

busma13:
- Add `isZipFile()` to `file_utils` which checks the magic numbers at the beginning of a buffer to see if they match those of a zip file.
- `AssetStorage.save()` uses `isZipFile()` to check if the uploaded asset is a zip and throws an error if it isn't.
- Add `fake_zip.zip` and add a test that calls `AssetStorage.save()` on it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "teraslice-workspace",
     "displayName": "Teraslice",
-    "version": "2.6.3",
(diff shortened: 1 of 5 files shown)


godber (approved):

sotojn (approved):
This looks good

Issue #3728, opened 2024-08-22
Fix ts-scripts (docker) images test, also rename to be more explicit


godber:
This test was disabled as part of the ESM refactor:

https://github.com/terascope/teraslice/blob/master/packages/scripts/test/images-spec.ts#L36

It should be re-enabled or re-implemented in a way that is compatible with ESM.

Also, anywhere the test says just "images" ... reconsider the wording to be more explicit, we're talking about "docker images" here right?  A quick read without context is a little hard to understand.

busma13:
I'm leaning toward removing this test. I don't see any good way to refactor this code. In the end some functions will need to be mocked or spied because we can't call `docker pull` or `docker save`. 

Two options I don't like:
- Create a `dryrun` flag that will skip any docker command and then return a value from the functions that call docker commands so that the test has some change to test against. Passing a return value that is only used for testing seems wrong though.
- Create a `dryrun` flag that skips docker commands, but saves a fake file to the cache. That way we know `saveAndZip` was called at least.

godber:
It's probably safe to delete this skipped text.

PR #3815, opened 2024-11-06
[ts-scripts] remove saveImages() test.

busma13:
The `saveImages()` function can't be tested without mocking because it uses `docker pull` and `docker save`, which we don't want to do in CI. That would defeat the purpose of creating a cache of docker images. This is only used for CI caching purposes, so loss of test coverage is acceptable.

Ref: #3728


sotojn (approved):
LGTM

godber (approved):

Issue #3743, opened 2024-09-09
PromMetrics exports execution and controller metrics for old/stopped executions


briend:
After setting teraslice `v2.1.0` to use the new internal prom metrics, the number of metrics exported can (seemingly) grow without bounds.  Old executions for the same `job_id` will accumulate, whereas previously only (apparently) non-terminal statuses like `running` were exported.  If old executions are eventually removed automatically, maybe this new behavior makes sense, but I'm not sure if they are or if there are plans for that.

```
terafoundation:
  prom_metrics_enabled: true
```

Only one of these is in state `running` and the other two are `stopped`; previously only `1` series was exported and now there are `3`.  Same for `teraslice_execution_status`, `teraslice_controller_.*`, etc.

```
teraslice_execution_info{assignment="master", container="teraslice", ex_id="99d401f2-9d11-4ce9-8882-fbfc246ba2f3", image="terascope/teraslice:v2.1.0-nodev18.19.1", instance="node1", job="teraslice-ops2", job_id="92f53103-7de6-4bb3-929b-e2bb13598502", name="teraslice-ops2", namespace="ts-ops2", pod="teraslice-ops2-master-84d7d64f7b-5n9sj", prometheus="ops/mon-prometheus", service="teraslice-ops2", version="2.1.0"}  1

teraslice_execution_info{assignment="master", container="teraslice", ex_id="a837cda0-7801-43e0-96e5-47ebfd3f1303", image="terascope/teraslice:v2.1.0-nodev18.19.1", instance="node1", job="teraslice-ops2", job_id="92f53103-7de6-4bb3-929b-e2bb13598502", name="teraslice-ops2", namespace="ts-ops2", pod="teraslice-ops2-master-84d7d64f7b-5n9sj", prometheus="ops/mon-prometheus", service="teraslice-ops2", version="2.1.0"} 1

teraslice_execution_info{assignment="master", container="teraslice", ex_id="undefined", image="terascope/teraslice:v2.1.0-nodev18.19.1", instance="node1", job="teraslice-ops2", job_id="undefined", name="teraslice-ops2", namespace="ts-ops2", pod="teraslice-ops2-master-84d7d64f7b-5n9sj", prometheus="ops/mon-prometheus", service="teraslice-ops2", version="2.1.0"}  1
```

Also note one of these has `job_id="undefined"` and `ex_id="undefined"`, which may be an additional bug, since the ex that show up with the normal `/txt/ex` look normal:

```
[...]

PR #3821, opened 2024-11-12
[teraslice] fix 'teraslice_version' and 'version' prom metrics labels

busma13:
This PR makes the following changes to match the external prom-metrics exporter exactly:
- The `teraslice_version` label value should start with a `v`
- The `version` label should be the complete docker image tag unless there is an `_`. In that case the tag will be everything before the underscore.

ref: #3743


sotojn (approved):
LGTM

godber (approved):

Issue #3685, opened 2024-07-16
Add Asset/Teraslice Version Dependency Check on Asset Load


godber:
Add the ability for an asset to state what versions of Teraslice it requires and Teraslice should use this to validate the asset against it's current version at the time the asset is uploaded.

Right now the `asset.json` file looks like this:

```json
{
    "name": "elasticsearch",
    "version": "4.0.0"
}
```

I propose we extend it with a `dependencies` property, like the `package.json` 

```json
{
    "name": "elasticsearch",
    "version": "4.0.0"
    "dependencies": {
        "teraslice": "^2.0.0"
   }
}
```

Other things to consider:

* Should the output of `/txt/assets` change (show incompatible versions)
* We could publish an asset registry as a json file in our docs and use it to render a compatibility chart

kstaken:
This seems more akin to the `engines` setting in package.json rather than `dependencies`. You're defining what the minimum runtime environment needs to be which is more of an execution time dependency rather than a build time dependency.

godber:
> This seems more akin to the engines setting in package.json rather than dependencies. You're defining what the minimum runtime environment needs to be which is more of an execution time dependency rather than a build time dependency.

Yeah, excellent observation.  I have no strong opinions on how the key is named and had only decided to mimic npm because I wanted to avoid inventing my own thing.  These are still things I consider open questions:

* What should we name the Teraslice version requirement property in the asset?
* Is there some reason we should mimic npm or even better, are there any arguments for not mimicking it?

busma13:
Possible keys:
```
{
    "name": "elasticsearch",
    "version": "4.0.0",
    "needs": {
        "teraslice": "^2.0.0"
    },
    "requires": {
        "teraslice": "^2.0.0"
    },
    "minimum-versions": {
        "teraslice": "^2.0.0"
    },
    "minimum-teraslice-version": "^2.0.0"
}
```

PR #3822, opened 2024-11-12
[teraslice] Asset load dependency check

busma13:
- adds a check on asset upload to see if the asset is compatible with the teraslice version
- creates a test asset with a high minimum teraslice version
- test the test asset to ensure an error is thrown
- fix a bug that leaves behind an s3 object in some situations if an asset fails to upload

ref: #3685


packages/teraslice/src/lib/utils/file_utils.ts
```diff
@@ -62,6 +62,9 @@ export interface AssetJSON {
     arch?: string;
     platform?: string;
     node_version?: number;
+    minimum_versions?: {
+        teraslice: string;
+    };
```
busma13, line 67: We probably want to think hard about what we name this and if it should be an object that holds several versions or not.

godber: I'd say it's unavoidable since you can't know what breaking changes can be implemented in the future.

As long as this value won't need to be validated by convict that nested value looks OK.  But every time I've tried to nest something like that I've gotten hung up when it goes through convict for validation (I think that was the issue).  If we wanted to play it dumb and safe we could just do this: `"minimum-teraslice-version"`.

godber:
We really shouldn't be adding large binary files to the repo, especially one this large.  At this point, since you've pushed the branch it's already there.  So leave it.

packages/teraslice/src/lib/utils/file_utils.ts
```diff
@@ -80,46 +83,54 @@ export async function verifyAssetJSON(id: string, newPath: string): Promise<Asse
...
+    }
+    if (!metadata.platform) {
+        delete metadata.platform;
+    }
+    if (!metadata.arch) {
+        delete metadata.arch;
+    }
+    if (metadata.minimum_versions) {
+        const terasliceVersion = getPackageJSON().version;
+        if (semver.gt(metadata.minimum_versions.teraslice, terasliceVersion)) {
```
godber, line 129: What happens if the asset contains:

```
    minimum_versions: {
    };
```

So `metadata.minimum_versions.teraslice` is undefined?

Maybe we should just go back to a single property, no nested thing.

godber (changes requested):
Lets discuss this in the morning.

packages/teraslice/src/lib/utils/file_utils.ts
busma13: Yeah this would have thrown an error because semver.gt() won't take undefined as a parameter.

busma13:
I think the single property makes more sense. Any other compatibility check can have it's own top level property. This also gives me a chance to use a smaller asset.

godber (approved):

PR #3830, opened 2024-11-14
[scripts] k8s-env automatically switches to es7 on Arm Macs 

sotojn:

### New features

- Introduces a check within `k8s-env` command that checks the host system’s architecture and the Elasticsearch version. If the host architecture matches ARM and the Elasticsearch version is 6.x, `k8s-env` will automatically switch the Elasticsearch version to `7.9.3` to ensure compatibility.

### Bumps

- @terascope/scripts from `v1.5.0` to `v1.5.1`

### DIFF
--- a/packages/scripts/src/helpers/test-runner/services.ts
+++ b/packages/scripts/src/helpers/test-runner/services.ts
@@ -20,6 +20,7 @@ import { TestOptions } from './interfaces.js';
 import { Service } from '../interfaces.js';
 import * as config from '../config.js';
 import signale from '../signale.js';
+import os from 'node:os';
 
 const logger = debugLogger('ts-scripts:cmd:test');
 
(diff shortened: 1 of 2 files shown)


packages/scripts/src/helpers/test-runner/services.ts
```diff
@@ -625,8 +626,13 @@ async function checkElasticsearch(options: TestOptions, startTime: number): Prom
...
             const expected = config.ELASTICSEARCH_VERSION;
-
-            if (semver.satisfies(actual, `^${expected}`)) {
+            if (
+                semver.satisfies(actual, `^${expected}`)
+                // Or we may override semver match in this scenario
+                || (os.arch().includes('arm')
+                    && expected.startsWith('6') && actual === '7.9.3'
+                    && options.testPlatform.includes('kubernetes'))
+            ) {
```
busma13, line 635: If we do this for docker too this check isn't needed.

busma13 (commented):
Reviewed in case we do decide to go this route,  but we might just update the default ES version to 7

sotojn:
We've decided to instead change the default version of es 6 to es 7 so that we can have cross compatibility between arm and amd users. #3833

Issue #3825, opened 2024-11-13
Review and refactor ts-scripts sync


busma13:
While working on #3823 I ran into an issue where sync assumes that we will be using a minor range(^), so `yarn sync` failed when it changed all the `~` back to `^`.

We at minimum need to convert `sync` to expect a patch range, but we have run into other problems with sync doing things we did not expect recently.
If we can review the code and determine what functionality is still needed, and it doesn't look to difficult to separate it out and discard the rest, that would be best.

busma13:
After taking a look, we are going to remove `generateTSConfig` and `executeHook` from the `sync` command. The tsconfig is easy to update and executeHook is not used anywhere.

godber:
Closed by #3831

PR #3824, opened 2024-11-13
[all] Update dependency range specificity

busma13:
This PR makes the following changes:
- Update all dependency versions to use patch level specificity(~) instead of minor level(^)
- Remove unused dependencies:
  - terafoundation:  
    - @types/mongoose
    - @types/redis
  - teraslice-cli:
    - events
    - yo

TODO:
- waiting on PR to modify ts-scripts sync command #3825 

ref: #3823

PR #3831, opened 2024-11-14
[all] syncAll refactor and patch dependency range fix

busma13:
- Update all places in ts-scripts where `^` was the expected default range for a package version and instead use `~`. This will match the patch level specificity we want in all our packages now.
- remove `executeHook()` function: this was a feature that let you run a script in the middle of the `sync` or `bump` commands. It had not been used for a long time.
- remove `generateTSConfig()` function: it was decided it would be best to handle tsconfig changes manually.
- Update all dependency versions to use patch level specificity(~) instead of minor level(^)
- Remove unused dependencies:
     - terafoundation:
        - @types/mongoose
        - @types/redis
     - teraslice-cli:
         - events
          - yo
          - @types/easy-table

ref: #3825, #3824


jsnoble (approved):
LGTM

Issue #3807, opened 2024-10-23
k8senv - Switch to es7 automatically on Arm Macs


godber:
Couldn't `k8senv` detect if the user is running on ARM MacOS and use es7 automatically instead of this:

> ERROR: image: "elasticsearch:6.8.6" not present locally

godber:
We didn't do the switch to ES 7 for arm, instead we made es7 the default.  es6 tests still run though.

PR #3833, opened 2024-11-18
[scripts] change default Elasticsearch test version to es 7.9.3

sotojn:

## Scripts 

- Changes the default Elasticsearch version from `6.8.6` to `7.9.3`
  - This resolves an issue of incompatibility of ARM users and the Elasticsearch `6.8.6` image
  - Bumped **@terascope/scripts** from version `1.5.0` to `1.5.1`

Ref to issue #3807


packages/elasticsearch-api/package.json
```diff
@@ -16,7 +16,7 @@
     "scripts": {
         "test": "TEST_RESTRAINED_ELASTICSEARCH='true' ts-scripts test . --",
         "test:debug": "TEST_RESTRAINED_ELASTICSEARCH='true' ts-scripts test --debug . --",
-        "test:elasticsearch6": "TEST_RESTRAINED_ELASTICSEARCH='true' ts-scripts test . --",
+        "test:elasticsearch6": "TEST_RESTRAINED_ELASTICSEARCH='true' ELASTICSEARCH_VERSION='6.8.6' ts-scripts test . --",
         "test:elasticsearch7": "TEST_RESTRAINED_ELASTICSEARCH='true' ELASTICSEARCH_VERSION='7.9.3' ts-scripts test . --",
         "test:elasticsearch8": "TEST_RESTRAINED_ELASTICSEARCH='true' ELASTICSEARCH_VERSION='8.1.2' ts-scripts test . --",
         "test:opensearch1": "TEST_RESTRAINED_OPENSEARCH='true' ts-scripts test --debug . --",
```
busma13, line 22: should we be explicit with the opensearch version as well?

packages/teraslice/package.json
```diff
@@ -27,7 +27,7 @@
         "start": "node service.js",
         "test": "TEST_ELASTICSEARCH='true' ts-scripts test . --",
         "test:debug": "TEST_ELASTICSEARCH='true' ts-scripts test --debug . --",
-        "test:elasticsearch6": "TEST_ELASTICSEARCH='true' ts-scripts test . --",
+        "test:elasticsearch6": "TEST_ELASTICSEARCH='true' ELASTICSEARCH_VERSION='6.8.6' ts-scripts test . --",
         "test:elasticsearch7": "TEST_ELASTICSEARCH='true' ELASTICSEARCH_VERSION='7.9.3' ts-scripts test . --",
         "test:opensearch1": "TEST_OPENSEARCH='true' ts-scripts test . --",
```
busma13, line 32:
```suggestion
        "test:opensearch1": "TEST_OPENSEARCH='true' OPENSEARCH_VERSION='1.3.0' ts-scripts test . --",
```

packages/elasticsearch-store/package.json
```diff
@@ -22,7 +22,7 @@
         "build:watch": "yarn build --watch",
         "test": "TEST_ELASTICSEARCH='true' ts-scripts test . --",
         "test:debug": "TEST_ELASTICSEARCH='true' ts-scripts test --debug . --",
-        "test:elasticsearch6": "TEST_ELASTICSEARCH='true' ts-scripts test . --",
+        "test:elasticsearch6": "TEST_ELASTICSEARCH='true' ELASTICSEARCH_VERSION='6.8.6' ts-scripts test . --",
         "test:elasticsearch7": "TEST_ELASTICSEARCH='true' ELASTICSEARCH_VERSION='7.9.3' ts-scripts test . --",
         "test:elasticsearch8": "TEST_ELASTICSEARCH='true' ELASTICSEARCH_VERSION='8.1.2' ts-scripts test . --",
         "test:opensearch1": "TEST_OPENSEARCH='true' ts-scripts test . --",
```
busma13, line 28:
```suggestion
        "test:opensearch1": "TEST_OPENSEARCH='true' OPENSEARCH_VERSION='1.3.0' ts-scripts test . --",
```

packages/elasticsearch-api/package.json
busma13:
```suggestion
        "test:opensearch1": "TEST_RESTRAINED_OPENSEARCH='true' OPENSEARCH_VERSION='1.3.0' ts-scripts test --debug . --",
```

busma13 (changes requested):

packages/elasticsearch-api/package.json
godber: I think this is a decent suggestion but I think the situation is slightly different.  If we hardcode the version to `1.3.0` it will make it harder to update the opensearch version to a new 1.X version.  It's also outside the scope of this PR.  So I am going to merge it as is.  Definitely a good question to be asking.

godber (approved):

PR #3841, opened 2024-11-25
Op asset specificity

jsnoble:
- jobs will throw if there are name collisions and the op name has no asset identifier
- can now specify op/reader/apis names as `op-name@assetName:v1.4.0` or  `op-name@assetName`


godber:
When you're back Jared, you'll need to rebase this and bump the version to 2.10.0 since we did a 2.9.0 release today.

jsnoble:
To test this out, setup a running teraslice instance, and use teraslice-cli to upload assets

```
teraslice-cli assets deploy localhost terascope/elasticsearch-assets@v4.0.2
teraslice-cli assets deploy localhost terascope/elasticsearch-assets@v4.0.5
```

Sample Job that should fail because we don't know which asset to use for the op, please adjust index and any other settings for your setup
```
{
    "name": "test",
    "lifecycle": "once",
    "workers": 1,
    "analytics": true,
    "assets": [
        "elasticsearch:4.0.2",
        "elasticsearch:4.0.5"
    ],
    "operations": [
        {
            "index": "ts_test_example-1000",
            "size": 10000,
            "date_field_name": "created",
            "preserve_id": true
        },
        {
            "_op": "elasticsearch_bulk",
            "index": "op_asset_version_test",
            "preserve_id": true,
            "size": 10000
        }
    ]
}
```

Sample job that should pass, please make the same updates to index and any other changes neccessary
```
{
    "name": "test",
    "lifecycle": "once",
    "workers": 1,
    "analytics": true,
    "assets": [
        "elasticsearch:4.0.2",
        "elasticsearch:4.0.5"
    ],
    "operations": [
        {
            "_op": "elasticsearch_reader@elasticsearch:4.0.2",
            "index": "ts_test_example-1000",
            "size": 10000,
            "date_field_name": "created",
            "preserve_id": true
        },
        {
            "_op": "elasticsearch_bulk@elasticsearch:4.0.5",
            "index": "op_asset_version_test",
            "preserve_id": true,
            "size": 10000
        }
    ]
}
```

ciorg:
Here's my test

job config:
```
"assets": [
        "common_processors:0.16.0",
        "standard:1.2.0"
    ],
    "operations": [
        {
            "_op": "data_generator",
            "size": 1
        },
        {
            "_op": "filter",
            "field": "ip",
            "value": "0.0.0.0"
        },
        {
            "_op": "noop"
        }
    ]
```

When I tried to run with this config I got  the error `Multiple assets containing the same processor name _op: "filter", please specify which specific asset to use on its name`.   Added `1filter@common_processors:0.16.0` and the job updated and ran.

I also tried the job will multiple `filter` ops using both assets and it worked as expected.  I also tried just adding the asset name with no version, but it requires the version, which makes sense.

jsnoble:
Example of real job that can work with the elasticsearch-assets
```json
{
    "name": "test",
    "lifecycle": "once",
    "workers": 1,
    "analytics": true,
    "assets": [
        "elasticsearch:4.0.2",
        "elasticsearch:4.0.5"
    ],
    "apis": [
        {
            "_name": "elasticsearch_sender_api@elasticsearch:4.0.5",
            "index": "op_asset_version_test",
            "preserve_id": true,
            "size": 10000
        },
        {
            "_name": "elasticsearch_reader_api@elasticsearch:4.0.2",
            "index": "ts_test_example-1000",
            "size": 10000,
            "date_field_name": "created",
            "preserve_id": true
        }
    ],
    "operations": [
        {
            "_op": "elasticsearch_reader@elasticsearch:4.0.2",
            "api_name": "elasticsearch_reader_api@elasticsearch:4.0.2",
            "index": "ts_test_example-1000",
            "size": 10000,
            "date_field_name": "created",
            "preserve_id": true
        },
        {
            "_op": "elasticsearch_bulk@elasticsearch:4.0.5",
            "api_name": "elasticsearch_sender_api@elasticsearch:4.0.5",
            "index": "op_asset_version_test",
            "preserve_id": true,
            "size": 10000
        }
    ]
}
```

packages/job-components/src/operation-loader/utils.ts
godber, line 1: This file looks like it is easily unit testable.  I don't see any unit tests, perhaps I am overlooking them.

packages/job-components/src/operation-loader/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    return {
+        name,
+        assetIdentifier,
+        tag
+    };
+}
+
+function isAssetHash(id: string) {
+    return id.length === 40;
```
godber, line 50: This logic results in any `id` length of 40 being treated as an assetHash right?  Can't this be accidentally triggered?  In this case what is `id`?  Is it `elasticsearch` or `elasticsearch:v1.0.1`.  Is there a more concrete way for us  to ensure this string is really a hash?

packages/job-components/src/operation-loader/loader.ts
```diff
@@ -37,109 +40,31 @@ export class OperationLoader {
...
-            if (!fs.existsSync(basePath)) return;
-            if (!subfolders || !subfolders.length) return;
-
-            for (const folder of subfolders) {
-                const folderPath = pathModule.join(basePath, folder);
-                // we check for v3 version of asset
-                if (await this.isBundledOperation(folderPath, name)) {
-                    filePath = folderPath;
-                }
+    async loadProcessor(name: string, assetIds?: string[]): Promise<ProcessorModule> {
```
godber, line 43: Adding a tsdoc comment that describes these functions might be useful.

packages/job-components/src/job-validator.ts
```diff
@@ -57,35 +68,47 @@ export class JobValidator {
...
+            validateApisFns.push((job) => {
                 if (!schema.validateJob) return;
+
+                const originalName = apiConfig._name;
+                const { name } = parseName(originalName);
+
+                // for backwards compatible checks, alter name so it can be found
+                job.apis[index]._name = name;
                 schema.validateJob(job);
+                // revert name back to original
```
godber, line 103: It would be useful to know why we revert back to original name.  What things require it?

packages/job-components/test/operation-loader-spec.ts
```diff
@@ -350,4 +318,377 @@ describe('OperationLoader', () => {
             expect(results.Schema).not.toBeNil();
         });
     });
+
+    describe('namespaced versioned ops', () => {
+        const assetHash1 = 'op_asset_v1.0.0';
```
godber, line 323: Elsewhere we expect asset hashes to be 40 characters in length, is that condition not necessary in this context for some reason?  Perhaps the variable name is just confusing, maybe you mean something more like `assetIdentifier1`.

godber (changes requested):
Consider some of my comments inline, I think I found the parseName tests, I didn't expect them included inside another specfile.

Please try and document the extent of the changes you have made in the documentation.  What all is now possible?  What scenarios are prohibited?  What are tags?

busma13:
I was testing this and it looks like using a plain `apiName` with no `@` or using `apiName:tag` is now broken. I think `adjustNamesByMapping()` is using an undefined `assetIdentifier` when trying to look up the asset hash.

busma13 (approved):
Looks good. There were just a few typos, etc I noticed.

docs/jobs/configuration.md
```diff
@@ -98,4 +98,159 @@ The first operation in the [operations](#operations) list, reads from a particul
     ]
 }
 ```
+<!--Example with multiple apis->
```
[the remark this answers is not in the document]
jsnoble: this is following that pattern in the docs above of how to list code, not sure, just following current doc patterns

e2e/test/cases/assets/simple-spec.ts
```diff
@@ -72,6 +72,24 @@ describe('assets', () => {
             { blocking: true }
         );
 
+        /*
```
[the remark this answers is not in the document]
jsnoble: same as above

packages/job-components/src/job-validator.ts
jsnoble: we have to revert back due to issues with backwards compatibility because in the assets in checks with a `===` for exact matching of names for post schema validation checks that create api names

Issue #3849, opened 2024-12-02
yarn build failure in CI


busma13:
Occasionally the `yarn build` step will fail in CI, but no errors are shown. I am thinking we could run the command in verbose mode so we can at least see the issue and be able to resolve it in the future.

busma13:
I haven't seen this error since adding logging. We also upgraded to yarn 4. Closing for now.

Issue #3848, opened 2024-12-02
port in use errors in CI


busma13:
We keep getting port in use errors.
I would like to try using `ss` to list all ports in `TIME_WAIT` to see if they are not being shut down fast enough.

The other option is to assign random ports.

Example failures:
https://github.com/terascope/teraslice/actions/runs/11981509936/job/33407933870

```
● Test suite failed to run

    listen EADDRINUSE: address already in use :::55678

      140 |             await new Promise((resolve, reject) => {
      141 |                 // @ts-expect-error
    > 142 |                 this.app.listen(this.port, (err: Error) => {
          |                          ^
      143 |                     if (err) {
      144 |                         reject(err);
      145 |                         return;

      at Function.apply [as listen] (node_modules/express/lib/application.js:635:24)
      at listen (packages/teraslice/src/lib/cluster/services/assets.ts:142:26)
      at AssetsService.initialize (packages/teraslice/src/lib/cluster/services/assets.ts:140:19)
      at Object.<anonymous> (packages/teraslice/test/services/assets-spec.ts:67:9)
```

https://github.com/terascope/teraslice/actions/runs/11941437795/job/33286764202
```
docker: Error response from daemon: driver failed programming external connectivity on endpoint ts_test_kafka (fb60d1c2428b6a3ad031693a716c1897da234a8e33577b500cd45bd8242830ae): Error starting userland proxy: listen tcp4 0.0.0.0:49092: bind: address already in use.
```

busma13:
I have found port related errors in 3 different spots:
- assets service
- teraslice messaging server
- docker run of service images

These 3 different areas may need to be handled separately.

busma13:
The teraslice package has a `findPort()` function that some of the tests use. It returns a random open port. I am adding this to `assets-spec.ts`, which should resolve the first list item in my pervious comment.

busma13:
https://github.com/terascope/teraslice/actions/runs/11921771709/job/33226709312

This failure uses the findPort() function. I am guessing multiple tests run this function at the same time and there end up being collisions. We use `porty` to test if the port is open, but it looks like we start at the same port for all these tests and move up until `porty` says one is open. Maybe 2 tests get assigned the same `open` port at the same time.

busma13:
`findPort() allows you to set a port range, so I am giving each test a unique range to prevent collisions.

godber:
Or has the added netstat command helped narrow trouble down.

sotojn:
[...]

PR #3854, opened 2024-12-05
[teraslice, scripts] Ci test error logging and potential fixes

busma13:
- Adds the `--verbose` flag to the root package.json `build` script and removes `--silent` from `YARN_SETUP_ARGS`. This will hopefully reveal why the script fails occasionally. ref: #3849
- Replace unmaintained `porty` library with `get-port` library.
- Within the teraslice package, many tests use `findPort()` to get a random port to start a server on. The implementation did not actually return a random port and made conflicts more likely. Update `findPort()` to use `get-port`, increase the port range, and get a truly random port in that range. ref: #3848 
- When starting services, log all TCP ports just before calling `docker run`. This should give us more information about why the `docker run` commands occasionally fail with a `port already in use` error. ref: #3848


godber (approved):

packages/teraslice-messaging/src/messenger/server.ts
```diff
@@ -94,8 +93,31 @@ export class Server extends Core {
...
+
+                portTestServer.once('listening', () => {
+                    portTestServer.close(() => {
+                        return resolve(true);
+                    });
+                });
+
+                portTestServer.listen(port);
+            });
+        };
```
godber, line 117: Are `error` and `listening` the only two events here?  Is one of these guaranteed to be emitted?  

If I understand whats going on correctly here, you try listening on the port you mean to use, then you expect to close that port when you get the `error` or `listening` events ... I am trying to imagine cases where that won't happen.

godber (approved):
see me comments and lets discuss

godber (approved):

PR #3863, opened 2024-12-09
Teraslice external dependencies

jsnoble:
move deps to utils


packages/teraslice/test/services/messaging-spec.ts
```diff
@@ -123,7 +122,6 @@ describe('messaging module', () => {
     beforeEach(() => {
         firstWorkerMsg = null;
         secondWorkerMsg = null;
-        _.omitBy(connected);
```
busma13, line 126: What changed to make this unnecessary?

busma13 (approved):
Beside the one change to a test I don't understand this looks good.

packages/teraslice/test/services/messaging-spec.ts
jsnoble: the usage of omitBy was a silly way to make connected an empty object

Issue #3871, opened 2024-12-11
Node 22.12.0 incompatible with jest


busma13:
https://github.com/jestjs/jest/issues/15363
https://github.com/prettier/prettier/issues/16930

If I am reading these issues correctly, it looks like `jest.config.ts` will fail when it tries to import a package that imports `prettier` if running node v22.12.0 or v23. `xlucene-parser` has `prettier` as a nested dependency.

We need to pin node 22 at `22.11.0` in CI before github starts using node 22.12.0.

The [base-docker-image](https://github.com/terascope/base-docker-image) already contains node 22.12.0 and It is used in teraslice v2.9.1 and v2.9.2. I will test to see if I can find any issues with how these versions function.

godber:
Can you please describe the impact of this issue?

>We need to pin node 22 at 22.11.0 in CI before github starts using node 22.12.0.

Pinning is only a temporary solution to the problem since as you point out, we are already using this version of node.  Unless the p

busma13:
If the CI runners begin using node v22.12.0 as the default node 20 version, all jest test commands will then fail with this error: `Error [ERR_REQUIRE_ASYNC_MODULE]: require() cannot be used on an ESM graph with top-level await. Use import() instead. To see where the top-level await comes from, use --experimental-print-required-tla.`

sotojn:
For some reason node 22.12.0 won't allow top level awaits with the jest config. I've updated the jest configs to not use a top level await and it fixes the issue:

New jest config in all the packages
```
import { fileURLToPath } from 'node:url';
import baseConfig from '../../jest.config.base.js';

const dirPath = fileURLToPath(new URL('.', import.meta.url));

const config = baseConfig(dirPath);

export default config;

```

busma13:
This was not actually related to prettier at all.

busma13:
After removing top level awaits from the `jest.config.ts`s we ran into an issue with how a nested dependency `unicorn-magic@0.3.0` was being imported into jest. The temporary fix is to create a patch of it's package.json.

busma13:
I figured out why we were having issues with `unicorn-magic`. We are overriding the default `transformIgnorePatterns` array in `jest.config.base.js` with an empty array.
The default is `["/node_modules/", "\\.pnp\\.[^\\/]+$"]. Since we removed that we were running our @swc transform on all of our node_modules unnecessarily, and in this case it broke something.

PR #3874, opened 2024-12-12
Fix jest config top level await

busma13:
- Replace top level `await`s in all `jest.config.ts` files with regular imports, as these will throw errors in node 22.12.0.
- Install `patch-package` and `postinstall-postinstall` as root level dev dependencies.
  - `patch-package` allows for packages within node_modules to be patched.
  - `postinstall-postinstall` allows for a postinstall hooks to be run after `yarn remove`, not just `yarn` or `yarn add`.
  -  Add `patch-package` as a postinstall hook so any patches will be applied properly.
- Use `patch-package` to patch the `unicorn-magic@0.3.0` package.json to have a default export compatible with our jest config.
- Copy `patch` directory into dockerfile.
- Remove `doc` from `yarnclean.ci`. This line deletes the `doc` directory in a dependency of `patch-package` during docker build.

ref: #3871


.yarnclean.ci
```diff
@@ -5,7 +5,6 @@ tests
 powered-test
 
 # asset directories
```
busma13, line 7:
```suggestion
# asset directories
docs
```

sotojn (approved):
LGTM

PR #3875, opened 2024-12-12
[scripts] Add tmpfs option to opensearch testing

sotojn:

- Adds `tmpfs` option to opensearch testing in `ts-scripts`
- Bumps `@terascope/scripts` from `v1.7.0` to `v1.7.1`

### DIFF
--- a/packages/scripts/src/helpers/test-runner/services.ts
+++ b/packages/scripts/src/helpers/test-runner/services.ts
@@ -83,8 +83,8 @@ const services: Readonly<Record<Service, Readonly<DockerRunOptions>>> = {
         image: config.OPENSEARCH_DOCKER_IMAGE,
         name: `${config.TEST_NAMESPACE}_${config.OPENSEARCH_NAME}`,
         tmpfs: config.SERVICES_USE_TMPFS
-        ? ['/usr/share/opensearch/data:uid=1000,gid=1000']
-        : undefined,
+            ? ['/usr/share/opensearch/data:uid=1000,gid=1000']
+            : undefined,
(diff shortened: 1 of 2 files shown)


busma13 (approved):
LGTM

packages/scripts/src/helpers/test-runner/services.ts
```diff
@@ -83,6 +83,9 @@ const services: Readonly<Record<Service, Readonly<DockerRunOptions>>> = {
         image: config.OPENSEARCH_DOCKER_IMAGE,
         name: `${config.TEST_NAMESPACE}_${config.OPENSEARCH_NAME}`,
         ports: [`${config.OPENSEARCH_PORT}:${config.OPENSEARCH_PORT}`],
+        tmpfs: config.SERVICES_USE_TMPFS
+            ? ['/usr/share/opensearch/data']
```
lesleydreyer, line 87: ‘/usr/share/opensearch/data:uid=1000,gid=1000’

packages/scripts/src/helpers/test-runner/services.ts
```diff
@@ -83,6 +83,9 @@ const services: Readonly<Record<Service, Readonly<DockerRunOptions>>> = {
...
+            : undefined,
         env: {
             ES_JAVA_OPTS: config.SERVICE_HEAP_OPTS,
```
lesleydreyer, line 90: OPENSEARCH_JAVA_OPTS

lesleydreyer:
the restrained opensearch should use OPENSEARCH_JAVA_OPTS instead too, after that LGTM

PR #3873, opened 2024-12-12
[CI] Resolve warnings in tests

sotojn:
This PR makes the following changes:

- Cleans up test warnings when running our jest test suites
- Adds a patch that resolves an issue with jest version 29
  - https://github.com/jestjs/jest/issues/14513
  - This is resolved in jest 30 but it's still in beta

PR #3880, opened 2024-12-16
[workspace] Set Jest transformIgnorePatterns to default

busma13:
- In `jest.config.base.js` explicitly set `transformIgnorePatterns` to the default (`['/node_modules/', '\\.pnp\\.[^\\/]+$']`), which will skip running transforms on node-modules or pnp files. We were overriding the default with an empty array. Standard practice is to only transform specific dependencies if there is a need.
- Remove patch for `unicorn-magic` dependency, as it is no longer being broken by the jest transform.
- Add README to `./patches` explaining how to make a patch using `patch-package`. This also prevents the `patches` directory from being removed if there are no patches, breaking the docker build.

Ref: #3871, #3874


sotojn (approved):
Looks good

jest.config.base.js
```diff
@@ -48,7 +48,10 @@ export default (projectDir) => {
             `<rootDir>/${parentFolder}/*/dist`,
             `<rootDir>/${parentFolder}/teraslice-cli/test/fixtures/`
         ],
-        transformIgnorePatterns: [],
+        // do not run transforms on node_modules or pnp files
+        transformIgnorePatterns: [
```
godber, line 52: With this change is the patch on `unicorn-magic` still necessary?

busma13: No it's not. This PR deletes the patch.

godber: Oh yeah, sorry, I saw the `patches/README.md` was added, but didn't catch that the patch itself was deleted.  So we keep the patch readme and dependencies still?  Just in case?  I think that makes sense.

busma13: #3873 would patch jest, so I figured I would leave all the patching mechanisms that I added previously. I think it makes sense to have everything set up for a quick patch. We can remove this once we upgrade to yarn 4.

godber (approved):

Issue #3869, opened 2024-12-11
[ci] Update CI to use new Ubuntu 24.04


sotojn:
We are getting warnings in the `actions` tab in Github that are letting us know that the latest runners are going to be using Ubuntu 24.04. 

Warning:

> [build-release](https://github.com/terascope/teraslice/actions/runs/12266480280/job/34224692132)
> ubuntu-latest pipelines will use ubuntu-24.04 soon. For more details, see https://github.com/actions/runner-images/issues/10636

I've looked at our most recent CI job and found that the runners are still running `Ubuntu 22.04`
https://github.com/terascope/teraslice/actions/runs/12276444008/job/34253663177#step:1:8

Completion of the runners latest image migration will complete on January 17th, 2025. It's best that we switch to this new version now so we can get ahead of the curve.

godber:
Is there an MR for this?

sotojn:
All the tests seem to be passing with no issues after updating to ubuntu 24.04. I'm thinking now that we validated that it won't cause issues, we can just close the PR and allow github to update the latest tag to the new 24 in January.

godber:
Agreed, please close

PR #3883, opened 2024-12-18
[CI] Add ubuntu-24.04 to CI tests

sotojn:

- Replaces `ubuntu-latest` with `ubuntu-24.04` in CI test suites

Ref to issue #3869


sotojn:
Closing because this was a test to see if ubuntu 24.04 would cause issues. This is prove it doesn't.

PR #3885, opened 2024-12-18
[all] Update to TypeScript v5.7.2, remove "suppressImplicitAnyIndexErrors" rule

busma13:
- Update `typescript` from v5.2.2 to v5.7.2
- remove `suppressExplicitAnyIndexErrors` rule from all `tsconfig.json` files
- fix all typescript errors associated with removing the above rule
- add missing `@types/gc-stats` v1.4.3 dev dependency
- bump teraslice to v2.11.0

ref: #3772

### DIFF
--- a/packages/teraslice/src/interfaces.ts
+++ b/packages/teraslice/src/interfaces.ts
@@ -29,12 +29,18 @@ export interface ClusterMasterContext extends Context {
 }
 
 export enum ProcessAssignment {
+    node_master = 'node_master',
     cluster_master = 'cluster_master',
     assets_service = 'assets_service',
     execution_controller = 'execution_controller',
     worker = 'worker'
 
 }
+
+export function isProcessAssignment(value: string): value is ProcessAssignment {
+    return Object.values(ProcessAssignment).includes(value as ProcessAssignment);
+}
+
 interface BaseWorkerNode {
     worker_id: number;
     pid: number;
@@ -99,3 +105,10 @@ export interface ControllerStats extends ExecutionAnalytics {
     job_id: string;
     name: string;
 }
+
+export type MessagingConfigOptions = {
+    [assignment in ProcessAssignment]: {
+        networkClient: boolean;
+        ipcClient: boolean;
+    };
+};
--- a/packages/teraslice/src/interfaces.ts
+++ b/packages/teraslice/src/interfaces.ts
@@ -38,7 +38,7 @@ export enum ProcessAssignment {
 }
 
 export function isProcessAssignment(value: string): value is ProcessAssignment {
-    return Object.values(ProcessAssignment).includes(value as ProcessAssignment);
+    return value in ProcessAssignment;
 }
 
 interface BaseWorkerNode {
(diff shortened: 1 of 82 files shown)


packages/teraslice/src/interfaces.ts
```diff
@@ -29,12 +29,18 @@ export interface ClusterMasterContext extends Context {
...
+    node_master = 'node_master',
     cluster_master = 'cluster_master',
     assets_service = 'assets_service',
     execution_controller = 'execution_controller',
     worker = 'worker'
 
 }
+
+export function isProcessAssignment(value: string): value is ProcessAssignment {
+    return Object.values(ProcessAssignment).includes(value as ProcessAssignment);
```
jsnoble, line 41: you don't need to break down the object each time and check with an includes, a simple `has` check would be sufficent

packages/teraslice/src/lib/cluster/services/api.ts
```diff
@@ -135,24 +136,51 @@ export class ApiService {
         throw error;
     }
 
+    private queryToSearchParams(query: any) {
+        const searchParams: Record<string, string | number | boolean> = {};
+
+        for (const [key, value] of Object.entries(query)) {
```
jsnoble, line 142: This is pretty inefficient checking

packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetesV2/index.ts
```diff
@@ -12,6 +12,7 @@ import { ResourceType } from './interfaces.js';
 import { K8sJobResource } from './k8sJobResource.js';
 import { K8sServiceResource } from './k8sServiceResource.js';
 import { K8sDeploymentResource } from './k8sDeploymentResource.js';
+import { ClusterState } from '@terascope/types';
```
jsnoble, line 15: import from libraries are placed higher than local imports

packages/teraslice/src/lib/cluster/services/cluster/backends/native/messaging.ts
```diff
@@ -334,14 +346,25 @@ export class Messaging {
...
+            if (typeof server === 'string' || typeof server === 'number') {
+                // test server
+                this.io = socketIOServer(server, {
+                    path: '/native-clustering',
+                    pingTimeout: this.configTimeout,
+                    pingInterval: this.configTimeout + this.networkLatencyBuffer,
+                    perMessageDeflate: false,
+                    serveClient: false,
+                });
+            } else {
```
jsnoble, line 358: why is there an if statment?

packages/types/src/xlucene-interfaces.ts
```diff
@@ -14,6 +14,10 @@ export enum xLuceneFieldType {
     Number = 'number',
 }
 
+export function isXLuceneFieldType(value: string): value is xLuceneFieldType {
+    return Object.values(xLuceneFieldType).includes(value as xLuceneFieldType);
```
jsnoble, line 18: This should not be parsed everytime, a simple `in` check would be sufficient

packages/teraslice/src/lib/cluster/services/api.ts
busma13: I will write an issue for improving types on this whole file.

packages/teraslice/src/lib/cluster/services/cluster/backends/native/messaging.ts
busma13: I refactored this, pulling the if statement out into it's own function, but the if was still required to narrow the type down.

Issue #3852, opened 2024-12-05
Set prom_metrics_display_url automatically


briend:
Previously in the Helm Chart, we were setting `TERASLICE_DISPLAY_URL` env variable manually to whatever we configured the ingress for.  It would be convenient if `prom_metrics_display_url` for the new internal metrics server was configured automatically for us, and include a trailing `/` to preserve the behavior of the old exporter.

Basically[ some logic in the templating](https://github.com/terascope/teraslice/blob/7806a1a4d148c4f77458708230f98a3d9bb4c931/helm/teraslice/templates/_helpers.tpl#L127-L129) that would do something like this:

- If ingress is enabled, pick the first host in the `hosts` array and set `prom_metrics_display_url` to that, but also set http(s) appropriately, and add a trailing `/` if necessary.  ex `https://teraslice.local/`
- Otherwise, if ingress is not enabled, leave `prom_metrics_display_url` unset.
- if `prom_metrics_display_url` is set by user, use that as-is without any modification.

godber:
Completed in PR #3886

PR #3886, opened 2024-12-19
[Chart] Set prom_metrics_display_url automatically

sotojn:

- Adds feature to helm chart template to support dynamic configuration of `prom_metrics_display_url`
  - If the user explicitly sets `prom_metrics_display_url` in the terafoundation config, the provided value is used without modification (except making sure a trailing slash is appended if missing). Ex: `http://teraslice.local` will end up as `http://teraslice.local/` in the configmap.
  - When `prom_metrics_display_url` is not set, the URL is dynamically constructed based on if the `ingress.enabled` configuration is `true` or `false`
    - If `true`, it will make a url based off of the value in `.Values.ingress.hosts`.
      - It will also add a `https://` prefix if `ingress.tls` is defined and non-empty. Otherwise will use `http://` prefix
    - If `false` it will leave `prom_metrics_display_url` unset

Ref to issue #3852

### DIFF
--- a/helm/teraslice/templates/_helpers.tpl
+++ b/helm/teraslice/templates/_helpers.tpl
@@ -129,8 +129,12 @@ terafoundation:
   {{- toYaml $filtered | nindent 2 }}
   {{- if hasKey . "prom_metrics_display_url" }}
   prom_metrics_display_url: {{ .prom_metrics_display_url }}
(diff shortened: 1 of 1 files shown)


briend:
Looks good to me. I tested it out with `helm template` and verified `prom_metrics_display_url` looked correct for each scenario

briend (approved):

PR #3893, opened 2025-01-07
Re-enable docker-compose-js and types unit tests

busma13:
- bump docker-compose-js from v1.4.2 to v1.4.3
- re-enable unit tests for docker-compose-js and types packages
- update docker-compose-js tests to account for changes to package since the tests were disabled


jsnoble (approved):
LGTM

PR #3896, opened 2025-01-09
Fix CI issues after yarn 4 update

sotojn:

- Fixes broken ci piplines that publish npm packages after a merge to master
  - This is caused by scripts using the old yarn 1 sub-command to publish npm packages which yarn 4 doesn't support
- bump: (minor) @terascope/scripts@1.9.0


busma13 (approved):
LGTM

Issue #3890, opened 2025-01-02
Publish the Teraslice Helm Chart


godber:
We introduced a helm chart a few months ago.  This chart needs to be published somewhere.  There are at least two options:

* Publish to github pages
  * does our existing documentation complicate this?
  * I have a slight preference for this since it doesn't introduce a new third party into our ecosystem
  * This may be useful https://github.com/helm/chart-releaser-action, also, both @mkhpalm and I have implemented this in github actions somewhere.
  * The charts could go into `ghcr.io`
* ArtifactHub
  * this would be an option if necessary

Edit 1: Added ghcr option.

sotojn:
As of now, I'm uploading the helm chart to GitHub Pages. Our current chart version is `v0.8.12`, and the teraslice version is `v2.12.2`. After some thought, I think we should align the chart’s major version with the major version of teraslice.

For example, if a user wants to deploy Teraslice `v2.x.x`, they would use the latest chart version `v2.x.x`. This ensures compatibility while allowing the chart’s minor and patch versions to evolve independently for features and fixes that may or may not directly relate to teraslice.

It’s worth noting that Teraslice receives frequent version changes, many of which do not affect the chart at all. This approach keeps the chart versioning manageable and avoids unnecessary updates when Teraslice changes don’t impact the chart.

This strategy is similar to how Elasticsearch manages its Helm charts: the major version of the chart matches the major version of Elasticsearch, ensuring compatibility between the two.

A great example of this would come from how elasticsearch handles their chart versioning:
https://artifacthub.io/packages/helm/elastic/elasticsearch#requirements

https://github.com/terascope/teraslice/blob/b36fe301243cd65381749a70c3c269183d060ef2/helm/teraslice/Chart.yaml#L1-L11

godber:
This is completed, some basic details on the charts can be found here:

* https://github.com/terascope/teraslice/pkgs/container/teraslice-chart
* https://terascope.github.io/teraslice/charts/

PR #3915, opened 2025-01-22
[teraslice-chart] Add Teraslice Helm Chart Release Workflow

sotojn:

- Adds initial release workflow for the Teraslice chart. 
  - This workflow will check `ghcr` for the listed `version` in the `Chart.yaml` to see if a tag already exists
    - If it exists, it will stop and move on to building docs
    - If not, ci will package the new chart in a tarball file and update the `index.yaml` file with the change and commit the change to the `index.yaml` to the master branch. It will then upload the tarball as an OCI artifact to the `terascope` gitHub container registry.
  - In both scenarios it will generate an `index.html` file and append these files in the `/charts` directory of the documentation build static files.
  - These actions are triggered on every merge to the master branch
- Updates the charts default values to use the latest teraslice version and node version

Ref to issue #3890


busma13 (changes requested):
Found a few typos

.github/workflows/publish-master.yml
```diff
@@ -102,10 +102,97 @@ jobs:
...
+      - name: Authenticate Helm to GHCR
+        run: |
+          echo "${{ secrets.GITHUB_TOKEN }}" | helm registry login ghcr.io --username $GITHUB_ACTOR --password-stdin
+
+
+      - name: Extract Helm Chart Version
+        id: extract-version
+        run: |
+          CHART_VERSION=$(yq eval '.version' ./helm/teraslice/Chart.yaml)
+          echo "CHART_VERSION=$CHART_VERSION" >> $GITHUB_ENV
```
busma13, line 134: Could we use GITHUB_OUTPUTS instead? I think the validation is better.

.github/workflows/publish-master.yml
```diff
@@ -102,10 +102,97 @@ jobs:
...
+      - name: Publish Helm Chart to GHCR
+        if: env.version_exists == 'false'
+        run: |
+          cd ./helm/helm-repo-site
+          echo $(ls)
+          mv ./teraslice-${{ env.CHART_VERSION }}.tgz teraslice-chart-${{ env.CHART_VERSION }}.tgz
+          echo $(ls)
+          CHART_FILE=$(ls teraslice-chart-${{ env.CHART_VERSION }}.tgz)
+          echo "Publishing $CHART_FILE to GHCR..."
+          helm push $CHART_FILE oci://ghcr.io/terascope --debug
```
godber, line 190: Should this line have the project name?

```suggestion
          helm push $CHART_FILE oci://ghcr.io/terascope/teraslice --debug
```

It seems unexpected that it's the organization.

sotojn: Github container registries are only name-spaced by either an organization or a user account. For example the `base-image` docker image lives in the organization namespace along side the `teraslice` image. They're linked to their repos because of the workflows that publish them. 

https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#pushing-container-images

> When you push a container image from the command line, the image is not linked to a repository by default. This is the case even if you tag the image with a namespace that matches the name of the repository, such as ghcr.io/octocat/my-repo:latest.
> 
> The easiest way to connect a repository to a container package is to publish the package from a workflow using ${{secrets.GITHUB_TOKEN}}, as the repository that contains the workflow is linked automatically. Note that the GITHUB_TOKEN will not have permission to push the package if you have previously pushed a package to the same namespace, but have not connected the package to the repository.
> 
> To connect a repository when publishing an image from the command line, and to ensure your GITHUB_TOKEN has appropriate permissions when using a GitHub Actions workflow, we recommend adding the label org.opencontainers.image.source to your Dockerfile. For more information, see “[Labelling container images](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#labelling-container-images)” in this article and “[Publishing and installing a package with GitHub Actions](https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions).” 

This can be verified using the gitHub api to request all packages that live in the namespace.
https://docs.github.com/en/rest/packages/packages?apiVersion=2022-11-28#list-packages-for-an-organization

Basically when ci pushes the tarball, it will set the repo to teraslice by default.

godber (approved):

busma13 (approved):
LGTM

Issue #3923, opened 2025-01-27
JSON parser errs when encountering unescaped unicode characters


ciorg:
Started seeing issues  where unescaped unicode characters cause issues when parsing from buffer to JSON from the kafka reader.  This started happening after the upgrade to node 22, though I'm not sure if's it's related to node 22 or something new in the data.

The error is: 
```
TSError: Failure to parse buffer, SyntaxError: Bad control character in string literal in JSON at position XXX
```

Example of value causing the issue:
```json
{
"FIELD1": "TEXT",
"FIELD2": "SOME TEXT HERE\u0000\u0000\"
}
```

The error originates from the `fromBuffer` function [here](https://github.com/terascope/teraslice/blob/e9a2afe196b98f4588a3b5b9a4c0dd1c06bdefc0/packages/utils/src/entities/data-entity.ts#L109C4-L126C6)

One solution could be for the JSON parser function to check for unicode characters and then escape the characters before parsing to JSON.  The same behavior would need to be added to the `json_parser` processor in standard assets as well, but the `json_parser` should probably use the same function as the kafka_reader.

ciorg:
Decided just to remove the null unicode characters, as that's all we're seeing right now.  If there are more unicode characters or if it becomes a bigger issues than the upstream provider needs to ensure they are sending valid JSON.

ciorg:
fixed with pr #3927

PR #3927, opened 2025-01-28
Parse json escape unicode

ciorg:
* removed unicode null values on json parse


ciorg:
this pr addresses this issue - https://github.com/terascope/teraslice/issues/3923

PR #3929, opened 2025-01-29
[e2e, scripts] Start K8s e2e from helmfile

busma13:
This PR makes the following changes:
- creates a `helm` directory in `e2e` with the following:
  - helmfile.yaml
  - values.yaml
  - modified elastic/elasticsearch-6.8.9 chart
  - modified elastic/elasticsearch-7.9.3 chart
  - modified confluentinc/cp-helm-charts-0.6.1 chart (kafka and zookeeper)
  - go templates for teraslice and services
- update the teraslice helm chart to include the `autoload_directory`
- update `ts-scripts test` to start k8s e2e tests via helmfile using the -use-helmfile` flag
- update e2e to use helmfile commands instead of the k8s api when started via helm
- update default service versions for tests
  - kafka version from `3.5` to `3.7`, which is `cp-kafka:v7.7.2`
  - opensearch1 from `1.3.10` to `1.3.11`
  - opensearch2 from `2.8.0` to `2.15.0`
- k8s e2e tests now run on opensearch2, not elasticsearch7
- update mappings between kafka and cp-kafka versions
- bump e2e version from `0.7.1` to `0.8.0`
- bump scripts version from `1.9.3` to `1.10.0`

ref: #3931

PR #3941, opened 2025-02-03
[e2e] update kind Configs used in k8s-env to put teraslice port mapping first.

busma13:
A change in #3929 required a change in the kind configs for the `k8s-env` command, but I forgot to make that change.


sotojn (approved):
Looks good

PR #3943, opened 2025-02-04
run yarn setup at end of bump commands

busma13:
- run `yarn setup` after the `ts-scripts bump` command modifies all `package.json` dependency versions. This ensures that the `yarn.lock` will be synced when the branch is pushed.
- bump scripts from 1.10.0 to 1.10.1

### DIFF
--- a/e2e/package.json
+++ b/e2e/package.json
@@ -45,7 +45,7 @@
         "ms": "~2.1.3"
     },
     "devDependencies": {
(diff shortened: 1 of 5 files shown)


godber (approved):
Looks good to me, but I am probably the worst person to ask.

sotojn (approved):
Looks good

Issue #3945, opened 2025-02-04
Track down Memory Leak in Teraslice v2.9.2


godber:
About a month ago we rolled out Teraslice `v2.9.2` into our production environment.  After about 3 or 4 weeks we started having Teraslice workers OOM, due to a newly introduced but very slow memory leak.

The things that changed in the jobs were:

```
teraslice:v2.6.3-nodev22.9.0  ---->  teraslice:v2.9.2-nodev22.12.0
kafka:5.1.0                   ---->  kafka:5.4.0
elasticsearch:4.0.3           ---->  elasticsearch:4.0.5
```

We've been tracking this internally so there's more information that in this issue.  I did some internal testing and rolled back the kafka asset to 5.1.0 and it didn't change the situation, the leak still existed (it takes about 18-24 hours to conclusively see the leak).

Here are the relevant versions from the latest releases:

| teraslice version | base image version | terafoundation_kafka_connector version | node-rdkafka version | librdkafka version | node 22 version |
|-------------------|--------------------|:--------------------------------------:|----------------------|--------------------|-----------------|
| v2.12.3 | node-base:22.13.0 | v1.2.1 | v3.2.1 | v2.6.1 | v22.13.0 |
| v2.12.2 | node-base:22.13.0 | v1.2.1 | v3.2.1 | v2.6.1 | v22.13.0 |
| v2.12.1 | node-base:22.13.0 | v1.2.1 | v3.2.1 | v2.6.1 | v22.13.0 |
| v2.12.0 | node-base:22.12.0 | v1.2.1 | v3.2.1 | v2.6.1 | v22.12.0 |
| v2.11.0 | node-base:22.12.0 | v1.2.1 | v3.2.1 | v2.6.1 | v22.12.0 |
| v2.10.0 | node-base:22.12.0 | v1.2.1 | v3.2.1 | v2.6.1 | v22.12.0 |
| v2.9.2 | node-base:22.12.0 | v1.2.1 | v3.2.1 | v2.6.1 | v22.12.0 |
| v2.9.1 | node-base:22.12.0 | v1.2.1 | v3.2.1 | v2.6.1 | v22.12.0 |
| v2.9.0 | node-base:22.11.0 | v1.2.0 | v3.2.0 | v2.6.0 | v22.11.0 |
| v2.8.0 | node-base:22.9.0 | v1.0.0 | v3.1.1 | v2.5.3 | v22.9.0 |
[...]

PR #3952, opened 2025-02-05
[teraslice-messaging] fix memory leak caused by abortController

busma13:
-bump teraslice-messaging from 1.10.3 to 1.10.4
- Remove the `abortController` from the execution-controller client
- Create an `abortController` in the core client `handleSendResponse` function if `sendAbortSignal` is set to true. This is only the case for when the execution-controller client sends the `worker:slice:complete` event. On a `server:shutdown` event the `abortController` will abort awaiting a server response so it can shutdown properly.  

ref: #3945

### DIFF
--- a/packages/teraslice-messaging/package.json
+++ b/packages/teraslice-messaging/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@terascope/teraslice-messaging",
     "displayName": "Teraslice Messaging",
-    "version": "1.10.3",
+    "version": "1.10.4",
     "description": "An internal teraslice messaging library using socket.io",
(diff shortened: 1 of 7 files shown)


busma13:
Using the promMetrics exporter I was able to see the resident memory of the worker increasing from ~225MB at. first scrape to ~367MB after 48000 slices on the current master branch. Using this branch the memory increased from ~230MB at first scrape to ~232MB after 72000 slices.

godber (approved):

godber:
Unfortunately I didn't catch until now that the teraslice version itself was not bumped.  Doesn't bumping `teraslice-messaging` result in `teraslice` being bumped?

PR #3967, opened 2025-02-13
Encrypted kafka

busma13:
TBD


busma13:
This PR encrypted the confluent/cp-kafka image. We are now using the apache/kafka image.

Issue #3980, opened 2025-02-28
asset pulls get rate-limited in tests


busma13:
If too many jobs in our test suite use `@terascope/fetch-github-release` to download assets at the same time our github account gets rate-limited, causing the test job to fail.

We could do a couple of things:
- cache the assets in github in an automated workflow(daily?) and pull from the cache in each job
- run one job at the start of each workflow run that pulls assets and passes them into all the jobs that need them

godber:
Does `fetch-github-release` have a local caching mechanism?

busma13:
> Does `fetch-github-release` have a local caching mechanism?

No it doesn't. I don't believe this would help in CI since all jobs run in separate runners, so they all have their own instance of `fetch-github-release`

busma13:
Or do you mean that `fetch-github-release` will check a specified cache before making a request to github?

PR #3992, opened 2025-03-10
[CI, e2e] Cache assets bundles in e2e workflows

sotojn:

- Adds a pipeline to ci that will manage and cache asset bundles 
  - This will resolve an issue getting rate limited by github #3980 
- Adds ci job that logs the Github rate limit for the runner token
- bump: (**patch**) teraslice-cli@2.10.5 
- bump: (**patch**) teraslice-test-harness@1.3.3


e2e/scripts/downloadAssets.js
```diff
@@ -0,0 +1,150 @@
...
+        filterAsset,
+        true, // Keep assets zipped
+        false, // Don't disable logging
+        dryRun // dry run is true
+    ))
+);
+const jsonAssetList = await Promise.all(promises);
+
+// If we are dry-running we will generate a list
+if (dryRun) {
```
godber, line 143: This reads a little funny.  You seem to mix `dry-run` and `generate-list`  But really your command just has two modes `download` and `generate-list`.  I think you just need to rename `dryRun`  or something.

```
Usage:
  node downloadAssets.js download       # Download assets into /tmp/teraslice_assets
  node downloadAssets.js generate-list  # Generate a txt file list of assets
```

Lets talk about this.

Is there a `dry-run` options when `download` is run?  Or do we just have two different modes for the command?

sotojn: Yeah I think I should just get rid of  the `dryrun` boolean and do something more straight forward. Instead of this:

```javascript
const userCommand = process.argv[2];
const dryRun = userCommand === 'download' ? false : true;
const promises = defaultAssetBundles.map(({ repo }) => downloadWithDelayedRetry(
    () => downloadRelease(
        'terascope',
        repo,
        ASSET_BUNDLES_PATH,
        filterRelease,
        filterAsset,
        true, // Keep assets zipped
        false, // Don't disable logging
        dryRun // dry run is true
    ))
);

// If we are dry-running we will generate a list
if (dryRun) {
    /// code that generates-list
}
```

I should do this:

```javascript
const userCommand = process.argv[2];
const promises = defaultAssetBundles.map(({ repo }) => downloadWithDelayedRetry(
    () => downloadRelease(
        'terascope',
        repo,
        ASSET_BUNDLES_PATH,
        filterRelease,
        filterAsset,
        true, // Keep assets zipped
        false, // Don't disable logging
        userCommand === 'download' ? false : true // if the user requested a download we don't want to dryrun
    ))
);

if (userCommand === 'generate-list' ) {
    /// code that generates-list
}
```

godber (approved):

PR #3997, opened 2025-03-20
[e2e, scripts, helm, dockerfile] Small fixes

busma13:
This PR makes the following updates:
- Remove unneeded `unpkg/clipboardJS` import from helm repo `index.html`
- fix order of checks for the `kafka-spec` encryption test
- fix dockerfile warnings when declaring ENVs
- `dockerNetworkExists()` function was using `execaCommand()` with a `|`, so it needs to set the `shell` option to true
- pin mkcert version within test.yaml


sotojn (approved):
LGTM

godber (approved):

Issue #3959, opened 2025-02-10
Update @kubernetes/client-node to 1.0.0


busma13:
@kubernetes/client-node released version 1.0.0, which has some breaking changes. We need to update and fix anything that breaks.

busma13:
We are waiting until a PR is merged to fix patch middleware. See the notes one #3998

PR #3998, opened 2025-03-20
[teraslice, scripts] upgrade kubernetes/client-node to v1.1.0

busma13:
- update `kubernetes/client-node` from v0.22.3 to v1.1.0 in teraslice and scripts packages
- modify k8s function calls to use parameter objects
- remove unneeded interfaces/types since API responses no longer use a wrapper object.
- update error handling since some 400 level responses from kubectl now throw an `ApiException` error internally.
- `k8s.patch` fix
  - middleware workaround
    - comment from issue with workaround I used: https://github.com/kubernetes-client/javascript/issues/2160#issuecomment-2726005739
    - PR that should fix this: https://github.com/kubernetes-client/javascript/pull/2316
  - switched from a json patch to a merge patch because json patch didn't work with the above fix. Merge patch seems to have some limitations, but we only use the patch function to scale workers, so it's not an issue. When we remove the patch we should switch this back to json patch.

ref: #3959

### DIFF
--- a/e2e/package.json
+++ b/e2e/package.json
@@ -47,7 +47,7 @@
         "ms": "~2.1.3"
     },
     "devDependencies": {
-        "@terascope/scripts": "~1.13.0",
+        "@terascope/scripts": "~1.14.0",
         "@terascope/types": "~1.4.1",
         "@terascope/utils": "~1.7.7",
         "bunyan": "~1.8.15",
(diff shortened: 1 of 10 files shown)


godber:
We will hold off until this PR is merged and the fix released: https://github.com/kubernetes-client/javascript/pull/2316

busma13:
closed in favor of #4029

PR #4006, opened 2025-03-26
[e2e, helm] autoload_directory should be created in e2e helm, not be default in helm chart

busma13:
This PR fixes a mistake that added `autoload_directory` to the default teraslice helm `values.yaml`. This setting is only needed for e2e, so is now set in `e2e/helm/templates/teraslice.yaml.gotmpl`.

The chart is bumped to version 2.3.1

I also added `icon` and `home` links to the chart.


godber (approved):

busma13:
> When we merge this, should we see this update: https://terascope.github.io/helm-charts/

We will if we merge #4003 first

busma13:
I missed the fact that the deployment had autoload volume and volumeMount hardcoded. Fixed in #4008

PR #4007, opened 2025-03-27
[CI, helm] Fix error triggering helm charts workflow, remove '-chart' from chart name

busma13:
- Fixes a logic error in the `publish-master` workflow that prevented the new job that would trigger a workflow in `terascope/helm-charts` from running
- Update the name field in the chart to just be `teraslice`, not `teraslice-chart`
- Bump chart to 2.3.2


sotojn (approved):
LGTM

PR #4013, opened 2025-03-28
Lib file path fix

jsnoble:
- fix teraslice-test-harness file extensions to be able to be importable by node 22 based esm import calls
- fixed some issues in data-mates bench importing json in node 22


busma13 (approved):
LGTM

Issue #3898, opened 2025-01-10
Docker 'bind: address already in use' port errors


busma13:
I've closed #3848 as all other port related errors have been resolved.

We now only get port related issues involving the docker setup. The new logs in the comment below did catch that port 49206 was in `TIME-WAIT` when docker attempted to run the opensearch container. Maybe it is worth randomizing those ports too. Or we could check for `TIME-WAIT` on the ports we need and either wait for them to clear, or use a fallback port at that time. Maybe there is a way to clear a `TIME-WAIT`?

sotojn:
Another occurrence of port issues. This time in the es asset repo:

https://github.com/terascope/elasticsearch-assets/actions/runs/12717423594/job/35453891432

sotojn:
Found a strange one that isn't docker related but still a port in use errror:

https://github.com/terascope/teraslice/actions/runs/14223323243/job/39856712470?pr=4022

busma13:
port error in kafka-assets: https://github.com/terascope/kafka-assets/actions/runs/15739842701/job/44362378967

busma13:
We are also getting docker port errors in our internal project that uses ts-scripts. It increased dramatically when we stopped randomizing ports. The new plan is to use a bash script to get a random port, check that it isn't being used with a `netstat` command, then pass the port in as an env var to `ts-scripts test`. I will remove the `isPortInUse()` check in ts-scripts, as it can occasionally port a checked port into time-wait.

Austin came up with this command to get a list of ports already being used:  

`netstat -tan4 | grep tcp | awk '{print $4}' | tr ':' ' ' | awk '{print $2}' | sort -n | uniq`  

There are differences between the `netstat` commands on MacOS and Linux.
- the delimiter between host and port is different. Linux uses `:` and MacOS uses `.`. We need to keep this in mind while parsing the list.
- `-4` is not a valid flag on mac. In linux it is used to only get ipv4 addresses. It should be fine to include the ipv6 addresses in our command.

This is a modified command that works on both platforms:  

`netstat -tan | grep '^tcp' | awk '{print $4}' | tr '.:' ' ' | awk '{print $NF}' | sort -n | uniq`

busma13:
Looking back, I found at least one instance of a random port being in use since the last "fix". 
https://github.com/terascope/teraslice/actions/runs/16332264292/job/46137433470
I do believe this is happening less frequently, but we haven't had as many teraslice PRs recently.

[...]

PR #4022, opened 2025-04-01
[E2e, utils] Validate service ports when running ci tests

sotojn:

- Adds port validation to all tests that need services
  - If a service port is in use, scripts will now find a random port that isn't in use and assign the new port to said service

Ref to issue #3898

### DIFF
--- a/packages/utils/src/ip.ts
+++ b/packages/utils/src/ip.ts
@@ -7,6 +7,7 @@ import isCidr from 'is-cidr';
 import ipaddr, { IPv4, IPv6 } from 'ipaddr.js';
 import { parseIp, stringifyIp } from 'ip-bigint';
 import ip6addr from 'ip6addr';
+import net from 'node:net';
 import { isString } from './strings.js';
 import {
     toInteger, isNumberLike, toBigIntOrThrow,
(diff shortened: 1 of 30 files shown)


packages/utils/src/ip.ts
```diff
@@ -389,3 +390,47 @@ function _expandIPv6Part(part: string) {
...
+ */
+export function isPortInUse(port: number): Promise<boolean> {
+    return new Promise((resolve) => {
+        const testServer = net.createServer()
+            .once('error', (err: any) => {
+                if (err.code === 'EADDRINUSE') {
+                    resolve(true);
+                } else {
+                    resolve(false);
+                }
```
busma13, line 409: We should probably close the server on error too.

packages/utils/src/ip.ts
```diff
@@ -389,3 +390,47 @@ function _expandIPv6Part(part: string) {
...
+ * @returns A boolean of true if the address is in use or false if not
+ */
+export function isPortInUse(port: number): Promise<boolean> {
+    return new Promise((resolve) => {
+        const testServer = net.createServer()
+            .once('error', (err: any) => {
+                if (err.code === 'EADDRINUSE') {
+                    resolve(true);
+                } else {
+                    resolve(false);
```
busma13, line 408: What are the possible errors besides EADDRINUSE? Do they all mean the port is free?

busma13 (commented):
I have a few questions around `isPortInUse`. Looks good overall.

packages/utils/src/ip.ts
```diff
@@ -389,3 +390,49 @@ function _expandIPv6Part(part: string) {
...
+}
+
+/**
+ * Grabs a random port and checks if it's in use. Will recurse until a random unused
+ * port is found and return it.
+ * @param min The minimum port number to use
+ * @param max The maximum port number to use
+ * @returns A port that is available for use
+ */
+export async function getAvailablePort(min = 10000, max = 60000): Promise<number> {
```
godber, line 429: You add two functions, `isPortInUse` and `getAvailablePort`.  There are no new tests added and codcov reflects that.  Can tests be implemented for this?

You should also consider if the lines added in kind are testable.

godber (changes requested):
If there's an easy win on test coverage, you should consider taking the time to make that change.

packages/utils/src/ip.ts
sotojn: I'll make tests for both functions. As far as the lines added in kind, it's a little more difficult to test as we don't test most of the `kind.ts` file as of right now. 
https://app.codecov.io/gh/terascope/teraslice/blob/e2e-port-in-use-fix-ci/packages%2Fscripts%2Fsrc%2Fhelpers%2Fkind.ts

I'll look into it and see if it's easy or not to get coverage here

godber (approved):

PR #4023, opened 2025-04-02
fix recordExists to return false if no id

lesleydreyer:
- [x] fix record exists to return false if no id passed
- [x] fix a module not found warning

### DIFF
--- a/packages/elasticsearch-store/src/index-model.ts
+++ b/packages/elasticsearch-store/src/index-model.ts
@@ -205,7 +205,7 @@ export abstract class IndexModel<T extends i.IndexModelRecord> extends IndexStor
 
     async recordExists(id: string[] | string, clientId?: number): Promise<boolean> {
         const ids = validateIds(id, 'recordExists');
-        if (!ids.length) return true;
+        if (!ids.length) return false;
 
         const count = await this.countRecords({
             [this.config.id_field!]: ids,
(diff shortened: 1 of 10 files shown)


packages/elasticsearch-store/src/index-model.ts
```diff
@@ -205,7 +205,7 @@ export abstract class IndexModel<T extends i.IndexModelRecord> extends IndexStor
 
     async recordExists(id: string[] | string, clientId?: number): Promise<boolean> {
         const ids = validateIds(id, 'recordExists');
-        if (!ids.length) return true;
+        if (!ids.length) return false;
```
godber, line 208: @jsnoble do you have an idea where this `elasticsearch-store` function `recordExists` gets used?  I am wondering where I might see the impact of this change?  Is this used at all in processing data, or is it mainly state cluster type stuff?

jsnoble: - elasticsearch-store is mainly used in spaces, its the primary base class used in the metadata-manager, nearly every model uses that function
- its also used by the deleteRecord method, which is used by any of the delete commands in the manager, as well as the deployer

godber:
Unfortunately I merged Joseph's PR first, we would have gotten a free version bump on elasticsearch-store from that one.  Do you want to bump it on this PR @lesleydreyer ?

jsnoble (approved):
LGTM

Issue #4043, opened 2025-04-21
[scripts] ts-scripts errors when passing test directory argument within a package dir


sotojn:
## Context

When updating to yarn 4 I made the decision to use yarns `workspace` command to call the scripts binary within all of the other packages. This was because the update to yarn 4 made it so that the node modules no longer resolves dependencies (eventhough they're in there) unless it is explicitly stated in said package json. I orginally was going to add scripts as a dependency to all the packages in the monorepo but it caused circular dependency issues. 

Part of my new solution leveraged scripts directory argument. For example in the `data-mate` directory  the `test` script in the package json uses `../data-mate` as a path to run tests within data-mate as shown below:

```
        "test": "yarn workspace @terascope/scripts ts-scripts test ../data-mate --",
```

## The issue 

Dev users that want to use scripts test directory argument within a package will fail to run correctly.

Example trying to run a specific file in `data-mate`

```
yarn test ./test/data-frame-search-spec.ts
▶  test suite "unit" Initialized timer...
★  star      Running test for @terascope/data-mate
(node:2059) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
Error: Can't find a root directory while resolving a config file path.
Provided path to resolve: ./test/data-frame-search-spec.ts
cwd: /Users/user/Workspace/teraslice
    at resolveConfigPath (/Users/user/Workspace/teraslice/node_modules/@jest/core/node_modules/jest-config/build/resolveConfigPath.js:123:11)
    at readInitialOptions (/Users/user/Workspace/teraslice/node_modules/@jest/core/node_modules/jest-config/build/index.js:397:53)
    at readConfig (/Users/user/Workspace/teraslice/node_modules/@jest/core/node_modules/jest-config/build/index.js:147:54)
    at /Users/user/Workspace/teraslice/node_modules/@jest/core/node_modules/jest-config/build/index.js:463:18
    at Array.map (<anonymous>)
    at readConfigs (/Users/user/Workspace/teraslice/node_modules/@jest/core/node_modules/jest-config/build/index.js:457:10)
    at runCLI (/Users/user/Workspace/teraslice/node_modules/@jest/core/build/cli/index.js:152:29)
    at Object.run (/Users/user/Workspace/teraslice/node_modules/jest-cli/build/run.js:130:62)
◼  test suite "unit" Timer run for: 688ms

[...]

PR #4074, opened 2025-06-06
[tests] use scripts bin instead of yarn workspace

sotojn:

- Changes the way ts-scripts is called in all packages in the monorepo
  - This fixes a bug where the developer can't pass `ts-scripts` flags when executing tests

Ref to issue #4043


sotojn:
I've listed some other paths I tried going down before coming to this solution:

https://github.com/terascope/teraslice/issues/4043#issuecomment-2951069289

PR #4088, opened 2025-06-24
[development] Update example helm environment

sotojn:

- Fixes configuration error within example helm that breaks teraslice
- Adds kafka-ui and teraslice utility pod to the example helm environment 
- Adds quick start guide to example helm docs with links and docs to exposed services


sotojn:
I've update the docs with some requested improvements. I've included a note before running the quick start script that tells the user to check the utility/data directory for large files before deploying all the resources because of a known issue. Also Added some comments on the quick start script with some information about excluding `docker build` commands if already built as well as some commented out commands to use if the user would prefer pulling a teraslice image instead.

PR #4090, opened 2025-06-26
Change to not return results if variable is an empty array

lesleydreyer:
This change makes variables that are empty arrays be treated as an empty value that doesn't match anything. Previously it would've either thrown a parsing exception if used in a constant_score query with an empty filter, or return values if joined with an AND/OR, when it should've only returned values if joined with an OR.

- [x] Use match_none if the filter is empty to avoid a parsing exception on an empty filter
- [x] Ensure OR vs AND works correctly when combined with an empty array variable


busma13 (approved):
LGTM

Issue #4018, opened 2025-03-28
[tests] E2e test fails to get asset when using cache


sotojn:
While running tests I noticed that the helmfile encrypted e2e job failed. I ran it two more times with the same result. It couldn't find the elasticsearch-asset in the asset-storage bucket and would fail to generate the test data. It wasn't until I deleted the assets cache and re-ran the job that it passed. Maybe the cache was corrupted? If this problem occurs again, we should pull this cache locally with the git cli and see if there are any issues.

 
- **Attempt 1:** https://github.com/terascope/teraslice/actions/runs/14137473152/job/39612576208?pr=4005
- **Attempt 2:** https://github.com/terascope/teraslice/actions/runs/14137473152/job/39613661920?pr=4005
- **Attempt 3:** https://github.com/terascope/teraslice/actions/runs/14137473152/job/39614046792?pr=4005
- **Attempt 4 Removing cache:** https://github.com/terascope/teraslice/actions/runs/14137473152/job/39616379268?pr=4005

busma13:
An newer test failure where github still has logs: https://github.com/terascope/teraslice/actions/runs/15692228287/job/44217230338

```
▶  global setup Initialized timer...
ℹ  info      Loading asset cache..
⬤  debug     Loaded asset file elasticsearch-v4.3.1-node-22-bundle.zip from cache..
⬤  debug     Loaded asset file elasticsearch-v4.3.1-node-24-bundle.zip from cache..
⬤  debug     Loaded asset file kafka-v5.8.1-node-22-bundle.zip from cache..
⬤  debug     Loaded asset file kafka-v5.8.1-node-24-bundle.zip from cache..
⬤  debug     Loaded asset file standard-v1.5.1-node-22-bundle.zip from cache..
⬤  debug     Loaded asset file standard-v1.5.1-node-24-bundle.zip from cache..
✔  success   Finished loading asset cache!
⬤  debug     Skipping downloading assets in favor of cache..
⚠  warning   Deleting asset elasticsearch-v4.3.1-node-24-bundle.zip because node version does not match 22
⚠  warning   Deleting asset kafka-v5.8.1-node-24-bundle.zip because node version does not match 22
⚠  warning   Deleting asset standard-v1.5.1-node-24-bundle.zip because node version does not match 22
[...]

PR #4094, opened 2025-06-30
[e2e] Specified key not found fix

sotojn:

- Updates CI helmfile binary from `v0.171.0` to `v1.1.2`
- Fixes bug in e2e where assets aren't ready when example data needs to be generated

Ref to issue #4018


e2e/test/teraslice-harness.ts
```diff
@@ -588,6 +589,12 @@ export class TerasliceHarness {
             ]
         } as any;
 
+        signale.info(`Validating assets..`);
+        await pRetry(async () => {
+            await this.validateAssets(requiredAssets);
+        });
```
busma13, line 595: This will retry on an asset that isn't loaded. Should we only retry if the s3 blob is unavailable? That would probably male the code messier though.

busma13 (commented):
Were you going to add a check in job validation?

godber (approved):

PR #4100, opened 2025-07-02
Add search tests to xlucene-translator

lesleydreyer:
We discovered an array case that would've thrown an exception recently, but since we don't test the translated queries against a search engine, we couldn't have known. This sets up a foundation to start adding tests that run against opensearch, and also tests the fix for the exception that threw.


jsnoble (approved):
LGTM

PR #4102, opened 2025-07-03
Opensearch3

busma13:
TBD


packages/elasticsearch-store/src/test-helpers/elasticsearch.ts
```diff
@@ -134,7 +134,7 @@ export async function populateIndex(
 
     await client.indices.create({
         index,
-        waitForActiveShards: 'all',
+        wait_for_active_shards: 'all',
```
busma13, line 137: Our typescript types and non-test code have used snake_case for the last 3 years. It looks like this test file added a camelCase version at some time after that. Snake case has been valid since ES5. https://www.elastic.co/guide/en/elasticsearch/reference/5.0/docs-index_.html#index-wait-for-active-shards

packages/data-types/src/types/v1/domain.ts
```diff
@@ -37,7 +37,7 @@ export default class Domain extends BaseType {
             tokenizer: {
                 domain_tokens: {
                     reverse: 'true',
-                    type: 'PathHierarchy',
+                    type: 'path_hierarchy',
```
busma13, line 40: The snake_case version is present in ES7 (https://www.elastic.co/guide/en/elasticsearch/reference/7.17/analysis-pathhierarchy-tokenizer.html) and OS1 (I couldn't find a reference in the docs, but it is in the code: https://github.com/opensearch-project/OpenSearch/blob/1.3.20/modules/analysis-common/src/main/java/org/opensearch/analysis/common/CommonAnalysisPlugin.java#L380)

busma13:
Looking at codecov for other PRs, it looks like they have the same partial coverage when checking ES/OS versions. I think this is because the version is always 1, 2 or 3. If we don't try a version that doesn't exist that code path is never followed.

Issue #4075, opened 2025-06-09
Reduce Overall Test Time by targetting


godber:
Explore options for reducing the test runtimes by only running tests for relevant sub-packages.

godber:
But changing to it might be a huge hassle.

sotojn:
I've looked into a couple ideas and have a proposed solution. We should create a new sub command within scripts or just a script that can be ran before we trigger the the other tests. For example lets say we had a script that would somehow handle all the logic for what tests need to be ran and which ones don't. Our `test.yaml` ci file would look something like this:

```yaml
  determine-tests-based-off-commit:
    runs-on: ubuntu-latest
    needs: compute-node-version-vars
    steps:
      - name: Check out code
        uses: actions/checkout@v4
      - name: Setup Node 24
        uses: actions/setup-node@v4
        with:
          node-version: 24
          cache: 'yarn'
      - name: Install and build packages
        run: yarn && yarn setup
      - name: Determine tests to run
        id: determine
        # The output of determine-tests would be a json object that looks like this:
          # {
          #   "unit": true,
          #   "e2e": true,
          #   "website": false
          # }
        run: |
          echo "Running determine-tests..."
          JSON=$(yarn determine-tests)

          LINUX_UNIT_TESTS=$(echo "$JSON" | jq -r '.unit')
          E2E=$(echo "$JSON" | jq -r '.e2e')
          WEBSITE=$(echo "$JSON" | jq -r '.webiste')

          echo "unit=$UNIT" >> $GITHUB_OUTPUT
          echo "e2e=$E2E" >> $GITHUB_OUTPUT
          echo "website=$WEBSITE" >> $GITHUB_OUTPUT

  linux-unit-tests:
    if: needs.determine-tests-based-off-commit.outputs.unit == 'true'
    runs-on: ubuntu-latest
    needs: [determine-tests-based-off-commit, compute-node-version-vars]
    ....


  e2e-k8s-tests:
    if: needs.determine-tests-based-off-commit.outputs.e2e == 'true'
    runs-on: ubuntu-latest
    needs: [compute-node-version-vars, verify-build, cache-docker-images, cache-asset-bundles, determine-tests-based-off-commit]
    ...

  e2e-k8s-v2-tests:
    if: needs.determine-tests-based-off-commit.outputs.e2e == 'true'
    runs-on: ubuntu-latest
    ....

  test-website:
    if: needs.determine-tests-based-off-commit.outputs.website == 'true'
[...]

PR #4103, opened 2025-07-03
[testing] Optimize CI tests

sotojn:

- Adds a new script `determine-tests` that uses logic to see what tests should, and shouldn't run
  - At the moment this will check to see if there are only changes to the `docs` directory, to then not run any tests other than website tests
- Add better logic to `all-tests-passed` ci job that will allow skipped tests to be considered `passed`
  - This allows the new optimization logic to skip tests and allow merging. 

Ref to issue #4075

### DIFF
--- a/scripts/determine-tests.js
+++ b/scripts/determine-tests.js
@@ -21,7 +21,7 @@ function getChangedFiles() {
     //         });
     //     }
     // } else {
-    const diffOutput = execSync(`git diff --name-only master`, {
+    const diffOutput = execSync(`git diff --name-only origin/HEAD`, {
         encoding: 'utf8'
     });
     // }
@@ -48,7 +48,7 @@ export function getFileDiff(filePath) {
     //     }
     // } else {
     try {
-        const rawDiff = execSync(`git diff master -- "${filePath}"`, {
+        const rawDiff = execSync(`git diff origin/HEAD -- "${filePath}"`, {
             encoding: 'utf8'
         });
         return parseUnifiedDiff(rawDiff);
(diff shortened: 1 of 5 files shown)


scripts/determine-tests.js
```diff
@@ -0,0 +1,131 @@
+import { execSync } from 'child_process';
+
+// Don't leave console.logs in here
```
busma13, line 3:
```suggestion
```

scripts/determine-tests.js
```diff
@@ -0,0 +1,131 @@
...
+        unit: checkE2eTests(), // For now we do the same check as e2e
+        e2e: checkE2eTests(),
+        website: checkWebsiteTests(),
+    };
+
+    return JSON.stringify(result);
+}
+
+// eslint-disable-next-line no-console
+console.log(determineTestJobs());
```
busma13, line 131: This seems weird. Would this script be better as a command in ts-scripts?

scripts/determine-tests.js
```diff
@@ -0,0 +1,131 @@
...
+        return !changedFiles.every((file) => file.startsWith('docs/'));
+    }
+
+    // function checkUnitTests() {
+    //     // also do later
+    //     return true;
+    // }
+    const result = {
+        unit: checkE2eTests(), // For now we do the same check as e2e
+        e2e: checkE2eTests(),
```
busma13, line 123: We are using the e2e result for integration tests in test.yml. Can we add integration here, or just have `code-tests` (or something like that) since there is just one function being called for both?

busma13 (commented):
Every other file in ./scripts is a shell script. This makes me think this new script is in the wrong place. How much work would it be to add this to ts-scripts?

busma13 (approved):
LGTM

Issue #4098, opened 2025-07-01
docusaurus - add trailing slash by default


busma13:
Many links in our docs are missing trailing `/`. This causes a redirect, which causes algolia not to index the page. We can add the following config to all of our docusaurus configs:
```
    trailingSlash: true,
```
This will ensure the links all have trailing slashes, so no redirects will occur.

PR #4107, opened 2025-07-08
[CI, examples, website] Fix paths to get-random-port script, fix OS/ES defaults, add trailingSlash to doc links

busma13:
- Fix path to `get-random-port.sh script` in several CI jobs
- Fix which ES/OS services are started by default in `examples/helm/values.yaml`
- Add trailingSlash property to docusaurus config.
  - The property will add trailing slashes to documentation links, preventing redirects, which were preventing algolia from crawling some pages.
  - Fix `origin/HEAD` not found in `determine-tests.js` script

ref: #4098, #4113


sotojn (approved):
LGTM

PR #4120, opened 2025-07-16
Draft: Move client from elasticsearch-store to new package opensearch-client

lesleydreyer:
In order to resolve a circular dependency between xlucene-translator and elasticsearch-store, this moves half of elasticsearch-store into a new package - but re-exports everything moved to the new package with a deprecation notice that it moved to the new package. The elasticsearch-store keeps the index store, model, & manager, while the new package has the main client we use, elasticsearch test helpers, and a couple utils. Named it @terascope/opensearch-client instead of @terascope/elasticsearch-client since we use opensearch more these days - but open to better name ideas.


jsnoble (approved):
LGTM once test pass

PR #4122, opened 2025-07-18
WIP: vector support

jsnoble:


packages/data-types/src/types/v1/vector.ts
```diff
@@ -0,0 +1,117 @@
...
+        if (!isInteger(dimension)) {
+            throw new Error(`${this.field} must have an dimension property set to an integer`);
+        }
+
+        if (!isValidSpaceType(space_type)) {
+            throw new Error(`${this.field} must have a valid space_type property`);
+        }
+
+        if (!validAlgorithms(name)) {
+            throw new Error(`${this.field} must have a correct name property`);
```
lesleydreyer, line 40: maybe "method" or "engine algorithm" instead of "name"

packages/data-types/src/types/v1/vector.ts
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        if (!isValidEngine(engine)) {
+            throw new Error(`${this.field} must have a correct engine property`);
+        }
+
+        if (engine === 'lucene' && name === 'ivf') {
+            throw new Error(`${this.field} have conflicted configs, engine "lucene" cannot be paired with name "ivf"`);
```
lesleydreyer, line 48: maybe change to either "${field} has conflicting values, engine...." or  "Engine "lucene" cannot be paired with method "ivf" on vector data types"

packages/xlucene-parser/src/parser.ts
```diff
@@ -246,6 +246,32 @@ export class Parser {
         walkNode(this.ast);
     }
 
+    walkAST(cb: (node: i.Node) => void) {
```
lesleydreyer, line 249: maybe a comment above walkAST in case hovering over what it can be used for since don't think it covers all node types, unless it does then skip

packages/types/src/elasticsearch-interfaces.ts
```diff
@@ -24,6 +24,17 @@ export type BoolQuery = {
 
 export type BoolQueryTypes = 'filter' | 'should' | 'must_not';
 
+export interface KNNQuery {
+    knn: {
+        [field: string]: {
+            vector: number[];
+            k: number;
+            filter?: AnyQuery[];
```
lesleydreyer, line 32: we don't set filter to an array so maybe filter?: AnyQuery or AnyQuery|AnyQuery[]

packages/xlucene-translator/src/translator/utils.ts
```diff
@@ -2,375 +2,513 @@ import {
...
-        } else {
-            topLevelQuery = {
-                constant_score: {
-                    filter
+    return regexQuery;
+}
+
+function buildKNNQuery(
+    parser: Parser,
+    context: QueryContext,
```
lesleydreyer, line 404: maybe some types like this and think you could get rid of the let buildquery since you already define the knnquery variable, and get rid of the word conjunction in the else statement

```
let knnQuery: KNNQuery = {
        knn: {}
    };
    const must: (AnyQuery | BoolQuery)[] = [];

    parser.walkAST((node: Node) => {
        if (isFunctionNode(node) && node.name === 'knn') {
            const { variables, type_config } = context;
            const instance = initFunction({ node, variables, type_config });
            const { query } = instance.toElasticsearchQuery(
                getTermField(node),
                context
            );
            knnQuery = query as KNNQuery;
        } else if (isTermType(node)) {
            const query = buildTermLevelQuery(node, context);
            if (query) must.push(query);
        } else if (isNegation(node)) {
            const query = buildNegationQuery(node, context);
            if (query) must.push(query);
        } else if (isExists(node)) {
            must.push(buildExistsQuery(node));
        }
    });

    if (must.length) {
        for (const [field, config] of Object.entries(knnQuery.knn)) {
            const knnConjunction = {
                ...config,
                filter: {
                    bool: {
                        should: [
                            {
                                bool: {
                                    filter: [
                                        ...must
                                    ]
                                }
                            }
                        ]
                    }
                }
            };
            knnQuery.knn[field] = knnConjunction;
        }
    } else {
        for (const [field, config] of Object.entries(knnQuery.knn)) {
            knnQuery.knn[field] = { ...config };
        }
    }

    return knnQuery;
```

packages/xlucene-translator/test/vector-spec.ts
```diff
@@ -0,0 +1,156 @@
...
+
+const { makeClient, populateIndex, cleanupIndex } = ElasticsearchTestHelpers;
+
+describe('vector searches', () => {
+    const index = 'vector_search_';
+    let client: Client;
+    let clientMetadata: ClientMetadata | undefined;
+
+    const dataType = new DataType({
+        version: LATEST_VERSION,
```
lesleydreyer, line 14: not sure - maybe add a couple cases to the test/cases/translator folder too, even though this test is more important

packages/xlucene-translator/test/vector-spec.ts
```diff
@@ -0,0 +1,156 @@
...
+import { QueryAccess } from '../src/query-access/index.js';
+
+const { makeClient, populateIndex, cleanupIndex } = ElasticsearchTestHelpers;
+
+describe('vector searches', () => {
+    const index = 'vector_search_';
+    let client: Client;
+    let clientMetadata: ClientMetadata | undefined;
+
+    const dataType = new DataType({
```
lesleydreyer, line 13: you might be able to... 
- remove the @terascope/data-types from xlucene-translator package.json 
- move makeRecordDataType from elasticsearch-store to @terascope/opensearch-client ElasticsearchTestHelpers instead
- then just make the data type with makeRecordDataType instead of having to make xlucene translator dependent on @terascope/data-types

lesleydreyer:
lgtm - added a couple suggestions but feel free to ignore if you disagree

packages/data-types/src/types/v1/vector.ts
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        if (distribution === ElasticsearchDistribution.opensearch) {
+            if (majorVersion === 1 || (majorVersion === 2 && minorVersion < 10)) {
+                throw new Error('Vector datatypes are not supported with Opensearch versions < 2.10');
+            }
+        }
+
+        if (!isInteger(dimension)) {
+            throw new Error(`${this.field} must have an dimension property set to an integer`);
```
lesleydreyer, line 31: "a dimension"  instead of  "an dimension"

PR #4133, opened 2025-08-04
Float coercion fix

jsnoble:
- fix issue that type coerction would break if a value was set to `*`, for example this expression `count:{2 TO *]` if count type was a float. Fixed the issue so it works


lesleydreyer (approved):
LGTM

busma13 (approved):
Looks good Jared

PR #4139, opened 2025-08-12
ensure the opensearch-client tests are being run in ci

jsnoble:


busma13 (approved):
Austin had us drop ES6 tests from the other test suites. Do we want to still test it here?
I have no strong opinion.

LGTM otherwise.

godber:
Sorry, ES 6 should be excluded for the following reasons:

* it has no operational relevance
* its a waste of CI time
* we have an issue open to remove the code support for ES6.

PR #4144, opened 2025-08-21
[teraslice-test-harness] Move job-components peerDependency to a direct dependency

busma13:
This PR makes job-components a direct dependency of teraslice-test-harness instead of a peerDependency. It is used directly by the code. Peer dependency updates don't cause version bumps when using the ts-scripts bump script, causing the test harness to become incompatible with other libraries because new NPM releases weren't being made.


jsnoble (approved):
LGTM

PR #4156, opened 2025-09-17
[Helm] Remove legacy metrics exporter and stop scraping /cluster/stats

### LINKED ISSUE #3910: Prometheus Endpoint /cluster/stats problems
Currently the Teraslice helm chart creates a prometheus Service Monitor that scrapes two endpoints for metrics.  The metrics from `/cluster/stats` are redundant now (`teraslice_slices_processed` and others are provided by the new metrics provider), so this extra endpoint should be removed:

https://github.com/terascope/teraslice/blob/f9b4319c3c0f2d76a4bcaa1a4d7afa8947f532dc/helm/teraslice/templates/servicemonitor.yaml#L10-L16

Also the logic for that Service Monitor template could be cleaned up, it doesn't make sense to allow it to be created with an empty `endpoints` array, for instance, so maybe the whole file should be wrapped with:

`{{- if and .Values.serviceMonitor.enabled .Values.terafoundation.prom_metrics_enabled }}`

The other problem with the `/cluster/stats` endpoint is that normally it provides json, but it will provide Prometheus-formatted metrics with the correct request headers.  However, it delivers an invalid content-type for Prometheus 3.0 : `Content-Type: text/html`: https://prometheus.io/docs/prometheus/3.0/migration/#scrape-protocols

Adding `spec.fallbackScrapeProtocol: PrometheusText1.0.0` to this Service Monitor object would avoid scrape errors, if for some reason we still want to scrape it.  Since we don't need `/cluster/stats` prometheus-style metrics anymore, maybe that functionality can simply be removed.

Here's what those `/cluster/stats` metrics look like:
```
curl -Ss --header "Accept: application/openmetrics-text;" teraslice/cluster/stats -v

> GET /cluster/stats HTTP/1.1
> Accept: application/openmetrics-text;
< HTTP/1.1 200 OK
< Content-Type: text/html; charset=utf-8

# TYPE teraslice_slices_processed counter
teraslice_slices_processed{cluster="teraslice"} 1004456
# TYPE teraslice_slices_failed counter
teraslice_slices_failed{cluster="teraslice"} 24
# TYPE teraslice_slices_queued counter
teraslice_slices_queued{cluster="teraslice"} 16
# TYPE teraslice_workers_joined counter
teraslice_workers_joined{cluster="teraslice"} 12
# TYPE teraslice_workers_disconnected counter
teraslice_workers_disconnected{cluster="teraslice"} 0
# TYPE teraslice_workers_reconnected counter
teraslice_workers_reconnected{cluster="teraslice"} 0

```

briend:
- Remove references and templating for the EOL external metrics exporter: https://github.com/terascope/teraslice-exporter
- Stop scraping /cluster/stats  (redundant and has other compatibility issues)
- Don't hard-code service account name; avoid collisions in shared namespaces (reverts [this change](https://github.com/terascope/teraslice/commit/c2bfb2230ef47352cf8cca41a21423faa1fb2d6f#diff-1be2e204031160fc47f1612c395d4ca824f12ccdd4590495e2e7b6b38fbed8aaR93))

fixes #3910

### DIFF
--- a/helm/teraslice/README.md
+++ b/helm/teraslice/README.md
@@ -77,12 +77,6 @@ View the `values.yaml` for charts configuration settings.
 | `persistence.accessModes`         | Storage access modes                            | `["ReadWriteMany"]`       |
 | `extraVolumes`                    | Additional volumes                              | `[]`                      |
 | `extraVolumeMounts`               | Additional volume mounts                        | `[]`                      |
-| `exporter.enabled`                | Enable external exporter                        | `false`                   |
-| `exporter.image.repository`       | Exporter image repository                       | `terascope/teraslice-exporter` |
-| `exporter.image.tag`              | Exporter image tag. This image is archived as of right now. See repo [here](https://github.com/terascope/teraslice-exporter)                | `v0.4.0`                  |
-| `exporter.image.pullPolicy`       | Exporter image pull policy                      | `IfNotPresent`            |
-| `exporter.env.TERASLICE_URL`      | URL for the Teraslice service                   | `http://localhost:5678`   |
-| `exporter.env.PORT`               | Exporter port                                   | `8080`                    |
 | `serviceMonitor.enabled`          | Enable Prometheus service monitor               | `false`                   |
 | `serviceMonitor.interval`         | Scrape interval                                | `60s`                     |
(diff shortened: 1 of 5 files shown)


briend:
closing in favor of https://github.com/terascope/teraslice/pull/4157

PR #4157, opened 2025-09-17
[Helm] Remove legacy metrics exporter and stop scraping /cluster/stats

### LINKED ISSUE #3910: Prometheus Endpoint /cluster/stats problems

briend:
- Remove references and templating for the EOL external metrics exporter: https://github.com/terascope/teraslice-exporter
- Stop scraping /cluster/stats  (redundant and has other compatibility issues)
- Don't hard-code service account name; avoid collisions in shared namespaces (reverts [this change](https://github.com/terascope/teraslice/commit/c2bfb2230ef47352cf8cca41a21423faa1fb2d6f#diff-1be2e204031160fc47f1612c395d4ca824f12ccdd4590495e2e7b6b38fbed8aaR93))

fixes #3910

### DIFF
--- a/helm/teraslice/values.yaml
+++ b/helm/teraslice/values.yaml
@@ -50,8 +50,7 @@ worker:
 terafoundation:
   environment: production
   log_level: info
-  # internal metrics exporter.  Do not combine with the external metrics server
-  # exporter.enabled: false
+  # internal metrics exporter.
   prom_metrics_enabled: false
   prom_metrics_port: 3333
   prom_metrics_add_default: true
@@ -90,7 +89,7 @@ serviceAccount:
   annotations: {}
   # The name of the service account to use.
   # If not set and create is true, a name is generated using the fullname template
-  name: master-service-account
+  name: ""
 
 podSecurityContext: {} # fsGroup: 2000
 
@@ -172,18 +171,6 @@ extraVolumeMounts: []
   # - name: extras
   #   mountPath: /usr/share/extras
   #   readOnly: true
-
-# external exporter.  Do not combine with the internal metrics server
-# terafoundation.prom_metrics_enabled: false
-exporter:
-  enabled: false
-  image:
-    repository: terascope/teraslice-exporter
-    tag: v0.4.0
-    pullPolicy: IfNotPresent
-  env:
-    TERASLICE_URL: http://localhost:5678
-    PORT: 8080
 
 serviceMonitor:
   enabled: false
--- a/helm/teraslice/README.md
+++ b/helm/teraslice/README.md
@@ -77,12 +77,6 @@ View the `values.yaml` for charts configuration settings.
 | `persistence.accessModes`         | Storage access modes                            | `["ReadWriteMany"]`       |
 | `extraVolumes`                    | Additional volumes                              | `[]`                      |
 | `extraVolumeMounts`               | Additional volume mounts                        | `[]`                      |
-| `exporter.enabled`                | Enable external exporter                        | `false`                   |
-| `exporter.image.repository`       | Exporter image repository                       | `terascope/teraslice-exporter` |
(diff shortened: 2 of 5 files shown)


briend:
I  must have confused the CI a bit more here: 
https://github.com/terascope/teraslice/actions/runs/17808684763/job/50630942214?pr=4157

```
Determining the checkout info
  /usr/bin/git branch --list --remote origin/teraslice-chart-1
  /usr/bin/git tag --list teraslice-chart-1
  Error: A branch or tag with the name 'teraslice-chart-1' could not be found
```

`teraslice-chart-1` was the name of branch when I made the [original commit](https://github.com/terascope/teraslice/pull/4156), but I change branch names to something more sensible

```
git remote add upstream git@github.com:terascope/teraslice.git
git checkout -b teraslice-chart-exporter
git push --set-upstream upstream teraslice-chart-exporter
```

So I'm not sure how it still thought it should be `teraslice-chart-1`

briend:
I think I failed to follow the instructions here exactly: https://github.com/terascope/teraslice/pull/4156#issuecomment-3304520432

helm/teraslice/values.yaml
```diff
@@ -90,7 +89,7 @@ serviceAccount:
   annotations: {}
   # The name of the service account to use.
   # If not set and create is true, a name is generated using the fullname template
-  name: master-service-account
+  name: ""
```
busma13, line 92: Yeah, I don't understand why I made this change. Sorry about that

briend:
I'm making a new PR for this with a fresh clone from terascope repo

busma13:
> Does it hurt if I did bump it in my new PR? #4158

Joseph looked through the workflow and it should publish a new chart properly. My only worry was that the chart wouldn't get published without a teraslice release.

Issue #4159, opened 2025-09-17
remove prometheus format for 'cluster/stats' endpoint


busma13:
The `cluster/stats` endpoint accepts the `application/openmetrics-text` header and will return the stats in a format that is compatible with prometheus. PR #4158 removes the external prom metrics exporter capabilities from the teraslice helm file, which is the only known use case for this functionality. 

We should confirm there is no reason to keep this code and then remove it.

PR #4160, opened 2025-09-17
remove prometheus format from cluster/stats

busma13:
- Removes the ability of the `cluster/stats` endpoint to return stats in a prom metrics compatible format
- Updates docs

Ref: #4159

TODO: ensure this functionality is not used anywhere before merging

### DIFF
--- a/packages/teraslice/src/lib/cluster/services/api.ts
+++ b/packages/teraslice/src/lib/cluster/services/api.ts
@@ -14,9 +14,8 @@ import { makeLogger } from '../../workers/helpers/terafoundation.js';
 import { ExecutionService, JobsService, ClusterServiceType } from '../services/index.js';
 import type { JobsStorage, ExecutionStorage, StateStorage } from '../../storage/index.js';
 import {
-    makePrometheus, isPrometheusTerasliceRequest, makeTable,
-    sendError, handleTerasliceRequest, getSearchOptions,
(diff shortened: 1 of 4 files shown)


packages/teraslice/src/lib/cluster/services/api.ts
```diff
@@ -487,16 +486,11 @@ export class ApiService {
         });
 
         v1routes.get('/cluster/stats', (req, res) => {
-            const { name: cluster } = this.context.sysconfig.teraslice;
-
             const requestHandler = handleTerasliceRequest(req as TerasliceRequest, res, 'Could not get cluster statistics');
             requestHandler(async () => {
-                const stats = await executionService.getClusterAnalytics();
+                const stats = executionService.getClusterAnalytics();
```
godber, line 491: Why is the await being removed here?

godber:
>I think this is safe to merge.

Actually see review feed back

godber (commented):
See questions

packages/teraslice/src/lib/cluster/services/api.ts
busma13: `executionService.getClusterAnalytics()` is not an async function. It is just returning an object. I'm not sure why the await was there originally. It's hard to follow the git history to before the library was converted to typescript

godber (approved):

PR #4173, opened 2025-10-01
build docker images once per workflow

busma13:
- add build-dev-docker-images job to `test.yml` workflow. This builds a docker image per node version and saves it as an artifact using the workflow ID
- add steps to e2e test jobs that download docker image artifact and load into docker
- update `yarn test` commands to `SKIP_DOCKER_BUILD_IN_E2E` and set `DEV_DOCKER_IMAGE` to our workflow ID

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -164,13 +164,13 @@ jobs:
         uses: docker/build-push-action@v6
         with:
           tags: ${{ github.run_id }}-nodev${{ matrix.node-version }}:latest
-          outputs: type=docker,dest=${{ runner.temp }}/${{ github.run_id }}.tar
+          outputs: type=docker,dest=${{ runner.temp }}/${{ github.run_id }}-nodev${{ matrix.node-version }}.tar
 
       - name: Upload artifact
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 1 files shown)


godber:
> I think ensuring the docker image has all the same packages is enough of a win to go with these changes.

Agreed ... does this have any impact on the release workflow?  Like, we test a specific artifact, we know it's good ... do we then release that?  Or do we make another thing, and release that, without any DIRECT testing?

godber:
Whatever `e2e-k8s-v2-encrypted-tests (22)` is doing it's probably over doing it.  What is the specific intent or goal of that test suite?  What tests are run?  What tests are truly necessary to validate that addition of ... whatever encrypted thing is added?

busma13:
> does this have any impact on the release workflow?

No. The image is built again for the publish-master.yml and another time for the publish-tag.yml. 

I think saving to ghcr would be a big win here.

godber:
>I think saving to ghcr would be a big win here.

We could ALSO save this artifact kind of "at the end" to ghcr, and somehow pass the tag on down to the release.

godber:
Approved, please merge and rebase to v3!

PR #4178, opened 2025-10-07
add base metadata to root context logger

sotojn:

- Adds base metadata to all teraslice services such as `cluster_master`, `worker`, and `execution_controller`. 
  - This mostly effects assets that use the context root logger to log things, which before was missing important metadata such as `ex_id` and `worker_id`
- Documents proper way to add child loggers when the context is available


sotojn:
I feel like this path is best because of several reasons:

1. We don't have to around modifying all assets that already use the context logger for logging. They automatically get all the correct metadata now.
2. It's a `teraslice` ONLY change. Meaning no other `terafoundation` applications get affected.
3. It only requires two line changes across two files using existing functions.
4. Adds metadata to ALL logging that occurs within terafoundation because terafoundation always used the context logger.

Issue #4176, opened 2025-10-03
setup release pipeline for release/teraslice-v* branches


busma13:
We want to create artifacts as we progress through the teraslice 3.0.0 milestone

The following should be created whenever a teraslice version is bumped:
- git tag in this format: `v3.0.0-dev.N`
- github release - marked prerelease and containing `v3.0.0-dev.N` tag
- npm package tagged as `prerelease`
- docker images tagged `v3.0.0-dev.N-nodev22.x.x` and `v3.0.0-dev.N-nodev24.x.x`

ref: #4170

busma13:
My plan is to create a third publish related workflow, `publish-dev.yml`, that uses the `prerelease` type flag for `ts-scripts publish`. I also need to update `publish-tag.yml` to not run if the teraslice release that triggered the workflow has a pre-release semver version.

godber:
This logic looks good.  The releases are triggered by the version change?

busma13:
> The releases are triggered by the version change?

Correct. I am copying the relevant code from `publish-master.yml` which runs on all merges. It runs `ts-scripts publish` for both npm (sub-packages if they've been bumped) and docker (dev-local-nodev22 image), then creates a release if the teraslice version has been bumped. I just need to adjust the `publish` flags to do prereleases.

busma13:
The other issue is that `semver-compare-cli`, which we use to determine if version has changed, does not support pre-releases. I am finding an alternative right now.

busma13:
This is the semver cli tool that I am considering: https://github.com/fsaintjacques/semver-tool
This is the only thing I've found that runs in the command line and can compare prerelease versions correctly.

godber:
https://github.com/fsaintjacques/semver-tool looks good, I'd just vendor the current version of that file.

Issue #3946, opened 2025-02-04
Automated teraslice releases fail if the new version is a release candidate


busma13:
When updating the teraslice version by a patch with a release candidate suffix `semver-compare` does not think the version has been update, or maybe it throws an error. Because of this no automated release is created.

CI pipeline example:
https://github.com/terascope/teraslice/actions/runs/13143910242/job/36677598151

PR #4181, opened 2025-10-07
[CI] V3 release pipeline

busma13:
- add `semver-tools.sh` script
- update release workflows
  - use `semver-tool.sh` in `publish-master.yml` instead of `semver-compare-cli`. This will fix merges to master with a new semver pre-releases tag not being released #3946  
  - `publish-dev.yml` is triggered on a merge to `release/teraslice-v*` and does the following:
     - `ts-scripts publish -t prerelease npm` will publish all packages that currently have an updated pre-release version.
     - `ts-scripts publish -t prerelease docker` will publish node 22/24 docker images only if teraslice has a pre-release version and an image with that tag does not already exist.
     - create a github release if the teraslice version has increased and it is a pre-release.
     - create a slack announcement that a new pre-release has been published.
   - `publish-tag.yml` will not publish to npm or docker if the teraslice version is a prerelease
 

ref: #4176


godber (approved):
My comment here is minor, can be skipped or addressed on next PR

PR #4198, opened 2025-10-14
[development] Add k8s command custom config 

sotojn:

- Adds `--generate-config` flag to `k8s-env` command
  - This generates a templated custom config to be used with the new `--config-file` argument which allows the user to pass in a custom config
- Adds `hostVolumePath` option that allows the user to mount directories to specific services.
- Adds `hostPort` option that allows the user to specify which port to use for a given service on the host.
- Cross validation within helmfile go templates to ensure `hostVolumePath` configurations don't overlap.
  - Ensures `stateCluster` config is a service that enabled within the teraslice helm go template


busma13 (approved):
I think this will work fine as is. We still have other features we want to add, but being able to use this will help us decide what to do first.

Issue #4200, opened 2025-10-15
[scripts] Add more improvements to k8s command


sotojn:
I've been adding some more functionality to `yarn k8s` that allow the user to generate, edit, and pass in a custom configuration. The core functionally works, but still needs some much needed features to make using it as easy as possible. While working on this I also found some bugs and issues that will need to be addressed and will also list them below.

- Add more cross validation checking within `yarn k8s` 
  - Cross validation already occurs within the templating files, but it would be nice to catch errors before we actually have to run the `helmfile diff` command. It will save a lot of time because config issues will pop up a lot sooner.
- Add auto generation feature for certificates and teraslice images
  - When using a custom config, the user must already have built the image specified in the yaml file. We need a way for the user to specify that they want it built automatically. Also using a custom config forces the user to create and provide their own certificates. We should also be able to leverage the autogeneration that the k8s command already has to generate these for the user if specified.
- Fix debug logging. As of right now there is no option to view debug logging. This renders all debug logs useless in the k8s command code.
- Add helmfile templating testing to ensure all the validation checks are functional. There are a lot of scenarios that are hard to keep track of. We should test them all. 

Ref to custom config PR: #4198

PR #4208, opened 2025-10-22
[scripts] Add more k8s env tools

sotojn:

- Adds chaos-mesh, Prometheus operator with grafana,  and kafka ui to yarn k8s config 
- Fixes issue with rebuilding teraslice with a custom config
- Adds teraslice auto-build feature to k8s config file
- Adds better documentation on the templated config

Ref to issue #4200

### DIFF
--- a/e2e/package.json
+++ b/e2e/package.json
@@ -46,7 +46,7 @@
     },
     "devDependencies": {
         "@terascope/opensearch-client": "~2.0.0-dev.5",
-        "@terascope/scripts": "~2.0.0-dev.6",
+        "@terascope/scripts": "~2.0.0-dev.7",
         "@terascope/types": "~2.0.0-dev.5",
         "@terascope/utils": "~2.0.0-dev.5",
         "bunyan": "~1.8.15",
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
         "@eslint/js": "~9.35.0",
         "@swc/core": "1.13.5",
         "@swc/jest": "~0.2.39",
-        "@terascope/scripts": "~2.0.0-dev.6",
+        "@terascope/scripts": "~2.0.0-dev.7",
         "@types/bluebird": "~3.5.42",
         "@types/convict": "~6.1.6",
         "@types/elasticsearch": "~5.0.43",
--- a/packages/scripts/package.json
+++ b/packages/scripts/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@terascope/scripts",
     "displayName": "Scripts",
-    "version": "2.0.0-dev.6",
+    "version": "2.0.0-dev.7",
     "description": "A collection of terascope monorepo scripts",
     "homepage": "https://github.com/terascope/teraslice/tree/master/packages/scripts#readme",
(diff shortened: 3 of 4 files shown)


e2e/helm/templates/prometheus-stack.yaml.gotmpl
```diff
@@ -0,0 +1,16 @@
+grafana:
+    enabled: {{ .Values | get "prometheus_stack.grafana.enabled" true }}
+    adminUser: {{ .Values | get "prometheus_stack.grafana.adminUser" "admin" }}
+    adminPassword: {{ .Values | get "prometheus_stack.grafana.adminUser" "prom-operator" }}
```
busma13: should this be `adminPassword`?

packages/scripts/src/helpers/k8s-env/index.ts
```diff
@@ -142,8 +146,18 @@ export async function launchK8sEnv(options: K8sEnvOptions) {
 
     signale.pending('Loading teraslice image into kind cluster');
     if (options.configFile) {
-        const imageName = await getTerasliceImageFromConfigYaml(options.configFile);
-        await kind.loadTerasliceImage(imageName);
+        if (!await getConfigValueFromCustomYaml(options.configFile, 'teraslice.image.build')) {
+            const repo = await getConfigValueFromCustomYaml(options.configFile, 'teraslice.image.repository');
+            const tag = await getConfigValueFromCustomYaml(options.configFile, 'teraslice.image.tag');
+            const imageName = `${repo}:${tag}`;
```
busma13: All of these values are retrieved earlier in `launchK8sEnv`. We should save them for reuse.

packages/scripts/src/helpers/scripts.ts
```diff
@@ -1493,7 +1500,30 @@ export async function createMinioSecret(k8sClient: K8s): Promise<void> {
...
-    return `${customConfig.teraslice.image.repository}:${customConfig.teraslice.image.tag}`;
+
+    const value = get(customConfig, valuePath, undefined);
+    if (!value) {
+        throw new Error(`Failed to find a config value ${valuePath} in the custom yaml at ${configFilePath}`);
+    }
+    return value;
+}
+
+export async function setConfigValuesFromCustomYaml(
```
busma13:
```suggestion
export async function setConfigValuesForCustomYaml(
```
Would this make more sense as a name?

busma13 (changes requested):

packages/scripts/src/helpers/k8s-env/index.ts
```diff
@@ -85,8 +91,10 @@ export async function launchK8sEnv(options: K8sEnvOptions) {
     }
 
     try {
-        // We never want to build in the case we use a custom config
-        if (!options.configFile) {
+        if (
+            !options.configFile
+            || (await getConfigValueFromCustomYaml(options.configFile, 'teraslice.image.build'))
```
busma13: I would have stored this as a variable too since we use it twice. I find it hard to read. Maybe on the next PR?

Issue #4202, opened 2025-10-17
[scripts] Add TLS options for kafka and opensearch docker services


sotojn:
We only have TLS support with helm/kubernetes related e2e tests. This means if we want to do encrypted integration tests with kafka and opensearch, we would not be able to do it. It's important to note that tests within assets use the docker services to stand up integration tests.

We may even want to create some encryption tests within the assets once this is implemented.

PR #4214, opened 2025-10-28
 [scripts] Add TLS Kafka to scripts services 

sotojn:

- Adds TLS functionality to the Kafka docker service deployed with scripts 
  - This made several changes that allow for certificates to be generated and used outside of teraslice testing

Ref to issue #4202

### DIFF
--- a/e2e/package.json
+++ b/e2e/package.json
@@ -47,7 +47,7 @@
     "devDependencies": {
         "@terascope/core-utils": "~2.0.0-dev.7",
         "@terascope/opensearch-client": "~2.0.0-dev.7",
-        "@terascope/scripts": "~2.0.0-dev.10",
+        "@terascope/scripts": "~2.0.0-dev.11",
         "@terascope/types": "~2.0.0-dev.6",
         "bunyan": "~1.8.15",
         "fs-extra": "~11.3.1",
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
         "@eslint/js": "~9.35.0",
         "@swc/core": "1.13.5",
         "@swc/jest": "~0.2.39",
-        "@terascope/scripts": "~2.0.0-dev.10",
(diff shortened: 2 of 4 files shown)


packages/scripts/src/helpers/config.ts
```diff
@@ -160,6 +163,14 @@ export const JEST_MAX_WORKERS = process.env.JEST_MAX_WORKERS
 
 export const NPM_DEFAULT_REGISTRY = 'https://registry.npmjs.org/';
 
+export const CERT_PATH = process.env.CERT_PATH
+    || ((ENCRYPT_KAFKA || ENCRYPT_MINIO || ENCRYPT_OPENSEARCH)
+        ? fs.mkdtempSync(path.join(os.tmpdir(), 'ts-CAs'))
+        : 'tmp/ts-certs'
+    );
+// We need to set CERT_PATH through process.env so config in e2e can use it.
+process.env.CERT_PATH = CERT_PATH;
```
busma13: We set env vars here: packages/scripts/src/helpers/test-runner/utils.ts getEnv()

packages/scripts/src/helpers/config.ts
```diff
@@ -160,6 +163,14 @@ export const JEST_MAX_WORKERS = process.env.JEST_MAX_WORKERS
 
 export const NPM_DEFAULT_REGISTRY = 'https://registry.npmjs.org/';
 
+export const CERT_PATH = process.env.CERT_PATH
+    || ((ENCRYPT_KAFKA || ENCRYPT_MINIO || ENCRYPT_OPENSEARCH)
```
busma13: Can we use the new ENCRYPTION_ENABLED var here instead?
```suggestion
    || (ENCRYPTION_ENABLED
```

busma13 (approved):
LGTM

Issue #4204, opened 2025-10-17
[terafoundation] Integrate Kafka connector into terafoundation


sotojn:
We are making an update to `terafoundation_kafka_connector` that changes from `node-rdkafka` client to the confluent one. 

https://github.com/terascope/kafka-assets/issues/871

This will remove the need to build `node-rdkafka` into the base image and allow us to make kafka into a built-in connector. We also need to ensure that this new connector client supports TLS encryption. We may want to think about how we can best expose all the client configuration within our config. 

This issue should be addressed first before we procceed #4202

Followed by this issue #4203

PR #4220, opened 2025-11-04
Integrate kafka connector

sotojn:

- Adds `terafoundation_kafka_connector` to terafoundation
- Replaces [base docker image](https://github.com/terascope/base-docker-image) with node-alpine

NOTE: We can potentially get binary mismatches because `kafka-assets` only builds assets with whatever os the github runner uses. This was always a problem but never seemed to be an issue 

Ref to issue #4204

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "teraslice-workspace",
     "displayName": "Teraslice",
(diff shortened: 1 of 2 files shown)


busma13 (approved):
LGTM

PR #4222, opened 2025-11-05
remove deprecated event handler

busma13:
This PR removes the event handler for `client:initialization:error`. Terafoundation no longer emits that error. it was removed in #1393


busma13:
replaced by #4227

Issue #4223, opened 2025-11-06
Fix asset pre-release tag handling in Teraslice


sotojn:
## Summary
We introduced a pre-release development pipeline for assets that publishes builds with a `dev` tag. This change causes several issues in Teraslice.

## Problems
1. **E2E Test Failures**  
   E2e tests parse asset zip filenames to extract metadata. The addition of the `dev` tag breaks this parsing logic.

2. **Incorrect Asset Caching**  
   The asset cache pipeline always pulls the latest assets from GitHub. If a pre-release (`dev`) asset is the latest, it ends up in the cache, causing E2E tests to unintentionally use development builds during CI runs on `master`. If the latest release is a dev release, we would need to cache said dev release and also cache the latest non-dev release.

3. **Job Validation Issues**  
   When resolving assets by name, a `dev` asset does not correctly resolve to its base name.

godber:
Ok, for these problems what are the solutions.  My thoughts (albeit somewhat uninformed on the details):

1. Make e2e asset parsing accept dev assets, this may have to be conditional on being on a release/dev branch.
2. Make asset caching accept dev assets, also may be conditional on being on a release/dev branch
3. I think we make sure that @jsnoble see's this problem and he can probably recommend the correct course of action.


For 1 and 2, we currently make assumptions about how things work for CI/PRs on the master branch, we need to make sensible assumptions for the newly created release/dev branches.

PR #4224, opened 2025-11-07
[teraslice, e2e] Add dev asset compatibility

sotojn:

- Allows teraslice to satisfy and validate an asset range that includes pre-release tags 
- Adds `USE_DEV_ASSETS` env var to e2e to determine if dev assets should be used in ci
- Fixes issues with version parsing with semver  
- Adds functionality to asset downloads to ensure both dev assets and non dev assets are downloaded into cache

Ref to issue #4223

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -965,6 +965,7 @@ jobs:
         elasticsearch-api-tests,
         elasticsearch-store-tests,
         teraslice-elasticsearch-tests,
+        xlucene-translator-tests,
         lint-and-sync,
         test-website,
         determine-tests-based-off-commit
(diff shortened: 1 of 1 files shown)


e2e/test/download-assets.ts
```diff
@@ -51,10 +51,46 @@ export const defaultAssetBundles = [
...
+    // Ensure the files are of the correct extension
+    if (!assetName.endsWith('.zip')) {
+        throw new Error(`Invalid file extension for asset file ${assetName}. Should be a .zip file.`);
+    } else {
+        const firstDashIndex = assetName.indexOf('-');
+        if (firstDashIndex === -1) {
+            throw new Error(`Error parsing asset file name. The name should have dashes in it (-)`);
+        }
+        // Removes the asset name prefix
+        // EX: v4.4.0-node-22-bundle.zip -> v4.4.0-node-22-bundle.zip -> v4.4.0
```
busma13: shouldn't this start out with the name prefix?

e2e/test/download-assets.ts
```diff
@@ -51,10 +51,46 @@ export const defaultAssetBundles = [
...
+            throw new Error(`Error parsing asset file name. The name should have "-node-**" in it to determine node version.`);
+        }
+        // Removes node*
+        // EX: v4.4.0-node-22-bundle.zip -> v4.4.0
+        assetVersion = assetVersion.slice(0, nodeIndex);
+
+        assetVersion = semver.clean(assetVersion);
+        if (assetVersion === null) {
+            throw new Error(`semver was unable to clean asset version for ${assetName}`);
+        } else {
```
busma13: you don't need an else here

busma13 (approved):
LGTM

PR #4225, opened 2025-11-08
Update templates on field additions

lesleydreyer:
- [x] Update templated index templates if new fields are detected so that templates are kept up to date and less prone to potential issues if an index's dynamic property is dynamic: true instead of the default: false


jsnoble (approved):
LGTM

PR #4249, opened 2025-12-15
Fix SchemaValidator null handling, add missing dependencies

busma13:
- add missing dependencies throughout packages
- Fix how SchemaValidator handles null values - do not replace with default


jsnoble (approved):
LGTM

PR #4257, opened 2026-01-08
Schema name fix

jsnoble:


packages/xlucene-translator/src/query-access/query-access.ts
```diff
@@ -237,8 +242,8 @@ export class QueryAccess<T extends Record<string, any> = Record<string, any>> {
         }
         const params = { ..._params };
 
-        const parser = this._restrict(query, _overrideParsedQuery);
-
+        // const parser = this._restrict(query, {}, _overrideParsedQuery);
```
busma13: still need this comment?

PR #4271, opened 2026-01-15
Move kafka connector dependency from terafoundation to teraslice

busma13:
- move the `terafoundation_kafka_connector` dependency from terafoundation to teraslice.
- bump terafoundation to v2.0.1
- bump teraslice to v3.0.1

 ref: #4272


jsnoble (approved):
LGTM

Issue #4244, opened 2025-12-11
[scripts] fix k8s dev mode


sotojn:
## Problem

When running the ts-scripts `k8s` command with `--dev`, it fails to run.

## Solutions

We need to fix the way we parse the main Dockerfile as we moved away from the base image and now do multistage build with `node:alpine`.

PR #4299, opened 2026-01-29
[scripts] Fix k8s dev mode 2

sotojn:

- Updates Dockerfile.dev file to use `node:alpine` as the base images is deprecated
- Adds logic that fixes mounting issues when parsing the original Dockerfile 
- Overall fixes `--dev`  functionality 

Ref to issue #4244


busma13 (approved):
LGTM

PR #4319, opened 2026-02-18
test opensearch ci problem 

sotojn:


sotojn:
We have implemented some logs from this test branch here 

#4341

PR #4343, opened 2026-03-03
fix isElasticsearchError

### LINKED ISSUE #4342: Possibly incorrect error at getFullErrorStack
Calling getFullErrorStack on a graphql error returns `` Elasticsearch Error: ``

Maybe it should return ```Opensearch Error```  if it is an opensearch error instead of elasticsearch, 
and if it doesn't return anything in _normalizeESError maybe it should drop the ``:`` at the end.

But seems like this error should not return as an Elasticsearch error because it happened calling graphql validation, not elastic/opensearch.
```
  GraphQLError: Cannot query field "a" on type "Query".
      at Object.Field ........etc...... {
    path: undefined,
    locations: [ [Object] ],
    extensions: [Object: null prototype] {}
  }
```

lesleydreyer:
We had a function isElasticsearchError that wasn't tesing against elastic/opensearch so we never realized it broke. 

It was considering an error to be from elasticsearch if it has a toJSON method, but other libraries like Graphql could have a toJSON method, and the current elastic/opensearch don't anymore, so the function isn't quite right. 

The toJSON method is from the legacy elasticsearch client https://github.com/elastic/elasticsearch-js-legacy/blob/16.x/src/lib/errors.js. 

The newer @elastic/elasticsearch and opensearch don't have the toJSON method...
i.e. v5 - https://github.com/elastic/elasticsearch-js/blob/5.x/lib/errors.js
i.e. v9 - https://github.com/elastic/elastic-transport-js/blob/main/src/errors.ts
i.e. OS - https://github.com/opensearch-project/opensearch-js/blob/main/lib/errors.js

--- 
### CHANGES 
it was discussed we should avoid adding more dependencies to core-utils so fixed it via checking error prototypes, and added opensearch tests to ensure it works in future

- [x] remove unused legacy elasticsearch deps
- [x] fix isElasticsearchError to work with newer elastic/opensearch clients and rename
- [x] add tests that run against opensearch to prevent future breakage

### EXAMPLE

It will turn an error like this...

image

to look like this...

image

Closes #4342

### DIFF
--- a/packages/core-utils/src/errors.ts
+++ b/packages/core-utils/src/errors.ts
@@ -1,4 +1,4 @@
-import type { AnyObject, Logger } from '@terascope/types';
+import type { Logger } from '@terascope/types';
 import STATUS_CODES from './status-codes.js';
 import { getFirst } from './arrays.js';
 import { isFunction } from './functions.js';


jsnoble (approved):
LGTM

Issue #4327, opened 2026-02-23
Remove Metrics worker class so we can remove all node-gyp dependencies


busma13:
Occassionally `gc-stats` will fail to build during a CI pipeline, which gets annoying. It is used to get metrics in the Worker `Metrics` class. This is one of 2 dependencies that still use node-gyp. The other is `event-loop-stats` which is also only used in `Metrics`.  Metrics can be turned on in a jobConfig via `performance_metrics`, but we now use `PromMetrics` in production to scrape similar data for dashboards. If the metrics are needed for local testing we can turn on PromMetrics instead.

PR #4345, opened 2026-03-04
[teraslice] Remove performance_metrics class and all node-gyp dependencies

sotojn:

- Remove `performance_metrics` class
  - This includes removing its implementation within the worker and ex classes
  - The `performance_metrics` class contained `event-loop-stats` and `gc-stats` which are the remaining node-gyp dependencies
- Removed `performance_metrics` setting from the job config
- Removed `performance_metrics` from `job` and `ex` mappings
- Removed `event-loop-stats` and `gc-stats` from teraslice deps
- Remove any `performance_metrics` docs 

Users can enable default metrics with the built-in prom-client, you get the same GC and event loop data.
                                                              
Add the following to your terafoundation config:                                                                                                                                                          

```yaml
terafoundation:
  prom_metrics_enabled: true
  prom_metrics_add_default: true         # <----- This will get you equivalent metrics 
  prom_metrics_port: 3333
```

Ref to issue #4327

### DIFF
--- a/helm/teraslice/Chart.yaml
+++ b/helm/teraslice/Chart.yaml
@@ -4,8 +4,8 @@ description: Teraslice -  Distributed computing platform for processing JSON dat
 home: https://github.com/terascope/teraslice
 icon: https://terascope.github.io/teraslice/img/logo.png
 type: application
-version: 3.9.0
-appVersion: v3.3.3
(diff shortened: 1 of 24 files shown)


busma13 (approved):
LGTM

PR #4355, opened 2026-03-11
Xlucene translator db tests

jsnoble:
- fix minor xlucene-parser geo bug
- add real db tests for rest of translator features to prove queries work, can remove redundant output of translator later when we work on query performance


packages/xlucene-translator/test/query-main-spec.ts
```diff
@@ -0,0 +1,1205 @@
...
+            const results = response.hits.hits.map((record) => record._source);
+
+            expect(results.length).toEqual(3);
+            expect(results).toMatchObject([
+                { bar: 'hello', bool: true, num: 50 },
+                { bar: 'hello', bool: false, num: 70 },
+                { bar: 'hello', bool: true, num: 40 },
+            ]);
+        });
+
```
lesleydreyer: maybe add another record that would match  h?llo  and  h.*o  other than hello

ciorg (commented):
LGTM

PR #4395, opened 2026-04-10
Core utils precision

ciorg:
There's an issue where the setPrecision function in core-utils does not properly truncate very small or very large numbers.  The reason is that setPrecision coverts the input to a string via `input.toString()` and then passes the value to setPrecisionFromString which uses the decimal index position to truncate the number string.   In cases where the number is   < 1×10^-6 or > 1×10^21`, toString()` converts the number to scientific notation which moves the decimal and throws off the truncation logic. 

An example is `0.00000012777405852461998.toString()` becomes `1.2777405852461998e-7`

The fix below should handle numbers between  ±`2^53 - 1`, which is just over 9 quadrillion.

### DIFF
--- a/packages/core-utils/package.json
+++ b/packages/core-utils/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@terascope/core-utils",
     "displayName": "Core Utils",
-    "version": "2.5.0",
+    "version": "2.5.1",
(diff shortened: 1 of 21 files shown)


jeffmontagna (approved):
Looks good to me.

PR #4398, opened 2026-04-14
[teraslice] Add curl back into teraslice image

sotojn:

- Adds `curl` to the teraslice image
  - We had this back in our old base image and need it 
- Bumps `teraslice` from `v3.5.2` to `v3.5.3`


busma13 (approved):
LGTM

Issue #4401, opened 2026-04-16
number functions in core-utils should handle inputs in scientific notation


ciorg:
When creating the number related functions in core-utils we didn't consider numbers in scientific notation.  Javascript automatically converts numbers to scientific notation if n >= 10^21 or n <= 10^-6.  

Since it is native to javascript most of our number functions can handle this format, but  it does cause problems for `isNumberLike`.  It returns a false negative if the input is a number string in scientific notation.   Even though the other function are probably fine we should add values in scientific notation to the tests just to be safe.

ciorg:
pr https://github.com/terascope/teraslice/pull/4403 fixed this bug

PR #4403, opened 2026-04-17
improved isNumberLike function and expanded number tests

ciorg:
* improved isNumberLike function be removing the regex
* expanded tests to include numbers in scientific notation
* addresses issue https://github.com/terascope/teraslice/issues/4401


jeffmontagna (approved):
Looks good.

PR #4406, opened 2026-04-22
[scripts] Export docker service logs to files in e2e

sotojn:
                                                                                                                                                                                                                   
- Service container logs are now automatically captured to `e2e/logs/<service>.log` files during e2e test runs 
  - This makies it possible to inspect service logs after a test completes or fails. 
  - Log files are flushed gracefully before cleanup
  - This feature can be turned off by adding `--logs=false` or by passing the env `FILE_LOGGING=false`
- This feature will also make logs dir for integration tests where the test command was executed
  - Adds uploading of integration test log artifacts to CI 
- Adds multi docker-compose support to Compose class in `docker-compose-js`
  - This enables extend and merge functionality
  - https://docs.docker.com/compose/how-tos/multiple-compose-files/
- Bumps **@terascope/scripts** from `v2.6.3` to `v2.7.0`
- Bumps **@terascope/docker-compose-js** from `v2.0.1` to `v2.1.0`


packages/scripts/src/helpers/test-runner/index.ts
```diff
@@ -301,12 +303,27 @@ async function runE2ETest(
...
+    // started via dockerRun, rather than relying on the clustering type to imply it.
     if (options.clusteringType === 'native') {
         try {
             tracker.addCleanup(
                 'e2e:services',
                 await ensureServices(suite, options)
             );
+            tracker.addCleanup(
+                'e2e:service-logs',
+                startServiceLogging(
```
busma13: if `--logs` is false shouldn't this be skipped?

packages/scripts/src/helpers/test-runner/index.ts
```diff
@@ -301,12 +303,27 @@ async function runE2ETest(
...
+    // charts into a kind cluster and dockerRun is not used.
+    // TODO: revisit this condition to more directly check whether services will be
+    // started via dockerRun, rather than relying on the clustering type to imply it.
     if (options.clusteringType === 'native') {
         try {
             tracker.addCleanup(
                 'e2e:services',
                 await ensureServices(suite, options)
             );
+            tracker.addCleanup(
```
busma13: If we called `startServiceLogging` from within `ensureServices` we would get service logs for integration tests as well.

busma13 (changes requested):

busma13 (approved):
LGTM. Just add some docs.

busma13 (approved):

PR #4409, opened 2026-04-24
[scripts] test runner - add playwright support

lesleydreyer:
- [x] Add support for Playwright to the test runner
- [x] Add support for running multiple frameworks per project (not recommend but if you're careful to not let the different config files interfere w/each w/options like testDir/testMatch/etc. it could work)

NO BREAKING CHANGES. 
Everything should continue to default to jest unless it detects this optional testFrameworks in the package.json like the example below, in which case it will run whichever frameworks are defined there.
```
"terascope": {
  "testSuite": "e2e",
  "testFrameworks": ["playwright"]
} 
```


jsnoble (approved):
LGTM, though this should not be a minor bump not a patch

Issue #4411, opened 2026-04-24
Add valkey client


busma13:
We need a wrapper class/client that will return a valkey client. We are going to try the [Glide](https://glide.valkey.io/) client first.

busma13:
Well, maybe we don't need a wrapper class. S3 doesn't have a wrapper. The opensearch-client needs a wrapper because we support different client versions and need to convert parameters. It also looks like we don't expose all of the features of the clients.

We can add a wrapper in the future. For now we can just return a client based on the config inputted.

Issue #4407, opened 2026-04-23
Add valkey connector


godber:
Add a connector for Valkey.  I think the [Valkey GLIDE](https://glide.valkey.io/overview/) client should be used, but maybe not since they don't support Alpine/MUSL.

I am not yet sure where this connector should live, do we add the API and connector right here in Teraslice?

More soon TBD

busma13:
The connectors for elasticsearch-next and s3 are included in terafoundation. I think Valkey could live there as well.

godber:
> The connectors for elasticsearch-next and s3 are included in terafoundation. I think Valkey could live there as well.

They both import the APIs from their respective assets.  My question was about BOTH the connector and API.  See

[s3 connector](https://github.com/terascope/teraslice/blob/7947d5ee897f6e7d18a7d5ac5a30e3e9aedf53fb/packages/terafoundation/src/connectors/s3.ts#L2)
[es next connector](https://github.com/terascope/teraslice/blob/7947d5ee897f6e7d18a7d5ac5a30e3e9aedf53fb/packages/terafoundation/src/connectors/elasticsearch-next.ts#L3)

sotojn:
So I found some conflicting docs on `node:alpine` support for valkey GLIDE. The link below states:

> Note: Currently Alpine Linux / MUSL is NOT supported.

https://glide.valkey.io/getting-started/quickstart/?lang=node#linux

But these other docs state:

> Alpine
All alpine versions that are using musl libc 1.2.3 (All Alpine non deprecated alpine versions) or higher should be supported. Tests are running on:
node:alpine (x86_64/amd64 and arm64/aarch64)

https://valkey.io/valkey-glide/node/#alpine

We use `node:alpine` as our base image and they seem to test on `node:lts-alpine`

https://github.com/valkey-io/valkey-glide/blob/7b643fb67d5547ddcf5d6bf4eddd6a8c3f29c692/.github/json_matrices/build-matrix.json#L51-L74

Maybe we can attempt to use it anyway and see what happens.

busma13:
`They both import the APIs from their respective assets.`

I didn't think this through obviously. Looking more closely:

- S3 imports `@terascope/file-asset-apis` to have access to the s3Client and configuration type definition
- ES-next imports the opensearch client from `@terascope/opensearch-client`, but doesn't import `elasticsearch-asset-apis`.

So the connector just needs the schema (and any validator function) and the ability to create a wrapper client for the service. That can live in terafoundation.

[...]

PR #4413, opened 2026-04-28
Valkey support

busma13:
- Adds the `@terascope/valkey` package. This is the terafoundation valkey connector. Uses `@valkey/valkey-glide` rust based client
- Adds Valkey support for docker, docker-compose and kubernetes.
- Adds`@terascope/valkey` as a dependency to teraslice
- Adds `@terascope/valkey` as a peerDependency to terafoundation

ref: #4407 , #4411


packages/scripts/src/helpers/test-runner/services.ts
```diff
@@ -835,6 +863,72 @@ async function checkUtility(options: TestOptions, startTime: number): Promise<vo
...
+            }
+
+            // remove when encryption supported
+            if (config.ENCRYPT_VALKEY) {
+                throw new Error('Valkey encryption not supported');
+            }
+
+            const response = await new Promise<string>((resolve) => {
+                const socket = net.createConnection(port, host, () => {
+                    socket.write('*1\r\n$4\r\nPING\r\n');
```
busma13, line 889: `PING` is a sufficient liveness probe, but it's not great for readiness when using cluster mode, replicas, or needing to load a snapshot. We don't have any of these use cases at the moment.

godber: inspecting the output of `info` is probably the next step after `ping` if it were necessary.

packages/valkey/test/valkey-spec.ts
```diff
@@ -0,0 +1,456 @@
...
+
+const baseConfig = {
+    addresses: [{ host: 'localhost', port: 6379 }]
+};
+
+function validate(extra: Record<string, unknown> = {}) {
+    return makeValidator().validate({ ...baseConfig, ...extra });
+}
+
+function expectThrows(extra: Record<string, unknown>, match: string | RegExp) {
```
godber, line 23: I think this might have been better named as `expectValidateToThrow`

godber: It probably isn't worth changing now, but I was confused why you'd want to throw so much.  But it's the behavior of the `validate()` call that is being tested.

packages/valkey/test/valkey-spec.ts
```diff
@@ -0,0 +1,456 @@
...
+        await expect(conn.client.get('key1')).resolves.toBe('value1');
+
+        await conn.client.set('key2', 'value2');
+        await expect(conn.client.exists(['key2'])).resolves.toBe(1);
+        await conn.client.del(['key2']);
+        await expect(conn.client.get('key2')).resolves.toBe(null);
+        await expect(conn.client.exists(['key2'])).resolves.toBe(0);
+
+        conn.client.close();
+    });
```
godber, line 455: Can you add the following new `it` demonstrations:

* `mset` and `mget` - https://valkey.io/commands/mset/, https://valkey.io/commands/mget/
* `incr` - https://valkey.io/commands/incr/
* `geosearch` - https://valkey.io/commands/geosearch/
  * should demonstrate setting and search both by point and polygon, may require discussion

godber (changes requested):
This looks great!  I would like you to demonstrate a few other select commands just so we have them handy.

godber (approved):
Lets just merge it and come back later to add some of these nice to haves.

PR #4415, opened 2026-04-30
[scripts] Test Runner - reduce jest config files

### LINKED ISSUE #4414: testing configs update
With the recent changes inside our testrunner to be able to use different frameworks (ie playwright vs jest). This brings up an issue surrounding  ease of use around this. Currently all packages in our monorepos have a jest.config.js file. We would need to add the playwright version or the future vite version of a config file for each one which will bloat the package.

The original reason why we had a jest.config.js file was that in a few rare cases we needed to do specific things to the config for that testing environment. However, almost 95% of the time the config just copies what the top level config provides.

@lesleydreyer  and I discussed ways to minimize the need to have a test config file for each package so we don't need to bloat the code with a test config for each type that would need a different framework

lesleydreyer:
Adds support to not require a jest.config.js in every package. 

The test runner will chunk the suite's packages into groups with config files and groups without.  For the groups without, it will look in the root for a ``jest.make-config.js`` (or playwright). This file should return a function that accept an array of directories and returns a json test config. 

If the framework supports --config as JSON it will pass the config as stringified JSON, otherwise it will write the file to tmp dir and pass file path to --config. Was going to make it check for existence before writing, but tmp folders get cleaned regularly I think, plus it seems like it would be confusing to developers if they made a change to the config and it didn't reflect in the tests right away... thought about adding a flag to the test runner CLI like 'rewrite-config' or something, but that also seems like it would get confusing to people.

NOTE: Increased CI batch size from 1 to 6, but it doesn't speed it up much.

NOTE: eslint didn't seem to need the custom jest config globals.window so i deleted it

Closes #4414


packages/teraslice-client-js/jest.config.js
```diff
@@ -1,8 +1,12 @@
 import { fileURLToPath } from 'node:url';
-import baseConfig from '../../jest.config.base.js';
+import baseConfig from '../../jest.make-config.js';
 
 const dirPath = fileURLToPath(new URL('.', import.meta.url));
 
-const config = baseConfig(dirPath);
+// NOTE: flaky, fails a lot running in batch w/other packages in CI,
+// adding this config here is a hack to make it run w/less packages -
```
jsnoble: should write a ticket about this if we bring it in

jsnoble (approved):
LGTM

Issue #596, opened 2017-10-27
Track slice retries


kstaken:
It would be helpful to be able to see how many times a slice retries before it is considered a success or failure. Should probably be part of the state record for the slice.

jsnoble:
Is this still useful in case of kafka where we can't even trace to the source documents from the slice?

godber:
Without traceability to which specific records in the kafka case, I think the benefit of this is pretty marginal.

sotojn:
My original implementation tracked retries in an object that included op_name, error messages (truncated to 500 chars), and the amount of times the specific operation failed a slice. This was canned for now as it adds a lot of potential overhead to every retried slice record. Maybe if we were able to dynamically enable this level of detail  in the future we could add it. Heres an example slice below that failed all 3 times:

```json
{
        "_index": "ts-dev1__state-2026.04",
        "_id": "6a71fd8a-0fbc-4df9-9e0a-fc67beff7f2c",
        "_score": 1.8382792,
        "_source": {
          "slice_id": "6a71fd8a-0fbc-4df9-9e0a-fc67beff7f2c",
          "slicer_id": 0,
          "slicer_order": 5,
          "request": "{}",
          "state": "error",
          "ex_id": "8098e4a1-7474-4dd2-9fc5-ea901b959c9c",
          "_created": "2026-04-29T23:34:13.211Z",
          "_updated": "2026-04-29T23:34:23.422Z",
          "error": "TSError: Local: Queue full, caused by Error: Local: Queue full\n    at pRetry (file:///app/source/packages/core-utils/dist/src/promises.js:103:21)\n    at async SliceExecution.run ...."
          "retry_log": [
            {
              "op": "kafka_sender",
              "retry_count": 3,
              "errors": [
                "Local: Queue full",
                "Local: Queue full",
                "Local: Queue full"
              ]
            }
          ]
        }
      }
```

godber:
This shipped in Teraslice v3.8.0: https://github.com/terascope/teraslice/releases/tag/v3.8.0

PR #4416, opened 2026-04-30
[teraslice] add retry_count to slice record

sotojn:

- Adds `retry_count` to slice records under the condition a slice fails at least once
  - This is so we don't add un-meaningful data to all slice records 
- Adds tests that ensure slice records include `retry_count` at the appropriate times

Ref to issue #596


busma13 (commented):
LGTM. I am not going to approve until we get the valkey npm repo set up, so we don't merge by accident before that's ready.

busma13 (approved):

PR #4420, opened 2026-05-05
imporove logging in elasticsearch-api

sotojn:
TBD


godber:
With regards to logging the query, maybe we can add that as a conditional and have a `log_queries: false` that Teraslice sets to true when creating the object.

sotojn:
This PR is stale and we've decided against this because we haven't determined what we do and don't want to show the operator in the logs.

Issue #4429, opened 2026-05-12
[terafoundation] Add logger registry to makeLogger function


sotojn:
## Issue 

It turns out that bunyan has no mechanism to keep track of any child loggers. There actually isn't any relationship with a parent or child logger at all. Essentially when you call `logger.child()` it just creates a copy of the logger which would inherit fields (for example `worker_id`, `ex_id`) but nothing else. We are trying to make log levels dynamic through the teraslice api but have no effective way of changing all the child loggers to the same level. #4422 #4426 #4397

## Proposed solution

Mostly all loggers that get created throughout the code base use the foundation apis `makeLogger` function. The only exception I could find is within the schema validator which doesn't have the context when validating the terafoundation config. We can add a weakMap to the context that we will use to register each child logger that gets created with `makeLogger` and a new function on the foundation api that recursively changes the level on all child loggers.

PR #4430, opened 2026-05-13
[terafoundation] Add logger registry and setLogLevel to foundation APIs

sotojn:

- Adds a new foundation api `setLogLevel` [b1bee05]
  - This is paired with a registry that keeps track of all child loggers that are made with the `makeLogger` api
  - Now developers can change all log levels across all child loggers to the same log level
- Fixes issue where we create a new logger for every new slice that gets initialized [32343b4]
  - Now we use one logger that dynamically changes the `slice_id` field on init
- Updates development logging documentation [628119f] [ae14229]
  - We now define a set of rules when making loggers and state best practices for wording logs
- Bumps **terafoundation** from `v2.10.0` to `v2.11.0` [60984a0]

Ref to issue #4429

### DIFF
--- a/packages/terafoundation/src/api/index.ts
+++ b/packages/terafoundation/src/api/index.ts
@@ -23,7 +23,7 @@ export default function registerApis(
     const childLoggers = new Set<WeakRef<ChildLogger>>();
     const childLoggerRegistry = new FinalizationRegistry((ref: WeakRef<ChildLogger>) => {
         childLoggers.delete(ref);
-        context.logger.debug('child logger GC\'d and pruned from registry');
+        context.logger.debug('child logger collected and removed from logger registry');
     });
 
     // connection cache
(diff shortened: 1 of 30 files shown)


sotojn:
@jsnoble These are the changes you may be concerned about https://github.com/terascope/teraslice/commit/32343b47a5421ea4710d3ef4d406c9197b3af8be

packages/terafoundation/src/api/index.ts
```diff
@@ -13,6 +13,16 @@ export default function registerApis(context: Terafoundation.Context): void {
...
     context.logger = createRootLogger(context);
 
+    // Tracks all child loggers so setLogLevel can update them all.
+    // WeakRefs allow GC'd loggers to be collected, and FinalizationRegistry
+    // removes their dead references from the Set to prevent unbounded growth.
+    type ChildLogger = ReturnType<typeof context.logger.child>;
+    const childLoggers = new Set<WeakRef<ChildLogger>>();
+    const childLoggerRegistry = new FinalizationRegistry((ref: WeakRef<ChildLogger>) => {
+        childLoggers.delete(ref);
+        context.logger.debug('child logger collected and removed from logger registry');
```
busma13:
```suggestion
        context.logger.debug('child logger garbage collected and removed from logger registry');
```

busma13 (changes requested):
One suggestion on a better log message

busma13:
And it looks like there are other packages that need to be bumped

busma13 (approved):
LGTM

jsnoble (approved):
LGTM

Issue #4431, opened 2026-05-13
Improve valkey docs


busma13:
Add example to valkey docs about how to make a connector using the teraslice context. This is how we would make a client within an asset. Since we don't have valkey asset there is no valkey-asset-apis for people to look at as an example.

We should probably do this for the other clients too.

Issue #4418, opened 2026-05-01
[Valkey] Add more documentation and tests


busma13:
To help people using the valkey connector, it would be nice to have documentation and tests that show some of the commands possible with the client.

* `mset` and `mget` - https://valkey.io/commands/mset/, https://valkey.io/commands/mget/
* `incr` - https://valkey.io/commands/incr/
* `geosearch` - https://valkey.io/commands/geosearch/
  * should demonstrate setting and search both by point and polygon, may require discussion

PR #4437, opened 2026-05-15
[valkey] improve docs

busma13:
This PR makes the docs clearer on how to use the Valkey package from terafoundation. Since it is a simple wrapper there aren't any other use cases currently.

Ref: #4418, #4431


godber (approved):
approved pending proper rendering check

PR #4438, opened 2026-05-15
[terafoundation] Add core fields to root logger

sotojn:

## Terafoundation

- Adds new param `loggerMetadataFields` to context creation
  - This allows terafoundation apps to add logger fields to the root logger on context creation

## Teraslice

- Rework EX string decoder to get information before context creation
  - This fixes an issue where the root logger is missing filterable fields like `ex_id`, `job_id`, and `worker_id` on logs
  - This also fixes an issue where we use have to use helper functions `makeContextLogger` or `makeExContextLogger` to create child loggers because the root logger was missing fields

This relates to  #4429 #4430


jsnoble (approved):
LGTM

PR #4442, opened 2026-05-19
[scripts] Add auto topic creation configuration to scripts

sotojn:

- Adds new env variable `KAFKA_AUTO_CREATE_TOPICS_ENABLE` to scripts
  - This allows the ability to set `auto.create.topics.enable` on the broker
  - This setting is `true` by default
  - Bumps **@terascope/scripts** from `v2.11.1` to `v2.12.0`


packages/scripts/src/helpers/test-runner/services.ts
```diff
@@ -100,6 +100,7 @@ const services: Readonly<Record<Service, Readonly<DockerRunOptions>>> = {
             KAFKA_TRANSACTION_STATE_LOG_REPLICATION_FACTOR: `${config.KAFKA_TRANSACTION_STATE_LOG_REPLICATION_FACTOR}`,
             KAFKA_TRANSACTION_STATE_LOG_MIN_ISR: config.KAFKA_TRANSACTION_STATE_LOG_MIN_ISR,
             KAFKA_GROUP_INITIAL_REBALANCE_DELAY_MS: config.KAFKA_GROUP_INITIAL_REBALANCE_DELAY_MS,
+            KAFKA_AUTO_CREATE_TOPICS_ENABLE: config.KAFKA_AUTO_CREATE_TOPICS_ENABLE,
```
jsnoble, line 103: is this the equivalence of 
```  
rdkafka_options: {
        allow.auto.create.topics': false,
},
```

for the client creation?

jsnoble (changes requested):

packages/scripts/src/helpers/test-runner/services.ts
busma13: Yes, same idea but enforced by the broker instead of the client.

jsnoble (approved):
LGTM

Issue #4441, opened 2026-05-18
Add eslint plugin to catch missing dependencies


busma13:
Because of our monorepo structure, it is easy to import a library that is already in the node_modules and forget to add it to the package.json. Everything works locally, but the npm package will not work when imported into other projects. There are a few eslint plugins that do this.

busma13:
It looks like we already install [eslint-plugin-import](https://github.com/import-js/eslint-plugin-import), but the plugin is not being imported and added to the config, even though we have `import` rules in several areas. There was a note added when we removed the import about compatibility with eslint v9. I've tested locally and everything works properly with v9.

 We have the rule we need, `import/no-extraneous-dependencies`, added to js files, but it has been set to off since first added. It is not included in the typescript or jest sections.

PR #4443, opened 2026-05-19
[eslint-config] enable import checks

busma13:
- Installs `eslint-plugin-import` in the `eslint-config` package
- Set the `import/no-extraneous-dependencies` rule to error for .ts and .js files within src and test directories. We exclude devDependencies because it would require moving any typescript types imports from `devDependencies` to `dependencies`.

Ref: #4441


lesleydreyer (approved):

jsnoble (approved):
LGTM, just test are failing

Issue #4445, opened 2026-05-20
[scripts] Make kafka.yaml.gotmpl configurationOverrides dynamic


sotojn:
## Issue

The kafka helmfile template at https://github.com/terascope/teraslice/blob/master/packages/scripts/helm/templates/kafka.yaml.gotmpl has a hardcoded configuration inside the `configurationOverrides` key. The hostname `kafka-headless.services-dev1.svc.cluster.local` is environment-specific which for the most part is fine as we always use the `services-dev1` namespace but the SSL/non-SSL listener settings cannot be changed without editing the template directly.

The upstream chart we use (https://github.com/terascope/helm-charts/tree/ee08d055a1037356e5854f69893af23272f187c4/charts/kafka) which natively supports both a `configurationOverrides` map for Kafka broker config and an `envOverrides` for container environment variables, but we have no way to pass user values into either from the `values.yaml` file which is the interface we use to deploy the whole teraslice environemnt. `envOverrides` would be really useful because we can simply passthrough the kafka env vars we already support straight into it. We can use `configurationOverrides` for any other miscellaneous configs we don't support through scripts env vars. 

## Proposed Solution                                                               
                                                                                                                                                                                                                   
1. Add `kafka.configurationOverrides` to `values.yaml`
  - We can flat map the  kafka broker config key-value pairs that will get merged on top of the template's  SSL/non-SSL defaults. User-supplied keys will win. This maps directly to the upstream chart's https://github.com/terascope/helm-charts/blob/ee08d055a1037356e5854f69893af23272f187c4/charts/kafka/values.yaml                                                                                                                                     
                                                                                                                                                                                                                     
2. Add `kafka.envOverrides` to values.yaml
  - A flat key-value map passed directly to the upstream chart's https://github.com/terascope/helm-charts/blob/main/charts/kafka/values.yaml. 
  - Note: `envOverrides` will override `configurationOverrides `when both set the same Kafka property, since env vars are applied by the container entrypoint after server.properties is written. I will note that in a comment so users are aware of the hierarchy
                                                                                                                                                                                                                     
  Example `values.yaml` additions in scripts:
```yaml
 kafka:
    # Merged on top of SSL/non-SSL listener defaults.
    # Any valid Kafka broker config key. User values override template defaults.
    # ref: https://kafka.apache.org/documentation/#brokerconfigs
    configurationOverrides: {}
    # Example:
    #   auto.create.topics.enable: false
[...]

PR #4454, opened 2026-05-26
[scripts] Add configurationOverrides to k8s

sotojn:

- Adds `kafka.configurationOverrides` and `kafka.envOverrides` to `values.yaml`, allowing users to pass dynamic Kafka broker config and container env vars without editing templates directly
- Refactors `kafka.yaml.gotmpl` to build SSL/non-SSL listener defaults as a dict and merge user-supplied `configurationOverrides` on top using sprig `merge` — user keys always win
- Moves listener settings into `configurationOverrides` (not `envOverrides`) because the upstream chart's `brokerConfigs` helper merges `configurationOverrides` with its internal defaults safely, while `envOverrides` is purely additive and will conflict with any key the chart already generates internally
- `envOverrides` is only used for behavioral settings the chart does not generate internally (e.g. `KAFKA_AUTO_CREATE_TOPICS_ENABLE`)
- Adds `isK8s` branching in `config.ts` using `process.env.TEST_PLATFORM` to set k8s-appropriate listener defaults for both SSL and non-SSL modes
- Wires Kafka listener config from `config.ts` into `kafka.configurationOverrides` via a `kafkaK8sConfigOverrides` const in `scripts.ts` so the k8s path uses the correct listener settings. Behavioral vars are passed through `envOverrides`

Ref to issue #4445


packages/scripts/helm/templates/kafka.yaml.gotmpl
```diff
@@ -11,29 +11,64 @@ readinessProbe:
...
+##@ Build the default configurationOverrides as a dict so we can merge user overrides on top.
+##@ The upstream chart's brokerConfigs helper merges configurationOverrides with its internal
+##@ defaults, so keys set here safely override chart defaults without creating duplicate env vars.
+##@ Use configurationOverrides for listener/topology settings. Use envOverrides only for settings
+##@ the chart does not generate internally (e.g. KAFKA_AUTO_CREATE_TOPICS_ENABLE) — envOverrides
+##@ is purely additive and will conflict with any key the chart already generates.
+##@ `:=` declares and assigns a new template variable.
+##@ `dict` creates a key-value map: https://masterminds.github.io/sprig/dicts.html
+##@ `merge` mutates the destination dict by adding keys from the source only if they are missing
+##@ in the destination — so the destination always wins: https://masterminds.github.io/sprig/dicts.html#merge-mustmerge
```
busma13: Do we need comments explaining go?

busma13 (approved):
LGTM

packages/scripts/src/helpers/scripts.ts
```diff
@@ -1117,9 +1117,34 @@ function generateHelmValuesFromServices(
                 if (!caCert) {
                     caCert = readCertFromPath(path.join(config.CERT_PATH, 'CAs/rootCA.pem')).replace(/\n/g, '\\n');
                 }
-                values.setIn(['kafka', 'ssl', 'enabled'], true);
```
busma13, line 1120: I think we need this

Issue #4450, opened 2026-05-21
[teraslice] Add isRestarable label to execution controller pods


sotojn:
## Issue

We support restartable slicers for teraslice slicers since this PR #3740. Each slicer has a function `isRestartable()` that returns a boolean which would tell if said slicer is restarable or not. We need a way of getting this information on job submission so when the master deploys the execution Job resource, that will add a label to the ex pods that will indicate if said pod is restartable or not.

## Proposed Solution

TBD

godber:
We may already have a convention for what a "restartable"/"non-restartable" job label looks like.   Make sure to check the docs and code for existing conventions.

godber:
Shipped in https://github.com/terascope/teraslice/releases/tag/v3.11.0

PR #4462, opened 2026-05-29
[teraslice] Add relocatable label to ex pods

sotojn:

- Adds a new boolean label `job-property.teraslice.terascope.io/relocatable` to execution pods
  - This lets operators know if an execution living on a node can be relocated or not
- Bumps teraslice from `v3.10.0` to `v3.11.0`

Ref to issue #4450

### DIFF
--- a/docs/configuration/clustering-k8s.md
+++ b/docs/configuration/clustering-k8s.md
@@ -508,7 +508,7 @@ For resources related to specific jobs they will be labeled with the following:
 
 The following labels are applied only to execution controller resources:
 
-* `job-property.teraslice.terascope.io/relocatable` - `true` if the execution controller can be safely drained/deleted, `false` otherwise. This is determined by calling `isRelocatable()` on the job's slicer at execution start.
+* `job.teraslice.terascope.io/relocatable` - `true` if the execution controller can be safely drained/deleted, `false` otherwise. This is determined by calling `isRelocatable()` on the job's slicer at execution start.
 
 Note that the Teraslice job name is used in creating and labelling some of the
 Kubernetes resources.  Rather than enforce kubernetes strict DNS naming
--- a/docs/configuration/clustering-k8s.md
+++ b/docs/configuration/clustering-k8s.md
@@ -505,6 +505,7 @@ For resources related to specific jobs they will be labeled with the following:
 * `teraslice.terascope.io/exId` - the `exId`
 * `teraslice.terascope.io/jobId` - the `jobId`
 * `teraslice.terascope.io/jobName` - Teraslice job name, possibly modified
+* `job.teraslice.terascope.io/relocatable` - `true` if the node running the execution controller can be safely drained, `false` otherwise
 
 Note that the Teraslice job name is used in creating and labelling some of the
 Kubernetes resources.  Rather than enforce kubernetes strict DNS naming
--- a/docs/configuration/clustering-k8s.md
+++ b/docs/configuration/clustering-k8s.md
@@ -505,7 +505,10 @@ For resources related to specific jobs they will be labeled with the following:
 * `teraslice.terascope.io/exId` - the `exId`
 * `teraslice.terascope.io/jobId` - the `jobId`
 * `teraslice.terascope.io/jobName` - Teraslice job name, possibly modified
-* `job.teraslice.terascope.io/relocatable` - `true` if the node running the execution controller can be safely drained, `false` otherwise
+
+The following labels are applied only to execution controller resources:
+
+* `job-property.teraslice.terascope.io/relocatable` - `true` if the execution controller can be safely drained/deleted, `false` otherwise. This is determined by calling `isRelocatable()` on the job's slicer at execution start.
 
 Note that the Teraslice job name is used in creating and labelling some of the
 Kubernetes resources.  Rather than enforce kubernetes strict DNS naming
--- a/packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetesV2/k8sResource.ts
+++ b/packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetesV2/k8sResource.ts
@@ -238,6 +238,13 @@ export abstract class K8sResource<T extends TSService | TSDeployment | TSJob> {
         }
     }
 
+    _setRelocatableLabel(resource: TSJob | TSDeployment, relocatable: boolean) {
+        const key = `${this.jobPropertyLabelPrefix}/relocatable`;
+        const value = String(relocatable);
+        resource.metadata.labels[key] = value;
+        resource.spec.template.metadata.labels[key] = value;
+    }
+
     _setJobLabels(resource: TSJob | TSDeployment) {
         if (this.execution.labels != null) {
             Object.entries(this.execution.labels).forEach(([k, v]) => {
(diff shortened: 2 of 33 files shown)


sotojn:
Here is an example of a kafka to kafka job:

```diff
kubectl -n ts-dev1 describe pod ts-exc-j-kafka-test-kb-size-c3b0c3ad-45c5-7lqpb

Name:             ts-exc-j-kafka-test-kb-size-c3b0c3ad-45c5-7lqpb
Namespace:        ts-dev1
Priority:         0
Service Account:  default
Node:             k8s-env-control-plane/172.18.0.2
Start Time:       Thu, 28 May 2026 15:34:46 -0700
Labels:           app.kubernetes.io/component=execution_controller
                  app.kubernetes.io/instance=ts-dev1
                  app.kubernetes.io/name=teraslice
                  batch.kubernetes.io/controller-uid=cbb49c33-0011-49b6-8255-e5281f5d7b09
                  batch.kubernetes.io/job-name=ts-exc-j-kafka-test-kb-size-c3b0c3ad-45c5
                  controller-uid=cbb49c33-0011-49b6-8255-e5281f5d7b09
                  job-name=ts-exc-j-kafka-test-kb-size-c3b0c3ad-45c5
+                  job.teraslice.terascope.io/relocatable=true
                  teraslice.terascope.io/exId=ddb4b15c-9df5-4a7c-ae1a-14337ab54cc2
                  teraslice.terascope.io/jobId=c3b0c3ad-45c5-4538-a7fc-392d7ee17e3c
                  teraslice.terascope.io/jobName=j-kafka-test-kb-size
```

godber:
Previously we had the concept of a `stateful` job which would result in the following label being set on some of the resources, `job-property.teraslice.terascope.io/stateful`.  I forget the exact details, but we should consider whether we should use this label for this purpose.  On first blush I think we should, but want to think about it.  This is what I was getting at in this comment here:

https://github.com/terascope/teraslice/issues/4450#issuecomment-4513733246

sotojn:
Here are some new ex descriptions.

data gen to kafka job:

```diff
Name:             ts-exc-dgen-to-kafka-0f3e977c-cb44-jdfvs
Namespace:        ts-dev1
Priority:         0
Service Account:  default
Node:             k8s-env-control-plane/172.18.0.2
Start Time:       Fri, 29 May 2026 14:38:22 -0700
Labels:           app.kubernetes.io/component=execution_controller
                  app.kubernetes.io/instance=ts-dev1
                  app.kubernetes.io/name=teraslice
                  batch.kubernetes.io/controller-uid=019a4189-198b-4b7f-8f7a-3ba32dc6fd74
                  batch.kubernetes.io/job-name=ts-exc-dgen-to-kafka-0f3e977c-cb44
                  controller-uid=019a4189-198b-4b7f-8f7a-3ba32dc6fd74
                  job-name=ts-exc-dgen-to-kafka-0f3e977c-cb44
+                 job-property.teraslice.terascope.io/relocatable=false
                  teraslice.terascope.io/exId=4cd80444-aa3c-42f7-bff0-934a53e0037d
                  teraslice.terascope.io/jobId=0f3e977c-cb44-42fe-bacb-2d1ad4c49f4c
                  teraslice.terascope.io/jobName=dgen-to-kafka
```

kafka to kafka job:

```diff
Name:             ts-exc-j-kafka-test-kb-size-07f758a9-90cd-2m95g
Namespace:        ts-dev1
Priority:         0
Service Account:  default
Node:             k8s-env-control-plane/172.18.0.2
Start Time:       Fri, 29 May 2026 14:45:08 -0700
Labels:           app.kubernetes.io/component=execution_controller
                  app.kubernetes.io/instance=ts-dev1
                  app.kubernetes.io/name=teraslice
                  batch.kubernetes.io/controller-uid=0579d5e0-2706-4289-99b1-89b682b3c081
                  batch.kubernetes.io/job-name=ts-exc-j-kafka-test-kb-size-07f758a9-90cd
                  controller-uid=0579d5e0-2706-4289-99b1-89b682b3c081
                  job-name=ts-exc-j-kafka-test-kb-size-07f758a9-90cd
+                 job-property.teraslice.terascope.io/relocatable=true
                  teraslice.terascope.io/exId=d8622d63-6b3e-49cb-b5d0-9a34d54bf571
                  teraslice.terascope.io/jobId=07f758a9-90cd-4f3c-a89e-65c1df388025
                  teraslice.terascope.io/jobName=j-kafka-test-kb-size
```

sotojn:
I've also ensured these only show up on ex pods and not worker pods

packages/teraslice/src/lib/cluster/services/cluster/backends/kubernetesV2/k8sResource.ts
```diff
@@ -238,6 +238,13 @@ export abstract class K8sResource<T extends TSService | TSDeployment | TSJob> {
         }
     }
 
+    _setRelocatableLabel(resource: TSJob | TSDeployment, relocatable: boolean) {
+        const key = `${this.jobPropertyLabelPrefix}/relocatable`;
```
godber: This should be 
```suggestion
        const key = `${this.jobLabelPrefix}/relocatable`;
```

docs/configuration/clustering-k8s.md
```diff
@@ -506,6 +506,10 @@ For resources related to specific jobs they will be labeled with the following:
 * `teraslice.terascope.io/jobId` - the `jobId`
 * `teraslice.terascope.io/jobName` - Teraslice job name, possibly modified
 
+The following labels are applied only to execution controller resources:
+
+* `job-property.teraslice.terascope.io/relocatable` - `true` if the execution controller can be safely drained/deleted, `false` otherwise. This is determined by calling `isRelocatable()` on the job's slicer at execution start.
```
godber: job labels are labels set due to the intrinsic properties of the job, whereas job property labels are pass through labels from properties explicitly set by the job author ... I think that was my thinking when I made the distinction in the past ... 

so this should be

```suggestion
* `job.teraslice.terascope.io/relocatable` - `true` if the execution controller can be safely drained/deleted, `false` otherwise. This is determined by calling `isRelocatable()` on the job's slicer at execution start.
```

see also change below

godber (changes requested):
In addition to the two inline recommended changes here, during the review we realized that this functionality should be split up.  The `job.ts` should determine if the execution is relocatable and set that as a property on the EX, then the kubernetes backend code should look for that property and apply labels as appropriate.

packages/types/src/teraslice.ts
```diff
@@ -310,6 +310,8 @@ export interface ValidatedJobConfig {
     prom_metrics_port?: number;
     /** This will only be available in the context of k8s */
     prom_metrics_add_default?: boolean;
+    /** This will only be available in the context of k8s. Set by the system at execution start. */
+    relocatable?: boolean;
```
busma13, line 314: Types needs to be bumped

packages/teraslice/src/lib/cluster/services/jobs.ts
```diff
@@ -67,6 +67,39 @@ export class JobsService {
         return validJob;
     }
 
+    /**
+     * Loads the slicer for the given job to determine if it is relocatable,
+     * then sets that value as a property on the validJob so it is persisted
+     * on the EX record. The K8s backend reads this property to apply the
+     * appropriate pod label.
+     */
+    private async _applyRelocatable(
```
busma13: Does this also need to be called any other times, like when recovering a job?

busma13 (changes requested):

busma13 (approved):
LGTM

godber (approved):

PR #4479, opened 2026-06-16
Add support for Gateway API HTTPRoute

mkhpalm:
### [Helm] Summary

Add Gateway API HTTPRoute support to the Helm chart.

Gateway API Docs: https://gateway-api.sigs.k8s.io/reference/api-types/httproute/

### Changes

* Added `HTTPRoute` support for terasclice

  * `annotations`
  * `parentRefs`
  * `hostnames`
  * `rules`
  * `filters`
  * `timeouts`

### Motivation

This change provides an alternative to Ingress resources by supporting the Kubernetes Gateway API. Users can now expose Teraslice through an HTTPRoute while retaining flexibility for advanced routing features such as header matching, request filters, and timeout configuration.

### DIFF
--- a/helm/teraslice/templates/httproute.yaml
+++ b/helm/teraslice/templates/httproute.yaml
@@ -1,0 +1,42 @@
+{{- if .Values.httpRoute.enabled -}}
+{{- $fullName := include "teraslice.fullname" . -}}
+{{- $svcPort := .Values.service.port -}}
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
+metadata:
+  name: {{ $fullName }}
+  labels:
+    {{- include "teraslice.master.labels" . | nindent 4 }}
+  {{- with .Values.httpRoute.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  parentRefs:
+    {{- with .Values.httpRoute.parentRefs }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
+  {{- with .Values.httpRoute.hostnames }}
+  hostnames:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  rules:
+    {{- range .Values.httpRoute.rules }}
+    {{- with .matches }}
+    - matches:
+      {{- toYaml . | nindent 8 }}
+    {{- end }}
+    {{- with .filters }}
+      filters:
+      {{- toYaml . | nindent 8 }}
+    {{- end }}
+    {{- with .timeouts }}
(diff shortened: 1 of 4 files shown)


helm/teraslice/templates/httproute.yaml
```diff
@@ -0,0 +1,42 @@
...
+metadata:
+  name: {{ $fullName }}
+  labels:
+    {{- include "teraslice.master.labels" . | nindent 4 }}
+  {{- with .Values.httpRoute.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  parentRefs:
```
busma13, line 15: should this be within the `with`?

busma13: or do we want it to fail if no `parentRefs`?

mkhpalm: No, you can technically have no parentrefs / [] I guess. All of this format just comes from helm boilerplate directly. I'd just leave that as default.

https://github.com/helm/helm/blob/a82bd25fc0b11c66af163e6b28c42216f96b6855/pkg/chart/v2/util/create.go#L354

https://github.com/helm/helm/blob/a82bd25fc0b11c66af163e6b28c42216f96b6855/internal/chart/v3/util/create.go#L355

Example:
```
helm create test; cat test/templates/httproute.yaml
Creating test
{{- if .Values.httpRoute.enabled -}}
{{- $fullName := include "test.fullname" . -}}
{{- $svcPort := .Values.service.port -}}
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: {{ $fullName }}
  labels:
    {{- include "test.labels" . | nindent 4 }}
  {{- with .Values.httpRoute.annotations }}
  annotations:
    {{- toYaml . | nindent 4 }}
  {{- end }}
spec:
  parentRefs:
    {{- with .Values.httpRoute.parentRefs }}
      {{- toYaml . | nindent 4 }}
    {{- end }}
  {{- with .Values.httpRoute.hostnames }}
  hostnames:
    {{- toYaml . | nindent 4 }}
  {{- end }}
  rules:
    {{- range .Values.httpRoute.rules }}
    {{- with .matches }}
    - matches:
      {{- toYaml . | nindent 8 }}
    {{- end }}
    {{- with .filters }}
      filters:
      {{- toYaml . | nindent 8 }}
    {{- end }}
      backendRefs:
        - name: {{ $fullName }}
          port: {{ $svcPort }}
          weight: 1
    {{- end }}
{{- end }}
```

busma13 (approved):
LGTM

Issue #4456, opened 2026-05-26
Docker security improvements


busma13:
Following this [guide](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html) I made a list of potential improvements:

- [x] set a USER in dockerfile to prevent privilege escalation
  - [x] in k8s this is done with securityContext.runAsUser in yaml file
- [x] drop linux kernel capabilities with `--cap-drop All` then add in anything needed
  - [x] k8s - securityContext.capabilities.drop
- [x] run images with `--security-opt=no-new-privileges`
  - [x] k8s - securityContext.allowPrivilegeEscalation: false
- [x] Set filesystem & volumes to read-only
  - [x] k8s - set securityContext.readOnlyRootFilesystem: true

busma13:
In #4480 I completely missed that worker and execution controllers are created from templates, so they are not affected by the helm chart security changes. We need to add the same `securityContext` settings to either the templates or the k8s resource classes.

PR #4480, opened 2026-06-16
[teraslice, scripts] Dockerfile security improvements

busma13:
- create an unprivileged user in teraslice Dockerfile and run teraslice as that user.
- add more secure defaults to teraslice helmfile `securityContext`.
- add check to `k8s-env` command to fails fast if using a config-file and encrypted services (combination not supported).

Ref: #4456

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,10 +65,23 @@ RUN apk --no-cache add tini ca-certificates bash curl
 WORKDIR /app/source
 RUN corepack enable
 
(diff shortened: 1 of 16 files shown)


mkhpalm (approved):
LGTM!

godber:
We need to make sure we explicitly test the asset uploads after this user change and other things.

sotojn (approved):

PR #4482, opened 2026-06-17
[e2e, scripts] Fix waitForTeraslice() false success

busma13:
- add readinessProbe for teraslice in k8s e2e

Before this change a teraslice pod that was slow to shut down could respond to the API request that `waitForTeraslice()` used to check if teraslice was up and ready. Then the old pod would shut down and the new pod might not be ready when the first job was submitted. Now `helmfile sync` will block until the new teraslice pod is up and responding.


busma13:
This is getting bundled into #4484

Issue #3522, opened 2024-01-11
update ts-scripts to build multi-platform images


sotojn:
There is an issue with creating and publishing docker images in CI that only produce `amd/64` images in docker hub.

We need to look into the `dockerBuild()` function in [packages/scripts/src/helpers/scripts:line391](https://github.com/terascope/teraslice/blob/9a44c8eae4d1d41ceb96aab80478c197b9e0d09f/packages/scripts/src/helpers/scripts.ts#L391) and change [line:406](https://github.com/terascope/teraslice/blob/9a44c8eae4d1d41ceb96aab80478c197b9e0d09f/packages/scripts/src/helpers/scripts.ts#L406C1-L410C2) to use `docker buildx build` in the args section. 

We can then either hard code `--platforms linux/amd64, linux/arm64, darwin/amd64` inside the arguments or make it an option that gets passed in.

godber:
When making this change we need to consider all of the use cases this build script gets used, including internal use cases, and make sure to provide the appropriate flexibility to handle those scenarios.

sotojn:
I have implemented a solution in a PR #3527 that implements multiplatform builds of teraslice images in the case of a `merge to master` or a `release`. The issue is that the arm image builds very slow which may be tolerabole because CI only runs buildx in the cases stated above. It seems to be taking over 40 minutes to build the `yarn && yarn build` layer. I had to put a `--network-timeout` flag in yarn to not get a network timeout error as shown here in CI.

https://github.com/terascope/teraslice/actions/runs/7576194489/job/20634457971#step:6:1871

Here where I got the yarn network timeout solution:
https://github.com/yarnpkg/yarn/issues/8242

We could stick with having to deal with long builds or there are a couple other solutions.
- **look at ARM options for Github Runners**
  - It seems its in a closed beta right now, but it could be a potential solution later by building on two different runners with their native platforms and then merging a manifest.
  - https://github.blog/changelog/2023-10-30-accelerate-your-ci-cd-with-arm-based-hosted-runners-in-github-actions/
- **look at the cross compilation Docker build option**
  - Instead of only using emulation to build, we can introduce cross compilation inside the docker file itself
  - Basically emulation copies the whole system of another architecture in software, while cross-compilation uses native binaries to create new ones for a different architecture. Cross-compilation is mainly for compiling from what I've read (hence the name maybe). They can be used together, like using emulation for installing packages and cross-compilation for building code in a Dockerfile. Which results in faster build times hopefully.
[...]

PR #4483, opened 2026-06-18
[CI] Add multi-arch teraslice image support

sotojn:

- Splits the single `build-docker` job into separate `build-docker-amd64` and `build-docker-arm64` jobs across all three publish workflows [3ec9280] [9fc2b56] [fd4a311]
  - `amd64` builds run on `ubuntu-latest`
  - `arm64` builds run on `ubuntu-24.04-arm`
  - Each job builds and pushes by digest only (no tag), then uploads the digest and image tag as a workflow artifact
- Adds a `create-docker-manifests` job that runs after both arch builds complete
  - downloads the digests from both arch jobs and creates a multi-arch manifest using `docker buildx imagetools create`
  - Tags the manifest with the versioned image tag (e.g. `ghcr.io/terascope/teraslice:v1.2.3-nodev22.14.0`)
- Replaces `pnpm ts-scripts publish docker` with `docker/build-push-action` [3ec9280] [9fc2b56] [fd4a311]
  - Uses `docker/setup-buildx-action` for multi-platform build support
  - Passes `NODE_VERSION`, `TERASLICE_VERSION`, `BUILD_TIMESTAMP`, and `GITHUB_SHA` as build args
- Adds a `compute-vars` job to `publish-dev.yml` to stop docker builds on prerelease versions [6cc20c7]
  - Checks if the current version has a prerelease tag using `semver-tool.sh`
  - All docker jobs conditionally run only when `is_dev_release == 'true'`, preserving the behavior of the old `pnpm ts-scripts publish -t prerelease docker` command
- Extracts `NODE_VERSIONS` as a top-level `env` variable in `publish-dev.yml` and `publish-tag.yml` [6cc20c7]
  - Node versions are passed through job outputs via `fromJSON()` for matrix expansion
- Updates `append-docker-tags-to-release` in `publish-tag.yml` to read image tags from the amd64 artifact instead of the old `docker-tags.txt` format [fd4a311]

Ref to issue #3522

### DIFF
--- a/.github/workflows/publish-dev.yml
+++ b/.github/workflows/publish-dev.yml
@@ -54,7 +54,7 @@ jobs:
       - name: Publish to npm
         run: pnpm ts-scripts publish -t prerelease npm
 
-  build-docker:
+  build-docker-amd64:
     if: github.event.pull_request.merged == true
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-dev.yml
```diff
@@ -69,30 +69,152 @@ jobs:
...
-        uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # v6.0.8
+      - name: Login to GHCR
+        uses: docker/login-action@650006c6eb7dba73a995cc03b0b2d7f5ca915bee # v4.2.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@d7f5e7f509e45cec5c76c4d5afdd7de93d0b3df5 # v4.1.0
```
busma13, line 108: does `docker/build-push-action` use this builder automatically?

.github/workflows/publish-dev.yml
```diff
@@ -69,30 +69,152 @@ jobs:
...
+
+  build-docker-arm64:
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-24.04-arm
+    permissions:
+      contents: read
+      packages: write
+    strategy:
+      matrix:
+        node-version: [22, 24]
```
busma13: can we make this a top level variable?

.github/workflows/publish-dev.yml
```diff
@@ -69,30 +69,152 @@ jobs:
...
-      - name: Login to Docker Hub
+      - name: Login to GHCR
         uses: docker/login-action@650006c6eb7dba73a995cc03b0b2d7f5ca915bee # v4.2.0
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Publish to docker
-        run: pnpm ts-scripts publish -t prerelease -n ${{ matrix.node-version }} docker
```
busma13, line 95: This would check if the version was a prerelease before publishing. Does the action do this?

.github/workflows/publish-tag.yml
```diff
@@ -87,57 +86,171 @@ jobs:
...
+  docker-build-publish-arm64:
+    needs: ensure-non-pre-release
+    if: needs.ensure-non-pre-release.outputs.is_dev_release == 'false'
+    runs-on: ubuntu-24.04-arm
+    permissions:
+      contents: read
+      packages: write
+    strategy:
+      matrix:
+        node-version: [22, 24]
```
busma13: make top level var?

busma13 (approved):
LGTM

PR #4487, opened 2026-06-22
[CI] Fix teraslice image dev tag

sotojn:

- Fixes the teraslice dev tag to only have the major node version in the tag
- Moves node version in `publish-master.yml` to env var for easier node version updates in the future  

Ref to mistake PR #4483


busma13 (approved):
LGTM

PR #4488, opened 2026-06-25
start testing node 26

jsnoble:


jsnoble:
This will wait till kafka has binary builds for node 26

Issue #4490, opened 2026-06-29
Replace pMaps in ts-scripts publishToNPM with for loop


godber:
I think I confirmed on an internal project that `yarn npm publish` can have problems inside this `pMap` (even when reducing the `concurrency` to 1 and expanding yarn's `httpTimeout`)

https://github.com/terascope/teraslice/blob/master/packages/scripts/src/helpers/publish/index.ts#L32-L35

I did a quick patch on the internal project and it appeared that the serialized for loop resolved the problems I was having internally.

godber:
My patch was as follows

```
+    // Publish serially rather than via pMap. Concurrent `yarn npm publish`
+    // calls against a slow registry trigger yarn 4.x's got/p-cancelable
+    // "onCancel handler was attached after the promise settled" crash on the
+    // timeout-retry path, which can leave partial/duplicate publishes (409
+    // Conflict). A plain serial loop keeps one publish in flight at a time and
+    // gives clear per-package logging. We mirror pMap's stopOnError:false by
+    // collecting failures and throwing them together at the end.
+    const result = [];
+    const errors = [];
+    for (const pkgInfo of listPackages()) {
+        try {
+            result.push(await npmPublish(pkgInfo, options));
+        }
+        catch (err) {
+            signale.error(`Failed to publish ${pkgInfo.name}`, err);
+            errors.push(err);
+        }
+    }
+    if (errors.length) {
+        throw new AggregateError(errors, `Failed to publish ${errors.length} package(s)`);
+    }
+    signale.info('publishToNPM-result', result);

```

PR #4491, opened 2026-06-29
[scripts] Replace pMaps in ts-scripts publishToNPM with for loop

sotojn:

- Replaces the `pMap` function with a for loop when running `publishToNPM`

Ref to issue #4490


busma13 (approved):
LGTM

PR #4494, opened 2026-07-07
[scripts] Split scripts.ts into smaller focused files

sotojn:

- Splits `packages/scripts/src/helpers/scripts.ts`into smaller files
  - `exec.ts` — process execution helpers (`exec`, `fork`, `runTSScript`, `pgrep`)
  - `git.ts` — git helpers (`getCommitHash`, `gitDiff`, `getChangedFiles`)
  - `docker.ts` — all docker helpers
  - `package-manager.ts` — package manager helpers (`build`, `setup`, `packageMngrRun`, `runTestFramework`, `packageMngrPublish`)
  - `kubernetes.ts` — `kubectl/kind/helm` install checks and cluster helpers
  - `helm.ts` — helmfile commands and chart update helpers
  - `certs.ts` — certificate generation and reading helpers
- Updates all internal importers to import directly from the new files instead of `scripts.ts`
- Deletes `scripts.ts`
- Passes `options.trace` / `options.debug` into `writeTerasliceConfig` so the Teraslice log level in the generated config matches what was passed to the test runner

Ref to issue #4383


packages/scripts/src/helpers/exec.ts
```diff
@@ -0,0 +1,120 @@
...
+
+export async function runTSScript(cmd: TSCommands, args: string[]): Promise<void> {
+    const scriptName = process.argv[1];
+    return fork({
+        cmd: scriptName,
+        args: [cmd, ...args],
+    });
+}
+
+export async function pgrep(name: string): Promise<string> {
```
busma13: I would have put this in misc

packages/scripts/src/helpers/helm.ts
```diff
@@ -0,0 +1,357 @@
...
+ * @throws {Error} If the `Chart.yaml` file cannot be read
+ * @returns {Promise<string>} Resolves with the Helm chart version as a string
+ */
+export async function getCurrentHelmChartVersion(): Promise<string> {
+    const chartYamlPath = path.join(getRootDir(), '/helm/teraslice/Chart.yaml');
+    const chartYAML = await load(fs.readFileSync(chartYamlPath, 'utf8')) as any;
+    return chartYAML.version as string;
+}
+
+function getTerasliceVersion() {
```
busma13, line 111: should this be in misc? Maybe move it if we ever need this for non-helm stuff.

packages/scripts/src/helpers/kubernetes.ts
```diff
@@ -0,0 +1,191 @@
...
+async function showAssets(tsPort: number) {
+    try {
+        const subprocess = await execaCommand(`curl ${config.HOST_IP}:${tsPort}/v1/assets`);
+        return subprocess.stdout;
+    } catch (err) {
+        return err;
+    }
+}
+
+export async function logTCPPorts(service: string) {
```
busma13: This one is only used for services started with docker. I think it could go in misc though

packages/scripts/src/helpers/package-manager.ts
```diff
@@ -0,0 +1,183 @@
...
+}
+
+/**
+ * Publish package to NPM using yarn (versions 2, 3, 4) or pnpm
+*/
+export async function packageMngrPublish(
+    pkgInfo: PackageInfo,
+    tag = 'latest',
+): Promise<void> {
+    const pm = getPackageManager();
```
busma13, line 179: Should we move getPackageManager() into this file too?

busma13 (approved):
There are a few functions that could live somewhere else, but that could be done later, or never.

Issue #4495, opened 2026-07-09
On job stop a shutdown race condition can cause slice to be marked complete but the es_bulk write fails


busma13:
I was looking into an elasticsearch e2e recovery test that was flaky, and it looks like the problem might be in the teraslice shutdown code. Original issue: https://github.com/terascope/elasticsearch-assets/issues/1601

I can reproduce the issue by repeatedly running an id_reader to es_sender job and stopping the job. This is in native clustering with 4 workers. 

Failure example:

The last slice to be dispatched, slice_id 731517f0-96fe-43a4-bf69-7c7a021f0ee3, is marked as complete in the state store:

```
curl -X GET "http://localhost:49210/ts_test-0t__state-2026.07/_search" -H 'Content-Type: application/json' -d'
{
  "query": {
    "bool": {
      "must": [
        { "term": { "slicer_order": 35 }},
        { "term": { "ex_id": "b4bceba2-e1e8-4866-a9a8-e0a51e8c9bd3" }}
      ]
    }
  }
}
' | jq
{
  "took": 6,
  "timed_out": false,
  "_shards": {
    "total": 1,
    "successful": 1,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": {
      "value": 2,
      "relation": "eq"
    },
    "max_score": 1.5905983,
    "hits": [
      {
        "_index": "ts_test-0t__state-2026.07",
        "_id": "731517f0-96fe-43a4-bf69-7c7a021f0ee3",
        "_score": 1.5905983,
        "_source": {
          "slice_id": "731517f0-96fe-43a4-bf69-7c7a021f0ee3",
          "slicer_id": 0,
          "slicer_order": 35,
          "request": "{\"keys\":[\"ec\"],\"count\":398}",
          "state": "completed",
          "ex_id": "b4bceba2-e1e8-4866-a9a8-e0a51e8c9bd3",
          "_created": "2026-07-09T14:34:11.393Z",
          "_updated": "2026-07-09T14:34:12.755Z"
        }
      },
      {
        "_index": "ts_test-0t__state-2026.07",
        "_id": "e99a3957-3e02-4a8f-b92f-68d6b0587ef8",
        "_score": 1.5905983,
        "_source": {
          "slice_id": "e99a3957-3e02-4a8f-b92f-68d6b0587ef8",
          "slicer_id": 1,
          "slicer_order": 35,
          "request": "{\"keys\":[\"fc\"],\"count\":394}",
[...]

PR #4499, opened 2026-07-15
prevent race condition when shutdown called while slice is initializing

busma13:
TBD

ref: #4495

### DIFF
--- a/packages/teraslice/src/lib/workers/worker/index.ts
+++ b/packages/teraslice/src/lib/workers/worker/index.ts
@@ -28,6 +28,7 @@ export class Worker {
     isShuttingDown = false;
     isProcessing = false;
     isInitialized = false;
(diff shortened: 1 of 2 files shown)


packages/teraslice/src/lib/workers/worker/index.ts
```diff
@@ -205,6 +215,11 @@ export class Worker {
                 }
             }
 
+            if (this.isShuttingDown || this.forceShutdown || this.shouldShutdown) {
+                this.isProcessing = false;
+                return;
+            }
```
busma13, line 221: Shutting down at this point is what causes one of the shutdown tests to fail. We had considered doing this same check immediately before `this.slice.initialize` instead/also.

busma13:
Closing. This problem is addressed by #4503 instead.

PR #4507, opened 2026-07-24
[Data-types] add documentation

busma13:
- Add a list of data types and examples to the data-types overview.md
- Add JSDoc comments to all data type classes so the auto-generated TSDoc pages have some information.
- Add a test to ensure all V1 types are linked from the overview.md


docs/packages/data-types/overview.md
```diff
@@ -13,3 +13,336 @@ pnpm add @terascope/data-types
...
+- [`Double`](./api/types/v1/double/classes/default.md) — double-precision float (ES `double`).
+- [`Number`](./api/types/v1/number/classes/default.md) — general-purpose numeric when the width doesn't matter (ES `double`).
+
+### String / keyword / text
+
+See [Choosing a string type](#choosing-a-string-type) below for how to pick between these.
+
+- [`Keyword`](./api/types/v1/keyword/classes/default.md) — exact-match string for filtering, sorting, aggregations.
+- [`String`](./api/types/v1/string/classes/default.md) — like `Keyword`, but always GraphQL `String` (no `_key`→`ID` special-case).
+- [`Text`](./api/types/v1/text/classes/default.md) — analyzed full-text string.
```
jsnoble: are snippets needed, a little to vague for use

packages/data-types/src/types/v1/geo-json.ts
```diff
@@ -5,6 +5,23 @@ import {
 import BaseType from '../base-type.js';
 import { GraphQLType, TypeESMapping } from '../../interfaces.js';
 
+/**
+ * An arbitrary GeoJSON geometry (point, line, polygon, etc.) stored as an
```
jsnoble: we don't use lines

jsnoble (approved):
LGTM

docs/packages/data-types/field-configs.md
```diff
@@ -0,0 +1,211 @@
...
+| `engine` | `'faiss'` \| `'lucene'` | `Vector` only | ES mapping |
+
+`DataTypeFieldConfig` has an open index signature (`[extra: string]: unknown`),
+so **unrecognized options are accepted and silently ignored** — as is an option
+placed on a type that doesn't read it. `dimension: 5` on a `Keyword`, or
+`format` on anything but a `Date`, produces no error and no output change. The
+one thing validated at construction time is that each field name is legal and
+has a `type`; everything else is validated (if at all) when a conversion runs.
+
+## `array`
```
godber: Should this be a subsection of `## Options`?

Same with the rest of these things?

packages/data-types/src/types/v1/binary.ts
```diff
@@ -5,6 +5,27 @@ import {
...
+ * - **ES/OpenSearch mapping:** `{ type: 'binary', doc_values: false }`.
+ *   `doc_values` is always forced off because binary fields are not
+ *   searchable and can be very large. Honors `indexed: false` (emitted as
+ *   `index: false`).
+ * - **GraphQL:** `String`.
+ * - **xLucene:** `String`.
+ *
+ * Because it is neither indexed for search nor stored as doc values, a
+ * `Binary` field is opaque storage — you can retrieve it by `_source` but
+ * cannot filter, sort, or aggregate on it.
```
godber: Ideally this would suggest an alternative ... which presumably is `Boolean`.

godber: This whole thing begs the question, what is this good for?

godber (changes requested):
There are a few minor things that are probably worth changing here.  But overall this is a great effort.

packages/data-types/src/types/v1/binary.ts
busma13: > This whole thing begs the question, what is this good for?

Storage. This data type was added because we store asset blobs in opensearch as binary. The teraslice store mappings used to be static, but now they are generated dynamically from our data type definitions, so we had to add this data type. If we want to search for an asset record we use the name or id field.

I would think the closest alternative data type that would allow you to search would be `text`. We use binary when we explicitly don't want to tokenize the data.

godber: Oh this is a `BinaryString`.  My bad ...

godber (approved):

PR #4508, opened 2026-07-27
[teraslice, job-components, types] Remove worker_version fields from teraslice and job configs, use v2 worker code by default

busma13:
- remove `worker_version` field from terafoundation.teraslice config and job config.
- Run the `v2` Worker code by default. Functions will have their original names.
- remove types and documentation changes.

Ref: #4495, #4503


sotojn (approved):
Overall looks good

godber (approved):

Issue #4512, opened 2026-07-31
[e2e] Test assets built with current PR packages


busma13:
Teraslice e2e tests run the current teraslice against the latest prerelease of the elasticsearch, standard and kafka assets. If a current teraslice package change is incompatible with a previous asset then we will see an error and tests will fail.

But we do not build the current packages into an asset and test it against the current teraslice or against a previous teraslice version. Doing so and seeing test failures would notify us **before a package is published** that it will break compatibility in some way. This will allow us to either make changes to the current PR to keep things compatible, or signal that the `minimum_teraslice_version` should be updated when the assets start using packages from this PR.

My plan is to add two e2e test runs:
- teraslice docker image built from current PR using assets built with packages from current PR
- `latest` teraslice image using assets built with packages from current PR

I suspect there will be many cases where the assets cannot be built successfully, so these tests should give us obvious feedback, but not actually fail the pipeline. Maybe a PR comment explaining the failure type and what manual testing should be considered before PR approval would be helpful?

busma13:
If an interface breaks between a teraslice package and code in an asset we get a build failure. In this case I just changed the name of the `warnings` field in job-components. Since kafka_reader_api  has been update to show validation warnings we get a type error. Elasticsearch-assets built successfully because has not been updated to surface warnings yet:
```
ℹ  info      Building elasticsearch-assets from source..
...
✔  success   Built elasticsearch-v5.1.8-node-24-bundle.zip from elasticsearch-assets source
ℹ  info      Building kafka-assets from source..
...

* verified 3 packages injected into asset/: @terascope/core-utils, @terascope/job-components, @terascope/types
* running pnpm run build in the asset repo
$ tsc --build
asset/src/kafka_reader_api/schema.ts:104:31 - error TS2339: Property 'warnings' does not exist on type '{ config: APIConfig & { [prop: string]: any; topic: string; group: string; offset_reset: "error" | "smallest" | "earliest" | "beginning" | "largest" | "latest"; _connection: string; ... 9 more ...; _dead_letter_action?: string | undefined; }; warningList: JobWarning[]; }'.

104             warnings: results.warnings
                                  ~~~~~~~~


Found 1 error.

[ELIFECYCLE] Command failed with exit code 2.
Command failed with exit code 2: pnpm run build
[...]

PR #4513, opened 2026-08-03
asset compat testing

busma13:
- Adds a job to `test.yml` that tests assets built with the teraslice packages from the current PR against 2 scenarios: the current PR teraslice build or the `latest` teraslice release. These jobs never fail. They are informative only.
A comment will be added if either have failing tests or setup errors (like when the asset doesn't even build).
- Adds several scripts:
  - build an asset with injected packages. The asset version is set to `9999` so that it will load instead other asset versions
  - a custom jest exporter that allows a written report to be produced, even when e2e tests bail
  - create a PR comment from the jest report

ref: #4512


busma13:
Link to a report Summary for a CI run where the asset compatibility tests succeed: https://github.com/terascope/teraslice/actions/runs/30925000606/attempts/1#summary-92051183538. In this scenario we do not create a comment.

sotojn (approved):

Issue #4514, opened 2026-08-06
`[data-mate]` Six defects in DataFrame sorting, coercion and string handling — with failing tests


jsnoble:
Six defects in `packages/data-mate`'s `DataFrame`, each reproduced by a failing test.

Two of them corrupt query results silently and at scale: **76% of multi-key sorts return the wrong
order**, and **any sort over a column containing nulls can misorder the non-null values**.

A branch adds `packages/data-mate/test/known-defects-spec.ts` — 12 tests that assert the *correct*
behaviour and therefore fail. No existing test or source file was modified.

```bash
# 12 failures, all in one file
node ../scripts/bin/ts-scripts test ../data-mate -- --testPathPatterns known-defects

# rest of the suite is unaffected: 28 suites, 2429 passed
node ../scripts/bin/ts-scripts test ../data-mate -- --testPathIgnorePatterns known-defects
```

| id | defect | severity |
|---|---|---|
| D1 | multi-key `orderBy` sums per-field comparisons instead of taking first-non-zero | **high** |
| D2 | a nil makes the comparator non-transitive, misordering **non-nil** values | **high** |
| D3 | `reverse()` splits astral characters into lone surrogates | medium |
| D4 | integer coercion truncates at the first non-digit | medium |
| D5 | date coercion delegates to the JS `Date` parser | medium |
| D6 | two existing tests assert the buggy output as expected | blocks fixing D1/D2 |

---

### D1 — multi-key `orderBy` sums per-field comparisons

`src/vector/Vector.ts:45-51`

```js
sortBy.reduce((acc, { vector, direction: d }, i) => {
    const res = vector.compare(a[i], b[i]);
    return acc + (d === 'asc' ? res : -res);   // <-- summed, should short-circuit
}, 0)
```

A correct comparator returns the first non-zero result. Summing means two keys that disagree cancel to
`0`, the comparator reports "equal", and the rows keep input order — so the primary key is ignored.

| | |
|---|---|
| input | `[{a:'y', b:2}, {a:'x', b:1}]` |
| call | `.orderBy('a:asc', 'b:desc')` |
| expected | `['x1', 'y2']` — `'x' < 'y'`, so `x1` first regardless of `b` |
| **actual** | **`['y2', 'x1']`** |

`compare('y','x')` is `+1`; `b:desc` inverts `compare(2,1)` to `-1`; sum is `0`.

[...]

PR #4515, opened 2026-08-10
Dataframe bug fix

jsnoble:
- fixes issues described [here](https://github.com/terascope/teraslice/issues/4514)
- Fixed: D1, D2, D3, D4, D6
- Open: D5


busma13:
The changes you've made so far all look good.

I would think integer coercion (D4) would be a small fix that is worth doing now. I can't imagine that we would ever expect this to happen and be relying on it anywhere. 

The date coercion (D5) looks trickier and will need a lot of thought to ensure we don't break any assumptions elsewhere. It probably deserves its own PR.

busma13 (approved):
LGTM

Issue #4506, opened 2026-07-22
[Teraslice] Functions not being awaited in Worker and WorkerExecutionContext


busma13:
While working on #4495 we ran across a few async functions that are not awaited. We should ensure that we are not relying on them not being awaited, then add awaits and test that everything works properly.

PR #4522, opened 2026-08-13
[teraslice, job-components] Await lifecycle hooks

busma13:
- There were two async [worker lifecycle event hooks](https://terascope.github.io/teraslice/docs/jobs/slices/#worker-lifecycle-events) that were not being properly awaited: `onSliceInitialized` and `onSliceFailed`. A search of both public and internal assets did not reveal any uses of these hooks, which is probably why this hadn't been noticed before.

Ref: #4506


jsnoble (approved):
LGTM

Issue #4485, opened 2026-06-22
Upgrade to typescript 6


busma13:
Typescript 6 has been out for a while. It is a bridge release that deprecates many features to get a repo ready for typescript 7, which should be much faster.
Release Notes: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html

Deprecations that may affect us:
- `moduleResolution: node` - switch to `nodenext`
- `baseUrl` - add prefix to `paths` directly
- `types` defaults to `[]` instead of automatically pulling in all types from node_modules. We specify `typeRoots` so I don't think we need to change anything.

busma13:
After making the changes for v6 I was able to update to v7 cleanly. Maybe we should go straight to v7?

busma13:
Nevermind, a clean CI build with typescript v7.0.2 will fail. The packages build concurrently and we don't include `references` in each package's tsconfig to ensure dependencies are built first.

Also `typedoc` and `typescript-eslint` break because v7 doesn't have an API implemented yet See https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/#running-side-by-side-with-typescript-6.0

PR #4523, opened 2026-08-14
Typescript v6

busma13:
- bump Typescript from v5.9.3 to v6.0.3
- tsconfig.json updates:
  - `moduleResolution: node` is deprecated - use `nodenext`
  - `module: ESNext` is incompatible with `moduleResolution: nodenext` - use `nodenext`
  - `baseUrl` field deprecated - remove and update paths where needed
  - `types` has no default - declare `node` and `jest` types as global
  - add `paths` objects for packages that do not declare their types properly
- add `.js` or `/index.js` to any imports still missing a file extension
- fix custom types for `latlon-geohash` and `datemath-parser` libraries
- fix imports for `yargs` amd `Ajv`
- move `lodash-es` and `validator` types to `dependencies` in core-utils so the types are re-exported
- fix publish config errors found with `publint`
- update `teraslice-cli assets registry` command to include `.js` extension on import lines

refs:
https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/
#4485


jsnoble (approved):
LGTM

Issue #4500, opened 2026-07-16
[docs] Update prometheus metrics documentation


busma13:
The docs for the prometheus metrics are very basic. We should at the minimum improve the examples for using `collect()` with an `inc` metric, explaining how the counter in the code should be reset after `collect()` runs.

We also need to fix the website so that the document shows up in the development section sidebar.

Issue #4056, opened 2025-05-14
[terafoundation] Add prom metrics wrapper classes for counter and gauges


sotojn:
I'm working on an internal terafoundation application that is being updated to use the new prom metrics. I made some facade class for creating and using `counters` and `gauges`. @godber pointed out that these are generic and should live in `terafoundation` itself. 

These are useful because these metric objects clean up redundant code and make creating metrics with branched out labels far simpler to create. It also considers if prom is disabled or unavailable so the user creating metrics don't have to worry about guarding against it in every spot a metric is utilized.

PR #4525, opened 2026-08-19
Update prom metrics docs

sotojn:

- Adds better docs for prom metrics for developers

Ref to issue #4500


sotojn:
I wonder if we should actually implement this to simplify the the api

https://github.com/terascope/teraslice/issues/4056

busma13:
> I wonder if we should actually implement this to simplify the the api
> 
> #4056

That's not a bad idea

godber (approved):

Issue #3892, opened 2025-01-07
Properly type teraslice API file


busma13:
When `api.ts` was converted to typescript the types were not all figured out, so there are a lot of coerced types. We should spend some type getting the `express` and `got` types to work properly with our code, especially with `TerasliceRequest` and `assetRedirect().

PR #4526, opened 2026-08-21
[teraslice] properly type the API service

sotojn:

- Types the `logger` that middleware attaches to every request via a `declare global` Express `Request` augmentation in `interfaces.ts`, and converts `TerasliceRequest`/`TerasliceResponse` to aliases
  - Removes `as TerasliceRequest` casts across `api.ts` and `assets.ts`
- Types the JSON-parse error middleware as an `ErrorRequestHandler` instead of suppressing it
- Widens `handleTerasliceRequest`'s callback to `() => unknown` (narrowed at the point of use), and expresses the legacy `slicer` alias as a spread instead of an `@ts-expect-error` mutation
- Removes all `@ts-expect-error` directives from `api.ts`

Type-only change, should be no runtime or http behavior change.

Ref to issue #3892

### DIFF
--- a/packages/teraslice/src/interfaces.ts
+++ b/packages/teraslice/src/interfaces.ts
@@ -23,10 +23,12 @@ declare global {
 }
 
 /**
- * Aliases kept for readability at call sites; `any` params let route handlers
- * with typed params (e.g. `Request<{ jobId: string }>`) pass without casting.
+ * Aliases kept for readability at call sites. The `P` generic forwards to
+ * express `Request`'s route-params type, so handlers can express typed params
+ * (e.g. `TerasliceRequest<{ jobId: string }>`) while defaulting to `any` lets
+ * untyped handlers pass without casting.
  */
-export type TerasliceRequest = Request<any>;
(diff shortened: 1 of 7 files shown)


packages/teraslice/src/interfaces.ts
```diff
@@ -8,11 +8,27 @@ import type {
...
+/**
+ * Every request passes through middleware that attaches `logger` (see ApiService
+ * and AssetsService). Declaration merging makes that known to the type system so
+ * express `Request` no longer has to be cast to a Teraslice-specific type.
+ */
+declare global {
+    // eslint-disable-next-line @typescript-eslint/no-namespace
+    namespace Express {
+        interface Request {
+            logger: Logger;
```
busma13, line 20: This only affects the teraslice package, so the express server in Terafoundation will not require a logger.

busma13 (approved):
LGTM. Merging the Logger into the Request type is weird, but I can't think of a better way to do this.

Can you bump the version?

jsnoble (approved):
LGTM, you can bring it in after adding the generic to the tereaslice request type

Issue #4492, opened 2026-06-30
[scripts] simplify ts-scripts sync command


busma13:
`ts-scripts sync` is supposed to keep all package versions in sync. We could instead use pnpm [Catalogs](https://pnpm.io/catalogs) for this. 

It also does some other things, like updating the readme and overview, seeing if files are committed, etc. We would want to keep some of this, so the code needs to be looked at closely.

busma13:
`ts-scripts sync` is used by the publish and bump scripts as well.

PR #4527, opened 2026-08-26
[scripts] replace ts-scripts sync command with pnpm catalog 

sotojn:

- Adds a pnpm [catalog](https://pnpm.io/catalogs) in `pnpm-workspace.yaml` and references it as `catalog:` across every `package.json`
  - pnpm now keeps external dependency versions in sync at install time, which is what `ts-scripts sync` used to do
- Removes the external version unification logic from `syncVersions` in `sync/utils.ts`
- Updates the `sync` command description to match new functionality 
- Updates `sync-spec.ts` to match the new behavior

No dependency versions actually change and `bump`/`publish`/`bump-asset` still call `syncAll` the same way.

Ref to issue #4492


busma13 (changes requested):
Scripts should get bumped

busma13 (approved):

Issue #4529, opened 2026-08-28
[CI] determine-tests-based-off-commit ci job not working right


sotojn:
## Issue

Austin noticed that one of my docs PR's ran all of the test suites. This isn't right as the point of the `determine-tests-based-off-commit` job is to check the diff and only run specific tests depending on what type of files are changed. In this case we would expect it to see that only `.md` files were changed, added, or removed, therefore only running website tests. This was not the case in this PR:
 
https://github.com/terascope/teraslice/pull/4525

PR #4534, opened 2026-09-01
[CI] fix bug in determine-tests-based-on-commit job

sotojn:

- Fixes the `determine-tests-based-off-commit` CI job so documentation-only PRs no longer run the unit, e2e, and integration suites
  - Previously the check only matched `docs/`, so a docs PR that also touched `website/` (e.g. `website/sidebars.json`) was misclassified as a code change and ran every suite
- Adds a `changedFiles.length > 0` guard so an empty diff isn't treated as docs-only

Ref to issue #4529

### DIFF
--- a/scripts/determine-tests.js
+++ b/scripts/determine-tests.js
@@ -117,8 +117,21 @@ export function parseUnifiedDiff(diff) {
     return changes;
 }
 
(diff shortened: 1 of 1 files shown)


busma13 (approved):
LGTM

PR #4535, opened 2026-09-02
[xLucene] Allow separate geo sort field than geo query

lesleydreyer:
Main changes - translator and geo-dist fn were preventing geo_sort params to override the default sort for a geo_distance query 

- [x] Fix opensearch sort types
- [x] Fix so geo sort args can override the sort that the geo distance query adds
- [x] Disambiguate sorts - Opensearch accepts url & req body params - request body sort seems to take precedence over string url sort, but it's not documented so is confusing - figured just having one sort is clearer and less confusing so merged them
- [x] Remove unnecessary client metadata defaults  in query access - it sets defaults in translator so not needed twice

### DIFF
--- a/packages/core-utils/package.json
+++ b/packages/core-utils/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@terascope/core-utils",
     "displayName": "Core Utils",
-    "version": "2.17.1",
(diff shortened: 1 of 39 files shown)


jsnoble (approved):
LGTM

Issue #4541, opened 2026-09-14
Pin pnpm to 24.19


sotojn:
## Issue 

We are getting a failing test on the master branch when we started using node 24.20.*. We know this is the case becasue we did a pin to a test branch to 24.19 and it resolves. I will update the issue here when I can find out what exactly changed in node to cause this.

Link to failed test:

https://github.com/terascope/teraslice/actions/runs/34654081741/job/103442843788?pr=4540#step:7:425

Here is the error in full below as well:

```
 FAIL     packages/terafoundation/test/apis/exporter-spec.ts
  ● prometheus exporter › shutdown › should shutdown the express server

    expect(received).rejects.toThrow(expected)

    Expected substring: "connect ECONNREFUSED 127.0.0.1:3344"
    Received message:   "Socket closed before the connection was established"

      84 |             await expect(() => got('http://127.0.0.1:3344/metrics', {
      85 |                 throwHttpErrors: true
    > 86 |             })).rejects.toThrow('connect ECONNREFUSED 127.0.0.1:3344');
         |                         ^
      87 |         });
      88 |     });
      89 | });

      at Object.toThrow (node_modules/.pnpm/expect@30.4.1/node_modules/expect/build/index.js:2155:20)
      at Object.toThrow (packages/terafoundation/test/apis/exporter-spec.ts:86:25)
```

PR #4542, opened 2026-09-14
pin node version to 24.19

sotojn:

- Pins node version to `24.19` becasue of a test issue on `24.20`

Ref to issue #4541


busma13 (approved):
LGTM

PR #4546, opened 2026-09-14
Unpin node 24

sotojn:

- Reverts the `24.19` node pins back to `24` in the workflow files
  - Reverts #4542
- Updates the prometheus exporter shutdown test so it passes on any node version
  - Node `24.20` reworked the `net` module so connecting to a closed port now rejects with `ERR_SOCKET_CLOSED_BEFORE_CONNECTION` instead of `ECONNREFUSED`, so the test now matches on the error code (accepting either) instead of the hard-coded message

Ref to issue #4541


busma13 (approved):
LGTM

Issue #4544, opened 2026-09-14
Replace minio docker image


busma13:
MinIO removed their images from dockerhub, so teraslice tests as well as file-assets tests will no longer pass.

Using [ceph](https://ceph.io/en/) seems like a good idea since we already use it in production, but it is not a drop in replacement.

Alternatives I've seen mentioned:
- VaultS3 - https://vaults3.com/
- rustfs - https://rustfs.com/
- silo - https://silo.pgsty.com/. forked from minio
- buckit - https://buckit.sh/. forked from minio
- garage - https://garagehq.deuxfleurs.fr/
- SeaweedFS - https://seaweedfs.github.io/
- VersityGW - https://github.com/versity/versitygw/ (ASG edit after the fact)

I'll do some quick research on each of these to see if any seem like they may stick around.

godber:
I would only consider garage or seaweedfs in this list.


But taking the time to do Ceph will probably pay off in that it will still be here in ten years.

busma13:
I think it would be beneficial for Joseph and I to get some Ceph experience. I'll start down that route and see how it goes.

busma13:
We broke this into several parts:

- [x] Ceph from docker compose #4547
- [ ] Ceph docker TLS support #4549
- [ ] Ceph from Rook in kubernetes #4550
- [ ] Ceph Rook TLS support
- [x] Assets
  - [x] https://github.com/terascope/chaos-assets/pull/147
  - [x] https://github.com/terascope/file-assets/pull/1434

PR #4554, opened 2026-09-22
[scripts] Add TLS for ceph kind

sotojn:

- Add TLS support for Ceph on the kubernetesV2/kind test path, following the Docker TLS work in #4549
- Auto generate the RGW certs with mkcert for both the env var and config file k8s-env paths
- Serve the RGW over https in cluster (teraslice connector) and on the host NodePort, backed by a k8s secret
- Cap Ceph memory limits so the whole cluster fits on one kind node
- Add a prometheus scrape interval config 

Ref to issue #4544


busma13 (approved):
LGTM

